PR #91, opened 2024-01-22
fix: update node to lts/20, semantic release, misc

solaris007:
- [x] Update to LTS Node 20
- [x] Update major semantic-release + monorepo
- [x] Remove renovate dependency restrictions for semantic-release
- [x] Add chai to dependency restrictions
- [x] Fix rum-api-client test

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 executors:
   node-java:
     docker:
-      - image: cimg/node:18.19-browsers
+      - image: cimg/node:20.11.0-browsers
 
 orbs:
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 executors:
   node-java:
```
iuliag, line 3: Less relevant, but why the naming `node-java`?

iuliag (approved):

.circleci/config.yml
solaris007: @iuliag because we need java for the data-access IT test - the embedded dynamodb is java-based

PR #102, opened 2024-01-23
feat: slack client to support multiple org

ekremney:

## Related Issues

Thanks for contributing!


dzehnder (approved):

packages/spacecat-shared-slack-client/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+      .reduce((acc, pair) => {
+        acc[pair.target] = new WebClient(pair.token);
+        return acc;
+      }, {});
+    this.log = log;
+  }
+
+  #getClient(target) {
+    if (!this.clients[target]) {
+      throw new Error(`Environment variable '${ENV_PREFIX}${target}' does not exist. Slack Client could not initialized.`);
```
solaris007, line 63: `could not *be* ...` or `failed to initialize`

packages/spacecat-shared-slack-client/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+const DEFAULT_PARAMS = {
+  unfurl_links: false,
+  unfurl_media: false,
+};
+
+export class SlackClient {
+  static createFrom(context) {
+    if (context.slackClient) return context.slackClient;
+
+    const expected = Object.values(SLACK_TARGETS).map((target) => `${ENV_PREFIX}${target}`);
```
solaris007, line 30: i suggest wrapping `${ENV_PREFIX}${target}` into a little function, it is used multiple times

packages/spacecat-shared-slack-client/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+    if (targetTokenPairs.length === 0) {
+      throw Error('No environment variable containing a slack token found');
+    }
+
+    this.clients = targetTokenPairs
+      .reduce((acc, pair) => {
+        acc[pair.target] = new WebClient(pair.token);
+        return acc;
+      }, {});
```
solaris007, line 57: could use destructuring:
```suggestion
      .reduce((acc, { target, token }) => {
        acc[target] = new WebClient(token);
        return acc;
      }, {});
```

packages/spacecat-shared-slack-client/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+        ...DEFAULT_PARAMS,
+        ...opts,
+      });
+
+      return {
+        channelId: result.channel,
+        threadId: result.ts,
+      };
+    } catch (error) {
+      this.log.error(`Slack message failed to send to ${target}`);
```
solaris007, line 81: please include more details, i.e. message and/or stack trace

packages/spacecat-shared-slack-client/src/index.d.ts
```diff
@@ -0,0 +1,46 @@
...
+   */
+  constructor(targetTokenPairs: TargetTokenPair[]);
+
+  /**
+   * Asynchronous method to create a RUM backlink.
+   * @param {string} target - A string representing the target slack workspace/org
+   * @param {number} opts - An object containing the data sent to Slack API
+   * @returns A Promise resolving to an object containing the data returned by Slack API
+   */
+  postMessage(target: string, opts: object): Promise<object>;
```
solaris007, line 45: i would simplify the signature to postMessage(message, opts) - most people want to send a simple message and don't want to deal with knowing slack internal options, which still can be optionally provided...

solaris007 (commented):
i'd propose to split the client-factory part from the actual slack client.  this aligns with the principle of single responsibility, ensuring that each client instance is dedicated to a specific target (i.e., a Slack workspace). This redesign would simplify the client interface by removing the need to specify the target for every public method call.

Incomplete example:
```js
class SlackClientFactory {
  constructor(log) {
    this.clients = {};
    this.log = log;
  }

  // createFrom / lambda context handling here...

  getClientForTarget(target) {
    if (!this.clients[target]) {
      const token = this.getTokenForTarget(target);
      if (!token) {
        throw new Error(`No token found for target: ${target}`);
      }
      this.clients[target] = new SlackClient(token, this.log);
    }
    return this.clients[target];
  }

  getTokenForTarget(target) {
    return process.env[`${ENV_PREFIX}${target}`];
  }
}

class SlackClient {
  constructor(token, log) {
    this.client = new WebClient(token);
    this.log = log;
  }

  async postMessage(message, opts) {
    try {
      const result = await this.client.chat.postMessage({
        ...DEFAULT_PARAMS,
        text: message,
        ...opts,
      });

      return {
        channelId: result.channel,
        threadId: result.ts,
      };
    } catch (error) {
      this.log.error('Slack message failed to send');
      throw error;
    }
  }
}
```

packages/spacecat-shared-slack-client/src/index.d.ts
dbrrt, line 45:
```suggestion
  postMessage(target: string, opts?: object): Promise<object>;
```

ekremney:
I incorporated all the reviews and added the file upload method. Please re-review

packages/spacecat-shared-slack-client/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@adobe/spacecat-shared-slack-client",
+  "version": "1.2.1",
```
solaris007, line 3: are we already at this version? should probably be reset.

packages/spacecat-shared-slack-client/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@adobe/fetch": "4.1.1",
+    "@adobe/helix-shared-wrap": "2.0.0",
+    "@adobe/helix-universal": "4.4.1",
+    "@adobe/spacecat-shared-utils": "1.7.2",
+    "@slack/web-api": "7.0.1",
+    "aws4": "1.12.0"
```
solaris007, line 37: i think that should be removed

packages/spacecat-shared-slack-client/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "url": "https://github.com/adobe/spacecat-shared/issues"
+  },
+  "homepage": "https://github.com/adobe/spacecat-shared#readme",
+  "publishConfig": {
+    "access": "public"
+  },
+  "dependencies": {
+    "@adobe/fetch": "4.1.1",
+    "@adobe/helix-shared-wrap": "2.0.0",
+    "@adobe/helix-universal": "4.4.1",
```
solaris007, line 34: those should probably be removed.

helix-universal - is it worth it just for the type def?

packages/spacecat-shared-slack-client/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+    return {
+      channelId: result.channel,
+      threadId: result.ts,
+    };
+  }
+
+  async fileUpload(file) {
+    const result = await this.#apiCall('files.uploadV2', file);
+
+    if (result.files.length === 0) {
```
solaris007, line 76: are you sure `result.files` is always defined? I've seen alternate responses in the doc like:
```
{
    "ok": false,
    "error": "invalid_auth"
}
```

solaris007 (approved):
LGTM, only minor comments

packages/spacecat-shared-slack-client/src/index.js
ekremney: when `ok` is not `true`, the `WebClient` throws an error.

packages/spacecat-shared-slack-client/package.json
ekremney: I'd rather keep it for the type defs

solaris007 (approved):

PR #109, opened 2024-01-24
A shared library to pull AWS CostUsage data

sudo-buddy:

## Related Issues

Thanks for contributing!


packages/spacecat-shared-aws-costusage-client/src/index.js
```diff
@@ -0,0 +1,77 @@
...
+   * @param {string} startDate like 2021-01-01
+   * @param {string} endDate like 2021-01-31
+   * @param {string} granularity like MONTHLY
+   * @param {string[]} metrics like UnblendedCost
+   * @param {Object[]} groupBy like [{Key: "SERVICE", Type: "DIMENSION"}, {Key: "Environment", Type: "TAG"}]
+   * @param {Object} filter like {Tags: {Key: 'Adobe.ArchPath', Values: ['EC.SpaceCat.Services'], MatchOptions: ['EQUALS']}} 
+   * @returns {GetCostAndUsageResponse}
+   */
+
+  async getCostUsageData(startDate, endDate, granularity, metrics, groupBy, filter) {
```
ekremney, line 53: I think the required params here are only `service`, `startDate` and `endDate`. Rest of the params should be constant and handled internally.

An example API call could be like this:

```js
const service = 'S3'; // service code would be better if possible
const startDate = '2024-01-01';
const endDate = '2024-01-31';

const usageCost = await getCostUsageData(service, startDate, endDate);  

// example
console.log(`${service} used $${usageCost} in january 2023`);
// logs: "S3 used $12 in january 2023"
```

packages/spacecat-shared-aws-costusage-client/src/index.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+export default class AWSCostApiClient {
+  static createFrom(context) {
+    if(context.cogsApiClient) return context.cogsApiClient;
+    const client = new AWSCostApiClient(context);
+    context.cogsApiClient = client;
+    return client;
+  }
+
+  constructor(context) {
```
ekremney, line 29: we have `AWSCostApiClient.createFrom(context)` which takes the helix universalcontext as param and knows about implementation of helix universalcontext.

constructor should not know about helix universalcontext. In this case it should be like:

```js
AWSCostApiClient(region, credentials, log);
```

packages/spacecat-shared-aws-costusage-client/test/index.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+// eslint-disable-next-line import/no-named-default
+import { AWSCostApiClient } from '../src/index.js';
+import chaiAsPromised from 'chai-as-promised';
+import nock from 'nock';
+
+chai.use(chaiAsPromised);
+const { expect } = chai;
+
+describe('run aws api client', () => {
```
ekremney, line 25: this test does not verify the actual result. We can mock the amazon api behind the service using `nock`. Example for that: https://github.com/adobe/spacecat-api-service/blob/main/test/support/sqs.test.js#L82

ekremney:
@sudo-buddy as a rule of thumb, we put shared code which is used by at least two different components into spacecat-shared. Since this logic will only be used by audit worker, we can completely skip creating a new package here and add the code in the cogs handler in audit worker directly

solaris007:
closing in favor of https://github.com/adobe/spacecat-audit-worker/pull/115

PR #133, opened 2024-02-06
feat: experimentation RUM API method

raji-unni:
…xperimentation data

## Related Issues

Thanks for contributing!


packages/spacecat-shared-rum-api-client/src/index.js
```diff
@@ -120,6 +121,15 @@ export default class RUMAPIClient {
...
 
+  createExperimentationURL(params = {}) {
+    return createUrl(
+      APIS.RUM_EXPERIMENTS,
+      {
+        domainkey: this.domainkey, ...RUM_DEFAULT_PARAMS, ...params,
+      },
+    );
+  }
+
```
ekremney, line 132: it should rather be:

```suggestion
  async getExperimentation(params = {}) {
    return sendRequest(createUrl(
      APIS.RUM_EXPERIMENTS,
      { domainkey: this.domainkey, ...RUM_DEFAULT_PARAMS, ...params },
    ));
  }

```

so that, consumer of this API can use the method like:

```js
rumAPIClient.getExperimentation({ url: 'www.adobe.com' });
```

raji-unni: Ekrem, I am only trying to get a url from the function which I will be calling from the handler. So I have done some changes. Please check.

raji-unni: Have implemented the changes as discussed with you.

ekremney (approved):

PR #138, opened 2024-02-07
Adobe Analytics Client API call library

sudo-buddy:

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-aa-api-client/README.md
+++ b/packages/spacecat-shared-aa-api-client/README.md
@@ -1,0 +1,104 @@
+# Spacecat Shared - ADobe Analytics API Client
+
+A JavaScript client for Adobe Analytics API, part of the SpaceCat Shared library.
+
+## Installation
+
+Install the package using npm:
+
+```bash
+npm install @adobe/spacecat-shared-aa-api-client
+```
+
+## Usage
+
+### Creating and instance from Helix UniversalContext
+
+```js
+const context = {}; // Your AWS Lambda context object
+const aaApiClient = AAAPIClient.createFrom(context);
+
+```
+
+### Constructor
+
+`AAAPIClient` class needs AA API domain key to be instantiated:
+
+```js
+const domainKey = "your-domain-key";
+const aaApiClient = new AAAPIClient(domainKey);
+```
+
+### Creating a AA Backlink
+
+```js
+const url = "https://example.com";
+const expiryInDays = 7;
+
+const backlink = await aaApiClient.createRUMBacklink(url, expiryInDays);
+console.log(`Backlink created: ${backlink}`)
+```
+
+### Creating a 404 Report Backlink
+
+```js
+const url = "https://example.com";
+const expiryInDays = 7;
+
+const backlink = await aaApiClient.create404Backlink(url, expiryInDays);
+console.log(`Backlink created: ${backlink}`)
+```
+
+### Getting RUM Dashboard Data
+
+```js
+const url = "example.com";
+
+const rumData = await rumApiClient.getRUMDashboard({ url });
+console.log(`RUM data: ${rumData}`)
+```
+
+### Getting 404 checkpoints
+
+```js
+const url = "example.com";
+
+const backlink = await rumApiClient.get404Sources({ url });
+console.log(`404 Checkpoints: ${backlink}`)
+```
+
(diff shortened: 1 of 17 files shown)


packages/spacecat-shared-aa-api-client/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+Install the package using npm:
+
+```bash
+npm install @adobe/spacecat-shared-aa-api-client
+```
+
+## Usage
+
+### Creating and instance from Helix UniversalContext
```
alinarublea, line 15: We should add the AAClient methods here (the ones bellow belong to the RUMAPIClient)

packages/spacecat-shared-aa-api-client/src/helpers.js
```diff
@@ -0,0 +1,121 @@
...
+export function getPreviousDay(dateString, offset) {
+  if (offset < 0) {
+    throw new Error('Offset must be a positive number');
+  }
+  const currentDate = !dateString ? new Date() : new Date(dateString);
+  currentDate.setUTCDate(currentDate.getDate() - (offset || 1));
+  return currentDate.toISOString().split('T')[0];
+}
+
+export function parseDates(inputText) {
```
alinarublea, line 36: parseDates is not used by the AAClient we could remove it

packages/spacecat-shared-aa-api-client/src/helpers.js
```diff
@@ -0,0 +1,121 @@
...
+    } else if (trimmedSegment) {
+      acc.push(trimmedSegment);
+    }
+    return acc;
+  }, []);
+
+  return dates;
+}
+
+export function validateConfig(config) {
```
alinarublea, line 75: I think validateConfig config should be added to spacecat-shared-data-access config model and renamed to validateRUMtoAAConfig

packages/spacecat-shared-aa-api-client/src/helpers.js
```diff
@@ -0,0 +1,121 @@
...
+    if (!entryKey.startsWith('event')) {
+      throw new Error(`Invalid specification under events, should be event$NUMBER, not ${entryKey}`);
+    }
+    if (!entryValue.rumField && !entryValue.default) {
+      throw new Error(`Missing default or value mapping for event ${entryValue} in project configuration under rum2aaMapping section`);
+    }
+  });
+}
+
+export function stringToUint8Array(content) {
```
alinarublea, line 115: This one is no longer used and we can remove

packages/spacecat-shared-aa-api-client/src/utils.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { context as h2, h1 } from '@adobe/fetch';
+
+/* c8 ignore next 3 */
```
alinarublea, line 15: you can removed this, is not used by the AAClient

packages/spacecat-shared-aa-api-client/src/rumtoaamapper.js
```diff
@@ -0,0 +1,95 @@
...
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+import { EVARS_KEY, EVENTS_KEY } from './helpers.js';
+
+export default class RUMToAAMapper {
```
alinarublea, line 14: I think the mapper belongs to the importer lambada I would remove the class from here

ekremney (commented):
the token is only initialized in the static factory method `create`. If the consumers of this class initializes this class via its constructor directly, then it won't work, because the token will be null. Maybe a private init function which lazily acquires the token when any of the methods is first invoked?

PR #143, opened 2024-02-08
feat: switch to async firefall api

solaris007:


packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -90,24 +83,63 @@ export default class FirefallClient {
     this.log.debug(`${message}: took ${duration}ms`);
   }
 
-  async #apiCall(prompt) {
+  async #submitJob(prompt) {
     const body = JSON.stringify({
-      dialogue: { question: prompt },
-      llm_metadata: LLM_CONFIG,
+      input: prompt,
+      capability_name: 'gpt4_32k_completions_capability',
```
dzehnder, line 89: Shouldn't this be configurable?

iuliag (approved):

PR #152, opened 2024-02-13
feat: add configurations data access

alinarublea:
I am not sure if we need to handle specific configurations (ex: jobs -> adding a new job to the jobs configMap)
TODO: 

- [ ] IT tests


packages/spacecat-shared-data-access/test/it/db.test.js
```diff
@@ -63,6 +63,15 @@ function checkOrganization(organization) {
   schema.validate(organization);
 }
 
+function checkConfiguration(configuration) {
```
solaris007, line 66: should we use this check in the actual model ?

packages/spacecat-shared-data-access/docs/schema.json
```diff
@@ -525,6 +525,64 @@
           }
         }
       }
+    },
+    {
+      "TableName": "spacecat-services-configurations",
+      "KeyAttributes": {
+        "PartitionKey": {
+          "AttributeName": "GSI1PK",
```
solaris007, line 533: Should not be called `GSI1PK` as this would refer to a Global Secondary Index attribute in a main table used in a GSI. Call it `PK` for example.

packages/spacecat-shared-data-access/docs/schema.json
```diff
@@ -525,6 +525,64 @@
...
+        },
+        "SortKey": {
+          "AttributeName": "version",
+          "AttributeType": "S"
+        }
+      },
+      "NonKeyAttributes": [
+        {
+          "AttributeName": "queues",
+          "AttributeType": "S"
```
solaris007, line 544: i think this should be a map: `M`

packages/spacecat-shared-data-access/docs/schema.json
```diff
@@ -525,6 +525,64 @@
...
+        },
+        {
+          "AttributeName": "jobs",
+          "AttributeType": "S"
```
solaris007, line 548: i think this should be a list: `L`

solaris007 (approved):

PR #166, opened 2024-02-27
feat: add composeAuditURL to helpers

alinarublea:

## Related Issues

Thanks for contributing!


iuliag (commented):
This is the previous `getRUMUrl` function, no?
You should have a test case with a redirect to www as well, so that it's more clear from the test cases why someone would use the function and not just strip out the protocol.

iuliag (approved):

ekremney (approved):

PR #185, opened 2024-03-18
feat: add json to csv helper function

dzehnder:

## Related Issues
https://github.com/adobe/spacecat-api-service/issues/210
https://github.com/adobe/spacecat-api-service/pull/214

Thanks for contributing!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1491,6 +1491,20 @@
         "node": ">=v12.0.0"
       }
     },
+    "node_modules/@json2csv/formatters": {
+      "version": "7.0.6",
+      "resolved": "https://registry.npmjs.org/@json2csv/formatters/-/formatters-7.0.6.tgz",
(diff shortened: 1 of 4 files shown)


iuliag (approved):
🆒 
Wdyt about using in the test a JSON that contains different types of values, e.g. object, array, number, Boolean, or null?
It'd be good to see, with not additional configuration, what the output would be, so we know to use it in other areas where not all field values are strings.

ekremney (approved):

dzehnder:
> 🆒 Wdyt about using in the test a JSON that contains different types of values, e.g. object, array, number, Boolean, or null? It'd be good to see, with not additional configuration, what the output would be, so we know to use it in other areas where not all field values are strings.

Good point, thought about it but decided not to do it as it should be already tested in the library. But didn't think about usage examples, will add it 👍

PR #191, opened 2024-03-20
fix: don t expose rum domain key

alinarublea:

## Related Issues

Thanks for contributing!


solaris007:
@alinarublea i would add a test case that ensures that no rum api key is returned by the client

alinarublea:
@solaris007 This one is checking and the returned url here : https://github.com/adobe/spacecat-shared/pull/191/files#diff-8c38a9db780518d7c62aada503b3eb162e9bd9a40e255c24ed091cff1a2eba69R28 doesn t contain keys

solaris007 (approved):

Issue #193, opened 2024-03-20
Enable retrieval of a user's email address from IMS given an access token


blefebvre:
**Is your feature request related to a problem? Please describe.**
We need a way to fetch a user's email address given their IMS access token (which contains their IMS user ID).

**Describe the solution you'd like**
I would like a method that can be invoked with a single parameter, `getUserProfile(accessToken)`, which would return an object containing their email address as one of the fields.

**Describe alternatives you've considered**
Currently we deal directly with email addresses from the IMS group members API. However, to support the Cloud Manager use case, we need to adapt to support project collaboration channel requests via access token.

**Additional context**
n/a

PR #194, opened 2024-03-22
feat: Enable retrieval of a user's email address from IMS given an access token #193

blefebvre:
Ticket: #193

Proposed addition of a new method to the ImsClient, `getImsUserProfile(imsAccessToken)`.

Currently we deal directly with email addresses from the IMS group members API. However, to support the Cloud Manager use case, we need to adapt to support project collaboration channel requests via access token. This method will bridge the gap from IMS user token to a user's email address, which we can then use to send them a Slack invitation.

### DIFF
--- a/packages/spacecat-shared-ims-client/src/clients/ims-client.js
+++ b/packages/spacecat-shared-ims-client/src/clients/ims-client.js
@@ -330,14 +330,14 @@ export default class ImsClient {
         throw new Error(`IMS getImsUserProfile request failed with status: ${profileResponse.status}`);
       }
 
-      const profile = await profileResponse.json();
+      const { userId, email, roles } = await profileResponse.json();
 
       this.#logDuration('IMS getImsUserProfile request', startTime);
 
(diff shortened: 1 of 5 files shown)


packages/spacecat-shared-ims-client/src/clients/ims-client.js
```diff
@@ -297,4 +302,46 @@ export default class ImsClient {
...
+            Authorization: `Bearer ${imsAccessToken}`,
+          },
+        },
+      );
+
+      if (!profileResponse.ok) {
+        throw new Error(`IMS getImsUserProfile request failed with status: ${profileResponse.status}`);
+      }
+
+      const profile = await profileResponse.json();
```
solaris007, line 333: you could use destructuring:
```suggestion
const { userId, email, roles } = await profileResponse.json();
```

solaris007 (approved):
LGTM

solaris007 (approved):

PR #208, opened 2024-04-15
feat: API call for getting conversion data.

raji-unni:

## Related Issues

Thanks for contributing!


packages/spacecat-shared-rum-api-client/test/rum-api-client.test.js
```diff
@@ -205,4 +205,126 @@ describe('rum api client', () => {
...
+      })
+      .reply(200, JSON.stringify({
+        results: {
+          data: [{
+            checkpoint: 'convert',
+            source: 'get-a-demo',
+            target: null,
+            ids: 56,
+            pages: 3,
+            topurl: 'https://www.bamboohr.com/g2/',
```
alinarublea, line 228: I would replace all bamboohr reference, so we don t have any real customer url s in the tests, apart from that it looks good

raji-unni: Done.

alinarublea (approved):

PR #216, opened 2024-04-22
fix: ahrefs api call order_by

AndreiAlexandruParaschiv:
`sum_traffic_merged` can be used in the `order_by` only when we also set a `date_compared`, which we won't need/use


iuliag (approved):
Good find!

PR #219, opened 2024-04-22
feat: introduce google client

dzehnder:
This PR introduces the GoogleClient, designed to facilitate interactions with Google APIs for retrieving organic search data and listing sites. The class uses AWS Secrets Manager to securely manage Google OAuth credentials, ensuring secure and efficient API requests.


dzehnder:
relates to https://github.com/adobe/spacecat-shared/pull/219

packages/spacecat-shared-google-client/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+    return new GoogleClient(authClient, context.log);
+  }
+
+  constructor(authClient, log = console) {
+    this.authClient = authClient;
+    this.log = log;
+  }
+
+  async getOrganicSearchData(baseURL, startDate, endDate) {
```
ekremney, line 44: we could add `startRow` and `rowLimit` as params with default values 0 and 10. That way, this API could also provide a way to paginate

packages/spacecat-shared-google-client/src/index.js
```diff
@@ -0,0 +1,64 @@
...
+export default class GoogleClient {
+  static async createFrom(context) {
+    const authClient = new OAuth2Client(
+      context.env.GOOGLE_CLIENT_ID,
+      context.env.GOOGLE_CLIENT_SECRET,
+      context.env.GOOGLE_REDIRECT_URI,
+    );
+
+    authClient.setCredentials({
+      access_token: context.env.ACCESS_TOKEN,
```
ekremney, line 26: are access token and refresh token expected to come from env variables? 🤔 these should be short lived creds, unless we don't have a way to automatically rotate them, I guess we need to obtain these tokens freshly from google, not from env variables

dzehnder: the idea was to write a customer secret wrapper, that adds the secrets as env variables for the duration of the function execution.
Do you think this should be handled differently?

ekremney: > the idea was to write a customer secret wrapper, that adds the secrets as env variables for the duration of the function execution.

where the "customer secret wrapper" is expected to read those secrets from? request body or the query params? If that's the case, then we can get it from `context.data` or `context.params`

solaris007:
some improvements discussed f2f (checkauth method and input validation, version pinning)

solaris007 (approved):
LGTM

PR #223, opened 2024-05-02
feat: key events data access layer

ekremney:
Data layer impl. for https://github.com/adobe/spacecat-api-service/pull/247

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-data-access/src/index.d.ts
+++ b/packages/spacecat-shared-data-access/src/index.d.ts
@@ -130,6 +130,47 @@ export interface AuditConfig {
 }
 
 /**
+ * Represents a key event.
+ */
+export interface KeyEvent {
+  /**
+   * Retrieves the ID of the key event.
+   * @returns {string} The Key Event ID.
+   */
+  getId: () => string;
+
+  /**
+   * Retrieves the site id of the key event.
+   * @returns {string} site id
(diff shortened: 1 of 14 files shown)


packages/spacecat-shared-data-access/src/index.d.ts
```diff
@@ -129,6 +129,47 @@ export interface AuditConfig {
...
+   * Retrieves the creation timestamp of the key event.
+   * @returns {string} The creation timestamp.
+   */
+  getCreatedAt: () => string;
+
+  /**
+   * Retrieves the last update timestamp of the key event.
+   * @returns {string} The last update timestamp.
+   */
+  getUpdatedAt: () => string;
```
dzehnder, line 170: whats the purpose of having updatedAt for a Key Event if there is no way to update a key event? I would either add an update method to the API or remove this.

ekremney: `id`, `createdAt` and `updatedAt` come from the `Base` model. All data models have those fields by default since they all extend to the `Base` model: https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-data-access/src/models/base.js#L35

dzehnder (approved):

PR #239, opened 2024-05-23
feat: filter fetched organic search data from google by base url

dzehnder:
this PR introduces filtering for the organic search data requested from google.

Examples:

For the base URL `https://ust.com/en/insights`, we currently request the data for `https://www.ust.com/`, but it should be filtered to `https://www.ust.com/en/insights`.
For the base URL `https://sunstar.com`, currently all other subdomains are included, like `https://jp.sunstar.com`, however we only want `https://www.sunstar.com`

### DIFF
--- a/packages/spacecat-shared-google-client/src/index.js
+++ b/packages/spacecat-shared-google-client/src/index.js
@@ -14,6 +14,7 @@ import { google } from 'googleapis';
 import { OAuth2Client } from 'google-auth-library';
 import { GetSecretValueCommand, SecretsManagerClient } from '@aws-sdk/client-secrets-manager';
 import {
+  composeAuditURL,
   isArray,
   isInteger,
   isValidDate, isValidUrl,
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-google-client/src/index.js
```diff
@@ -108,6 +111,8 @@ export default class GoogleClient {
 
     await this.#refreshTokenIfExpired();
 
+    const auditUrl = await composeAuditURL(this.baseUrl);
```
iuliag, line 114: fyi: this will probably not work for customers with bot detection: https://cq-dev.slack.com/archives/C05A45JBP9N/p1715100249734359?thread_ts=1715068794.711109&cid=C05A45JBP9N

iuliag (approved):
LGTM 👍🏻 
With the case mentioned in the comment below.

ekremney (approved):

PR #252, opened 2024-06-05
feat: client for rum bundler

ekremney:
BREAKING CHANGE: With this change RUMAPIClient will not be using helix-run-query as datasource anymore

This PR introduces a breaking change. As a result of this, a new major version `2.0.0` of `spacecat-shared/rum-api-client` will be released.

The main change between major version is the datasource. `1.x` uses `helix-run-query` as datasource, whereas `2.x` will be using [rum-bundler API](https://github.com/adobe/helix-rum-bundler) as datasource. As a result of this change, `rum-api-client` will also crunch the data to calculate the insights because unlike `helix-run-query`, `rum-bundler` is serving raw data.

Two example queries to see how the data is crunched:
1. [cwv](https://github.com/adobe/spacecat-shared/blob/client-for-rum-bundler/packages/spacecat-shared-rum-api-client/src/functions/cwv.js)
2. [404](https://github.com/adobe/spacecat-shared/blob/client-for-rum-bundler/packages/spacecat-shared-rum-api-client/src/functions/404.js)

`spacecat-shared/rum-api-client@1.x` will be maintained at [rum-api-client@v1](https://github.com/adobe/spacecat-shared/tree/rum-api-client%40v1) branch until `helix-run-query` service is decomissioned.

## Related Issues

Thanks for contributing!


packages/spacecat-shared-rum-api-client/src/functions/cwv.js
```diff
@@ -0,0 +1,71 @@
...
+  return {
+    url,
+    lcp: quantile(CWVs['cwv-lcp'], 0.75) || null,
+    lcpCount: CWVs['cwv-lcp'].length,
+    cls: quantile(CWVs['cwv-cls'], 0.75) || null,
+    clsCount: CWVs['cwv-cls'].length,
+    inp: quantile(CWVs['cwv-inp'], 0.75) || null,
+    inpCount: CWVs['cwv-inp'].length,
+    ttfb: quantile(CWVs['cwv-ttfb'], 0.75) || null,
+    ttfbCount: CWVs['cwv-ttfb'].length,
```
langswei, line 50: Is this 75th percentile by date or across all dates spanning the given interval?

I ask because `id` is not guaranteed to be unique across dates.  The preferred grouping logic for events is unique combination of id+url+date.  Line 31 comment looks appropriate for a single day, but the quantile calls look appropriate for an interval of days.

langswei (approved):
Overall good.  Just left one comment related to aggregation of events which may not apply (you be the judge) and regardless would not be expected to have significant impact on the numbers -- although this should remind us to compare the numbers for similar features across the multiple apps in our RUM UI portfolio.

Other than that, RUM Explorer is timezone-aware so we may want to make Experience Success Studio timezone-aware at some point as well.  I don't know if that would be handled as changes in this app or not.

packages/spacecat-shared-rum-api-client/src/functions/cwv.js
ekremney: incorporated: https://github.com/adobe/spacecat-shared/pull/252/commits/53e2fa01d932163d66e7a299381c6241ca99bfd5

ekremney:
@langswei so far I have been comparing the numbers side by side with RUM explorer and I haven't seen difference except for the some LCP & CLS discrepancies which I let the guys know about: https://cq-dev.slack.com/archives/C07198KKQ07/p1717583673041269

But I don't know about the future, at some point we might diverge though. It will be kind of hard to keep them aligned especially given they have different requirements, one runs in browser the other one runs on server-side.

> Other than that, RUM Explorer is timezone-aware so we may want to make Experience Success Studio timezone-aware at some point as well. I don't know if that would be handled as changes in this app or not.

This is definitely something to keep in mind. For now, we are only using the `interval` param. It gets the data from `now-interval` until `now` regardless of the time zone.

packages/spacecat-shared-rum-api-client/src/common/rum-bundler-client.js
```diff
@@ -0,0 +1,89 @@
...
+    granularity = GRANULARITY.DAILY,
+    checkpoints = [],
+  } = opts;
+
+  if (!hasText(domain) || !hasText(domainkey)) {
+    throw new Error('Missing required parameters');
+  }
+
+  const multiplier = granularity.toUpperCase() === GRANULARITY.HOURLY ? ONE_HOUR : ONE_DAY;
+  const range = granularity.toUpperCase() === GRANULARITY.HOURLY ? interval * 24 : interval + 1;
```
solaris007, line 66: use constants here

solaris007 (approved):
bombastic

PR #257, opened 2024-06-08
feat: introduce organic keywords for ahrefs client and top keyword in top pages

dzehnder:
In context for [SITES-22601](https://jira.corp.adobe.com/browse/SITES-22601) (auto fix broken backlinks), we need access to the keywords of a page to determine the best possible alternate url (as suggested [by research](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=AEMSites&title=SEO+%7C+Broken+Backlinks+Auto-Suggest)) 

### Cost estimation for the API consumption:

- **Request per page:**
Worst Case:
Each row in the query costs 12 Units (_sum_traffic_: 10, _keyword_: 1, _best_position_url_: 1)
Using a limit of 200 rows per request: 2400 Units (12 x 200).
For 200 broken backlinks: **480k Units** (2400 x 200)

- **For import:**
How many keywords should be imported?
For example, page with 20k keywords (with traffic):
Getting all keywords: **240k Units** (12 x 20 000)
Getting top 2k keywords: **24k Units** (12 x 2000)

### DIFF
--- a/packages/spacecat-shared-ahrefs-client/src/index.js
+++ b/packages/spacecat-shared-ahrefs-client/src/index.js
@@ -192,9 +192,14 @@ export default class AhrefsAPIClient {
       output: 'json',
     };
     if (keywordFilter.length > 0) {
-      queryParams.where = JSON.stringify({
-        or: keywordFilter.map((keyword) => ({ field: 'keyword', is: ['iphrase_match', keyword] })),
-      });
+      try {
+        queryParams.where = JSON.stringify({
+          or: keywordFilter.map((keyword) => ({ field: 'keyword', is: ['iphrase_match', keyword] })),
+        });
+      } catch (e) {
+        this.log.error(`Error parsing keyword filter: ${e.message}`);
+        throw new Error(`Error parsing keyword filter: ${e.message}`);
+      }
     }
 
     return this.sendRequest('/site-explorer/organic-keywords', queryParams);
--- a/packages/spacecat-shared-ahrefs-client/src/index.js
+++ b/packages/spacecat-shared-ahrefs-client/src/index.js
@@ -174,4 +174,26 @@ export default class AhrefsAPIClient {
 
     return this.sendRequest('/site-explorer/metrics-history', queryParams);
   }
+
+  async getOrganicKeywords(url, country = 'us', keywordFilter = [], limit = 200) {
+    const queryParams = {
+      country,
+      date: new Date().toISOString().split('T')[0],
+      select: [
+        'keyword',
+        'sum_traffic',
+        'best_position_url',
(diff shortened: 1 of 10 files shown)


packages/spacecat-shared-ahrefs-client/src/index.js
```diff
@@ -174,4 +174,28 @@ export default class AhrefsAPIClient {
 
     return this.sendRequest('/site-explorer/metrics-history', queryParams);
   }
+
+  async getOrganicKeywords(url, country = 'us', keywordFilter = [], limit = 200) {
```
solaris007, line 178: missing input parameter validation

packages/spacecat-shared-ahrefs-client/src/index.js
```diff
@@ -174,4 +174,28 @@ export default class AhrefsAPIClient {
...
+        'best_position_url',
+      ].join(','),
+      order_by: 'sum_traffic:desc',
+      target: url,
+      limit: getLimit(limit, 2000),
+      mode: 'prefix',
+      output: 'json',
+    };
+    if (keywordFilter.length > 0) {
+      queryParams.where = JSON.stringify({
```
solaris007, line 194: this may throw

iuliag:
I'm thinking we might also want to try with just adding `top_keyword` to the top pages query https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-ahrefs-client/src/index.js#L122, as that would only increase the current cost of top pages with 1 unit per row (200 units for top pages of the whole site).

packages/spacecat-shared-ahrefs-client/src/index.js
iuliag: Do we need a country, what if this is a global site?
I wouldn't set the default to be `us`, but rather none.

iuliag (approved):
👍🏻 
With an observation to address below.

packages/spacecat-shared-ahrefs-client/src/index.js
dzehnder: yes the country is mandatory for this Ahrefs request because keywords are country specific.
For the Google API, this is not the case, there you can get them across all countries, which in some cases also creates confusing results (in this case needs to be filtered by subdomain or path, which results ultimately in the same issue).

I see two options for this:

- we extract the country from either url or hreflang tag
- we collect the keywords from the top-pages (already suggested by @iuliag through other channels). This might have some disadvantages as top-pages differ from keyword rankings. On the other hand, we can save an additional Ahrefs API request. We could also increase the limit in this case to import maybe top 2k pages for better accuracy

Issue #258, opened 2024-06-08
Import URL entities need a PENDING status


blefebvre:
**Is your feature request related to a problem? Please describe.**
- Currently, the import URL entities can be `RUNNING`, `COMPLETE`, or `FAILED`. A 4th status is necessary to indicate that a URL has been queued but processing of it has not yet begun.

**Describe the solution you'd like**
- Suggesting the addition of a `PENDING` status value for this purpose.

**Describe alternatives you've considered**
- `RUNNING` could be used, although I feel that this status is misleading for URLs that are not currently being processed.
- In terms of naming, `WAITING` or `QUEUED` are other other options.

PR #259, opened 2024-06-08
feat: Add PENDING status for import URL entities

blefebvre:
## Related Issues

- #258

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/importer/import-url.js
+++ b/packages/spacecat-shared-data-access/src/models/importer/import-url.js
@@ -12,12 +12,11 @@
 
 import { isValidUrl } from '@adobe/spacecat-shared-utils';
 import { Base } from '../base.js';
+import { ImportJobStatus } from './import-job.js';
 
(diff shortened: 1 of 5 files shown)


packages/spacecat-shared-data-access/src/models/importer/import-url.js
```diff
@@ -12,7 +12,13 @@
 
 import { isValidUrl } from '@adobe/spacecat-shared-utils';
 import { Base } from '../base.js';
-import { IMPORT_JOB_STATUS } from './import-job.js';
+
+export const IMPORT_URL_STATUS = {
```
solaris007, line 16: see comment here https://github.com/adobe/spacecat-api-service/pull/341/files#r1632614363

blefebvre: Good call @solaris007, this is much tidier on the API side:

```
import { ImportJobStatus } from '@adobe/spacecat-shared-data-access';
```

solaris007 (approved):

Issue #268, opened 2024-06-18
Import URL entities need a REDIRECT status, and additional properties for reporting


blefebvre:
**Description**
In order to support generating the import-report.xlsx file for an import job, we need the ability to track if a URL was skipped because it was a redirect, or if it failed for another reason.

Screenshot from an example import-report included below:

<img width="1176" alt="Screenshot 2024-06-18 at 2 05 45 PM" src="https://github.com/adobe/spacecat-shared/assets/1048207/08d8c91b-8b64-41f6-acf9-d4742626ad0d">

To support this, I'm proposing that we add the following properties to the import-url entities, in addition to the existing `status`:

```
reason: the reason that the import of a URL failed (could be the redirect destination in the case of a redirect)
path: absolute path to the resource that is being imported for the given URL
file: resulting path and filename of the imported .docx file
```

An additional status `REDIRECT` will also be added to `ImportUrlStatus` to facilitate generation of this report.

PR #269, opened 2024-06-18
fix: #268 Import URL entities need a REDIRECT status, and additional properties for reporting

blefebvre:
## Related Issues

- #268 

In order to support generating the import-report.xlsx file for an import job, we need the ability to track if a URL was skipped because it was a redirect, or if it failed for another reason.

Screenshot from an example import-report included below:

image

To support this, I'm proposing that we add the following properties to the import-url entities, in addition to the existing status:

```
reason: the reason that the import of a URL failed (could be the redirect destination in the case of a redirect)
path: absolute path to the resource that is being imported for the given URL
file: resulting path and filename of the imported .docx file
```

An additional status REDIRECT will also be added to `ImportUrlStatus` to facilitate generation of this report.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22086,7 +22086,7 @@
     },
     "packages/spacecat-shared-data-access": {
       "name": "@adobe/spacecat-shared-data-access",
-      "version": "1.29.0",
+      "version": "1.29.1",
       "license": "Apache-2.0",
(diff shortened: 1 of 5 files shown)


swetabar (approved):
lgtm

solaris007 (approved):

PR #270, opened 2024-06-18
feat: overwrite broken-backlinks

AndreiAlexandruParaschiv:
https://jira.corp.adobe.com/browse/SITES-22728 - Overwrite Broken Backlinks

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/config.js
+++ b/packages/spacecat-shared-data-access/src/models/site/config.js
@@ -31,6 +31,7 @@ export const configSchema = Joi.object({
       Joi.object({
         disabled: Joi.boolean().optional(),
         excludedURLs: Joi.array().items(Joi.string()).optional(),
+        manualOverwrites: Joi.array().items(Joi.string()).optional(),
       }).unknown(true),
     ).unknown(true),
   }).unknown(true),
--- a/packages/spacecat-shared-data-access/src/models/site/config.js
+++ b/packages/spacecat-shared-data-access/src/models/site/config.js
@@ -31,7 +31,10 @@ export const configSchema = Joi.object({
       Joi.object({
         disabled: Joi.boolean().optional(),
         excludedURLs: Joi.array().items(Joi.string()).optional(),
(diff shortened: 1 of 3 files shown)


packages/spacecat-shared-data-access/src/models/site/config.js
```diff
@@ -31,6 +31,7 @@ export const configSchema = Joi.object({
       Joi.object({
         disabled: Joi.boolean().optional(),
         excludedURLs: Joi.array().items(Joi.string()).optional(),
+        manualOverwrites: Joi.array().items(Joi.string()).optional(),
```
iuliag, line 34: This would not be an array of string, but rather an array of objects, with broken URL -> URL to overwrite it with.

packages/spacecat-shared-data-access/test/unit/models/site/audit-config-type.test.js
```diff
@@ -118,4 +118,20 @@ describe('AuditConfigType Tests', () => {
       expect(dynamoItem.excludedURLs).to.eql(urls);
     });
   });
+
+  describe('updateManualOverwrites Method', () => {
+    it('updates the manualOverwrites array', () => {
+      const auditConfigType = AuditConfigType();
+      const newManualOverwrites = ['brokenTargetURL', 'targetURL'];
```
iuliag, line 125: See the comment above, we need to pass both the URL to overwrite with and the URL that is broken and should be overwritten, in pairs.

iuliag (commented):
For overwrites, we need a list of pairs of URL to overwrite and URL to overwrite with, see example in: [Overwrite the suggested URL for a broken backlink
](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=AEMSites&title=SEO+%7C+Broken+Backlinks+Manual+and+Auto-Fix#SEO|BrokenBacklinksManualandAutoFix-OverwritethesuggestedURLforabrokenbacklink)

alinarublea (approved):

iuliag (approved):

Issue #181, opened 2024-03-12
Replace uses of `sqsEventAdapter` function with spacecat-shared-utils version


blefebvre:
With https://github.com/adobe/spacecat-shared/pull/128 being merged, search SpaceCat repos for copies of the `sqsEventAdapter` function and replace with the version in spacecat-shared-utils.

From the [usage instructions](https://github.com/adobe/spacecat-shared/tree/main/packages/spacecat-shared-utils#sqs-event-adapter):

```
import { sqsEventAdapter } from '@adobe/spacecat-shared-utils';

// ...

export const main = wrap(run)
  .with(dataAccess)
  .with(sqsEventAdapter) // Note this line
  .with(sqs)
  .with(secrets)
  .with(helixStatus);
```

PR #275, opened 2024-06-27
fix: sqs adapter records access

solaris007:
Fixes possible undefined errors. Blocks #181


packages/spacecat-shared-rum-api-client/test/index.test.js
```diff
@@ -31,6 +33,9 @@ describe('RUMAPIClient', () => {
   });
 
   it('runs the query', async () => {
+    nock('https://rum.fastly-aem.page')
```
solaris007, line 36: unrelated, but encountered a test failure with the test accessing the live API and getting a `401`

alinarublea (approved):

ekremney (approved):

PR #282, opened 2024-07-04
feat: utils deepEqual

solaris007:


ekremney (approved):
LGTM. 

Just one question: how does this work with objects with functions? For example, the models in shared lib often have functions such as `getBaseURL` etc.

PR #289, opened 2024-07-12
feat: audit top pages for language mismatch (SITES-22690)

rachel-kittens:
Added `variant.js` to `packages/spacecat-shared-rum-api-client/src/functions` to create languages & regions insights from RUM bundles.

## Related Issues
[SITES-22690](https://jira.corp.adobe.com/browse/SITES-22690)

Thanks for contributing!


.github/workflows/main.yaml
```diff
@@ -17,6 +17,8 @@ jobs:
           node-version: '20.x'
       - name: Install Dependencies
         run: npm ci
+      - name: Check ESLint version
```
solaris007, line 20: was this for debugging, could it be removed? eslint is locked <9 via renovate.

packages/spacecat-shared-rum-api-client/src/functions/variant.js
```diff
@@ -0,0 +1,125 @@
...
+        type3: {
+          preferredLanguages: {}, // Type 3 mismatches
+        },
+      },
+      regions: {}, // Tracks the count of events for each region
+    };
+    languageInsights.push(languageObject);
+  }
+  return languageObject;
+}
```
solaris007, line 37: for readability i'd use an early return if the lang object is defined, as well as checking if it `isObject` (import from shared utils).
```suggestion
function getOrCreateLanguageObject(languageInsights, language) {
  let languageObject = languageInsights.find((l) => l.language === language);
  
  if (isObject(languageObject)) {
    return languageObject;
  }
  
  languageObject = {
    language,
    count: 0,
    mismatches: {
      type1: { preferredLanguages: {} }, // Type 1 mismatches
      type2: { preferredLanguages: {} }, // Type 2 mismatches
      type3: { preferredLanguages: {} }, // Type 3 mismatches
    },
    regions: {},  // Tracks the count of events for each region
  };

  languageInsights.push(languageObject);
  return languageObject;
}
```

packages/spacecat-shared-rum-api-client/src/functions/variant.js
```diff
@@ -0,0 +1,125 @@
...
+          preferredLanguages = target.split(',').map((lang) => lang.trim());
+        }
+
+        if (source === 'page-language') {
+          pageLanguage = target;
+        }
+
+        if (source === 'user-region') {
+          userRegion = target;
+        }
```
solaris007, line 61: i'd suggest an else-if here to prevent other conditions from being evaluated:
```suggestion
        if (source === 'preferred-languages') {
          preferredLanguages = target.split(',').map((lang) => lang.trim());
        } else if (source === 'page-language') {
          pageLanguage = target;
        } else if (source === 'user-region') {
          userRegion = target;
        }
```

packages/spacecat-shared-rum-api-client/src/functions/variant.js
```diff
@@ -0,0 +1,125 @@
...
+      }
+    }
+
+    if (pageLanguage) {
```
solaris007, line 65: i would use an early continue as well:
```js
    if (!pageLanguage) {
      continue;
    }
 ```

solaris007 (approved):
please add some jsdoc or comments to the handler to understand input / output format and what's going on.

otherwise LGTM, only cosmetic comments.

packages/spacecat-shared-rum-api-client/src/functions/variant.js
rachel-kittens: When I make this change, I got an eslint error that says "no-continue"

PR #293, opened 2024-07-17
SITES-22537 feat: Pages with steady decline of conversion

Dereje24:

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-rum-api-client/src/functions/oppty.js
+++ b/packages/spacecat-shared-rum-api-client/src/functions/oppty.js
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2024 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import  { FlatBundle } from '../common/flat-bundle.js';
+import { pageviewsByUrl } from '../common/aggregateFns.js';
+
+function handler(bundles) {
+    const collectPageViews = bundles => pageviewsByUrl(bundles).filter(page => page.views >= 5000);
+    const pageviewUrls = collectPageViews(bundles).map(row => row.url);
+
+    return FlatBundle.fromArray(bundles)
+        .filter(row => pageviewUrls.includes(row.url) && row.checkpoint === 'convert')
+        .groupBy('url')
+        .map((groupedByUrl) => {
+            const { url, items } = groupedByUrl;
+            const totalViews = pageviews.find((row) => row.url === url).views;
+            const totalConversions = items.reduce((acc, cur) => acc + cur.weight, 0);
+            return {
+              url,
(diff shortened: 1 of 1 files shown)


packages/spacecat-shared-rum-api-client/src/functions/oppty.js
```diff
@@ -0,0 +1,39 @@
...
+        url,
+        conversionRate: totalConversions / totalViews,
+      };
+      })
+      .filter((row) => row.conversionRate < 0.05);
+}
+    
+export default {
+  handler,
+  checkpoints: ['pageview', 'convert'],
```
ekremney, line 37: there is no checkpoint called `pageview`. Pageviews are calculated by the root level `weight` property of the RUM bundles. Each bundle has "weight" amount of pageviews.

`convert` is not enough itself to calculate the conversion. We also use `click` and `formsubmit`

packages/spacecat-shared-rum-api-client/src/functions/oppty.js
```diff
@@ -0,0 +1,39 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { FlatBundle } from '../common/flat-bundle.js';
+import { pageviewsByUrl } from '../common/aggregateFns.js';
+
+function handler(bundles) {
+  const collectPageViews = bundles => pageviewsByUrl(bundles).filter(page => page.views >= 5000);
```
ekremney, line 17: `pageviewsByUrl` function return an `object`, key of with is url and the value of which is number of pageviews

ekremney (commented):
we dont have a way to calculate conversions accurately. Maybe we should change `Pages with steady decline of conversion rate` to `Pages with steady decline of click-through-rate`. What do you think @ramboz ?

packages/spacecat-shared-rum-api-client/src/functions/opportunities.js
```diff
@@ -69,11 +69,19 @@ function handler(bundles) {
   //     data[bundle.url][weekKey].metrics.push({ selector: source, ctr });
   // }
     // Initialize an object to hold the selectors and their counts
-    let globalSelectors = {};
+    const globalSelectors = {};
 
     for (const bundle of bundles) {
       // Initialize a Set to hold the unique selectors for this bundle
-      let uniqueSelectors = new Set();
+      const uniqueSelectors = new Set();
```
ramboz: I'd move the declaration outside of the loop so you re-use the same variable and avoid allocation of hundreds of different ones.
```js
    let uniqueSelectors;
    for (const bundle of bundles) {
      // Initialize a Set to hold the unique selectors for this bundle
      uniqueSelectors = new Set();
```

packages/spacecat-shared-rum-api-client/src/functions/opportunities.js
```diff
@@ -69,11 +69,19 @@ function handler(bundles) {
...
+
+      let totalClicks = 0;
```
ramboz: Same here, I'd move the declaration outside of the loop

packages/spacecat-shared-rum-api-client/src/functions/opportunities.js
```diff
@@ -83,13 +91,13 @@ function handler(bundles) {
 
       // Iterate over the unique selectors and increment their count in the global selectors object
       for (const source of uniqueSelectors) {
-        globalSelectors[source] = (globalSelectors[source] || 0) + 1;
+        const count = uniqueSelectors[source] || 0;
```
ramboz: Again here as well

PR #306, opened 2024-07-29
feat: Add API Key data attributes

swetabar:
## Related Issues
[SITES-23627](https://jira.corp.adobe.com/browse/SITES-23627)

Setup the dynamo DB table with the following attributes:
```
{  
"id": "298987DC-DEBC-45D2-9054-10053BF15F5E",  
"key": "(encrypted with crypto package)",  
"name": "Bruce's key",  
"imsUserId": "AD5E1C3F650AAAF30A495FD9@b13a68e1650aa8c6495ffe.e",  
"imsOrgId": "86DF666465096C340A495E89@AdobeOrg",
"createdAt": "2024-07-16T18:59:14.435Z",
"expiresAt": "2024-10-16T18:59:14.435Z",
"revokedAt": null,
"scopes": [    
     {      "name": "import.write",
            "domains": [        
                "https://adobe.com"      
            ]    
      }  
      ]
}
```

### DIFF
--- a/packages/spacecat-shared-data-access/src/dto/api-key.js
+++ b/packages/spacecat-shared-data-access/src/dto/api-key.js
@@ -21,7 +21,7 @@ export const ApiKeyDto = {
      *          imsOrgId: *, expiresAt: *}}
      */
   toDynamoItem: (apiKey) => ({
-    key: apiKey.getKey(),
+    hashedKey: apiKey.getHashedKey(),
     name: apiKey.getName(),
     imsUserId: apiKey.getImsUserId(),
     imsOrgId: apiKey.getImsOrgId(),
@@ -38,7 +38,7 @@ export const ApiKeyDto = {
      */
   fromDynamoItem: (dynamoItem) => {
     const apiKeyData = {
-      key: dynamoItem.key,
+      hashedKey: dynamoItem.hashedKey,
       name: dynamoItem.name,
       imsUserId: dynamoItem.imsUserId,
       imsOrgId: dynamoItem.imsOrgId,
--- a/packages/spacecat-shared-data-access/src/dto/api-key.js
+++ b/packages/spacecat-shared-data-access/src/dto/api-key.js
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2024 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { createApiKey } from '../models/api-key.js';
+
+export const ApiKeyDto = {
+
+  /**
+     * Converts an ApiKey object into a DynamoDB item.
+     * @param apiKey
+     * @returns {{createdAt: *, name: *, imsUserId, scopes, revokedAt, key: *,
+     *          imsOrgId: *, expiresAt: *}}
+     */
+  toDynamoItem: (apiKey) => ({
+    key: apiKey.getKey(),
+    name: apiKey.getName(),
+    imsUserId: apiKey.getImsUserId(),
+    imsOrgId: apiKey.getImsOrgId(),
+    createdAt: apiKey.getCreatedAt(),
+    expiresAt: apiKey.getExpiresAt(),
+    revokedAt: apiKey.getRevokedAt(),
+    scopes: apiKey.getScopes(),
+  }),
+
+  /**
+     * Converts a DynamoDB item into an ApiKey object.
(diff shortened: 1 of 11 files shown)


packages/spacecat-shared-data-access/src/dto/api-key.js
```diff
@@ -0,0 +1,53 @@
...
+export const ApiKeyDto = {
+
+  /**
+     * Converts an ApiKey object into a DynamoDB item.
+     * @param apiKey
+     * @returns {{createdAt: *, name: *, imsUserId, scopes, revokedAt, key: *,
+     *          imsOrgId: *, expiresAt: *}}
+     */
+  toDynamoItem: (apiKey) => ({
+    key: apiKey.getKey(),
```
blefebvre, line 24: Should this be `hashedKey` (or similar) so it does not get misinterpreted as the actual key?

packages/spacecat-shared-data-access/src/index.d.ts
```diff
@@ -534,6 +534,57 @@ export interface ImportUrl {
...
+ * Represents an API Key entity.
+ */
+export interface ApiKey {
+    /**
+     * Retrieves the ID of the API Key.
+     */
+    getId: () => string;
+
+    /**
+     * Retrieves the key of the API Key.
```
blefebvre, line 547: "hashed key value" instead of "key"?

packages/spacecat-shared-data-access/src/index.d.ts
```diff
@@ -722,6 +773,12 @@ export interface DataAccess {
       jobId: string,
       status: string,
     ) => Promise<ImportUrl[]>;
+  getApiKeyById: (
+      id: string,
```
blefebvre, line 777: Will we be querying by id, or by the hashed key value computed from the `x-api-key` header?

packages/spacecat-shared-data-access/src/models/api-key.js
```diff
@@ -0,0 +1,70 @@
...
+  if (hasText(newState.expiresAt) && !isIsoDate(newState.expiresAt)) {
+    throw new Error(`expiresAt should be a valid ISO 8601 string: ${newState.expiresAt}`);
+  }
+
+  if (hasText(newState.revokedAt) && !isIsoDate(newState.revokedAt)) {
+    throw new Error(`revokedAt should be a valid ISO 8601 string: ${newState.revokedAt}`);
+  }
+
+  if (!Array.isArray(newState.scopes)) {
+    throw new Error(`Invalid scopes: ${newState.scopes}`);
```
blefebvre, line 66: Should we have a list of valid scopes to check the values of this array against?

blefebvre (commented):
Thinking about the access pattern: querying the DB for a valid (non-expired or revoked) key based on the value provided in the `x-api-key` header, hashed by the crypto package once received.

packages/spacecat-shared-data-access/src/models/api-key.js
swetabar: I'm planning to check if the `name` attribute exists for now.

swetabar: Or do we want to check the accepted values of the name attribute based on an environment variable?

blefebvre: > Or do we want to check the accepted values of the name attribute based on an environment variable?

I don't think this is necessary, the scopes could be defined in code. Worst case we would need a small deployment to update the list.

solaris007 (approved):

packages/spacecat-shared-data-access/src/index.d.ts
```diff
@@ -534,6 +534,57 @@ export interface ImportUrl {
...
+export interface ApiKey {
+    /**
+     * Retrieves the ID of the API Key.
+     */
+    getId: () => string;
+
+    /**
+     * Retrieves the hashed key value of the API Key.
+     */
+    getHashedKey: () => string;
```
blefebvre, line 549: Thinking out loud: any reason that `hashedKey` couldn't be the partition key? Could avoid the need for a GSI.

packages/spacecat-shared-data-access/src/models/api-key.js
```diff
@@ -0,0 +1,95 @@
...
+
+import {
+  hasText, isIsoDate, isObject, isValidUrl,
+} from '@adobe/spacecat-shared-utils';
+import { Base } from './base.js';
+
+const scopeNames = ['sites.read_all', 'sites.write_all', 'organizations.read_all', 'organizations.write_all',
+  'audits.read_all', 'audits.write_all', 'imports.read', 'imports.write', 'imports.read_all'];
+const ApiKey = (data) => {
+  const self = Base(data);
```
blefebvre, line 21:
```suggestion
  'audits.read_all', 'audits.write_all', 'imports.read', 'imports.write', 'imports.read_all'];

const ApiKey = (data) => {
  const self = Base(data);
```

blefebvre (approved):
👍  no need to make the change to partition key, just something to consider.

packages/spacecat-shared-data-access/src/index.d.ts
swetabar: It could be the partition key. Then we'll need to move from the current model implementation. We won't inherit from Base anymore (so no default Id).

PR #308, opened 2024-07-30
fix: work-around for bamboos custom exp engine

ekremney:

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-rum-api-client/src/functions/traffic-acquisition.js
+++ b/packages/spacecat-shared-rum-api-client/src/functions/traffic-acquisition.js
@@ -52,6 +52,17 @@ function transformFormat(trafficSources) {
 /*
  * throw-away code for a single customer who customized the experimentation engine
  * this code will be removed once they start using the default exp engine
+ *
+ * this function fetches experiment manifests, then merges variants data into controls data
+ *
+ * ie:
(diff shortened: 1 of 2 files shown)


ramboz (approved):

packages/spacecat-shared-rum-api-client/src/functions/traffic-acquisition.js
```diff
@@ -47,8 +48,81 @@ function transformFormat(trafficSources) {
...
+    if (!acc[key]) acc[key] = [];
+    if (variantPaths.includes(new URL(cur.url).pathname)) {
+      // eslint-disable-next-line no-param-reassign
+      cur.url = new URL(controlPath, cur.url).href;
+    }
+    acc[key].push(cur);
+    return acc;
+  }, {});
+
+  const merged = Object.entries(byIdAndPath).flatMap(([, v]) => {
```
rpapani, line 110: @ekremney please check if this handles when the control split is 0 as well

ekremney: `mergeBundlesWithSameId` does not take splits into account. It is used to map variants to control group. See the jsdocs

PR #312, opened 2024-08-01
feat: Introduce scoped API keys

swetabar:
We are proposing the introduction of scoped API keys, which have access only to the specific scopes that are set on the API key entity in the data layer. This access will be enforced from the controller of each endpoint with code that follows this pattern:

_Updated code sample with latest PR changes, now that checkScopes will throw an error._

```
const { auth } = context;

try {
  auth.checkScopes(['imports.write']);
} catch (error) {
  throw new ErrorWithStatusCode('API key is missing required scopes', 401);
}
```

.. which would result in the `x-error` header on the response being set with an intuitive error message.

Initially, we are planning to add this new `ScopedApiKeyHandler` _after_ the existing `LegacyApiKeyHandler` in the `authHandlers` array for the following reasons:

- This will mean that the existing user & admin API keys will not be affected by this change
- It also means that we will not incur any additional DB cost for querying the API key table on each request
- This ordering of handlers is key: only when the Legacy handler cannot authenticate an API key will the new Scoped handler be invoked

We would eventually like to move the legacy API keys over to the "scoped" model too ([SITES-23628](https://jira.corp.adobe.com/browse/SITES-23628)), but not in this initial rollout. The first consumer of this new handler will be Import as a Service, which will enable us to get rid of the x-import-api-key header which we currently require, and avoid us having to share the Spacecat API key with clients of the service — they will now each get their own. 

It also opens the door to future consumers of the API (think Studio) being able to request an API key for a project, which would be limited to a specific domain and also expire automatically after some time.

## Related Issues
- [SITES-23631](https://jira.corp.adobe.com/browse/SITES-23631) - Introduce new ScopedApiKeyHandler, to be invoked after the existing LegacyApiKeyHandler
- [SITES-23630](https://jira.corp.adobe.com/browse/SITES-23630) - Check scope access for a given API Key
- [SITES-23403](https://jira.corp.adobe.com/browse/SITES-23403) - Spacecat API key management upgrade

Initial set of scopes:
```
sites.[read_all|write_all]
organizations.[read_all|write_all]
audits.[read_all|write_all]
imports.[read|{*}read_all{*}|write] - read_all implies access to all import jobs, including those that were not created by the given API key
```


packages/spacecat-shared-http-utils/src/auth/auth-wrapper.js
```diff
@@ -48,6 +53,44 @@ export function authWrapper(fn, opts = {}) {
       return new Response('Unauthorized', { status: 401 });
     }
 
+    if (!context.auth) {
+      context.auth = {
+        hasScopes: async (scopes) => {
```
blefebvre, line 58: To keep auth-wrapper focused, can the hasScopes implementation be moved to its own file?

I like the simple interface from a usability perspective (`hasScopes(scopes)`), so perhaps this function remains which calls the underlying impl in its own file? This would avoid having to pass additional args (or the entire context):

```
export async function hasScopes(scopes, apiKey) // where apiKey is the context.pathInfo?.headers['x-api-key'] val
```

packages/spacecat-shared-http-utils/src/auth/auth-wrapper.js
```diff
@@ -48,6 +53,44 @@ export function authWrapper(fn, opts = {}) {
...
+          // Pull the api-key from x-api-key header
+          const apiKeyFromHeader = context.pathInfo?.headers['x-api-key'];
+
+          if (!hasText(apiKeyFromHeader)) {
+            return new Response('Unauthorized', { status: 401 });
```
blefebvre, line 63: I would like to see this function always return a boolean, such that the individual controllers can decide how to handle a falsey result. 

Logging the reason for the `false` should happen, so we can diagnose issues on our side. 

On the other hand, we could consider returning a small object that would have details on the reason that hasScopes was false, ie:

```
{
  result: true
}

// or, in the falsey case
{
  result: false,
  reason: 'API key has been revoked'
}

// or, in the case of a missing scope (organization.read_all)
{
  result: false,
  reason: 'API key is missing the ["organization.read_all"] scope(s) required for this resource'
}
```

What do you think?

packages/spacecat-shared-http-utils/src/auth/auth-wrapper.js
```diff
@@ -48,6 +53,44 @@ export function authWrapper(fn, opts = {}) {
...
+          }
+
+          // Generate a hash of the API Key
+          const hashedKey = hashWithSHA256(apiKeyFromHeader);
+
+          if (!context.dataAccess) {
+            log.error('Data access required');
+            return new Response('Server error', { status: 500 });
```
blefebvre, line 71: This should throw a new error IMO.

packages/spacecat-shared-http-utils/src/auth/auth-wrapper.js
```diff
@@ -48,6 +53,44 @@ export function authWrapper(fn, opts = {}) {
...
+          }
+
+          // Fetch the api-key record from data access layer
+          const apiKeyRecord = await context.dataAccess.getApiKeyByHashedKey(hashedKey);
+
```
blefebvre, line 76: Case needs to be handed: no API key found for this hash.

packages/spacecat-shared-http-utils/src/auth/auth-wrapper.js
```diff
@@ -48,6 +53,44 @@ export function authWrapper(fn, opts = {}) {
...
+          // Check that the api key has not expired or been revoked,
+          const now = new Date().toISOString();
+          if (apiKeyRecord.getExpiresAt() < now) {
+            log.error(`API key has expired, name = ${apiKeyRecord.getName()}`);
```
blefebvre, line 80: Let's log the ID as well.

packages/spacecat-shared-http-utils/test/auth/auth-wrapper.test.js
```diff
@@ -101,4 +117,52 @@ describe('auth wrapper', () => {
...
+    }), context);
+    const val = await context.auth.hasScopes(['scope1', 'scope2']);
+    expect(val).to.be.true;
+  });
+
+  it('returns unauthorized when api key is missing', async () => {
+    await action(new Request('https://space.cat/', {
+      headers: { 'x-api-key': 'test' },
+    }), context);
+    context.pathInfo.headers['x-api-key'] = '';
```
blefebvre, line 133: Can you remove the header on line 131 instead of setting it to an empty string here?

packages/spacecat-shared-http-utils/test/auth/auth-wrapper.test.js
```diff
@@ -101,4 +117,52 @@ describe('auth wrapper', () => {
...
+      headers: { 'x-api-key': 'test' },
+    }), context);
+    const val = await context.auth.hasScopes(['scope1', 'scope2']);
+    expect(val).to.be.true;
+  });
+
+  it('returns unauthorized when api key is missing', async () => {
+    await action(new Request('https://space.cat/', {
+      headers: { 'x-api-key': 'test' },
+    }), context);
```
blefebvre, line 132: I'm not sure these Requests are necessary for the unit test below — unless, are they just adding things to the context?

If they are necessary, could the code be written once in `beforeEach`?

blefebvre (commented):
Nice to see this coming together!

packages/spacecat-shared-http-utils/src/auth/auth-wrapper.js
swetabar: We need the dataAccess object as well for the apiKey query.

packages/spacecat-shared-http-utils/src/auth/auth-info.js
```diff
@@ -50,4 +51,32 @@ export default class AuthInfo {
...
+  withScopes(scopes) {
+    this.scopes = scopes;
+    return this;
+  }
+
+  getScopes() { return this.scopes; }
+
+  getProfile() { return this.profile; }
+
+  getReason() { return this.reason; }
```
atopper, line 79: Will the reason be leaked to the caller (user) at all?  Could any 'reason' help a hacker to do their evil work?

packages/spacecat-shared-data-access/src/models/api-key.js
```diff
@@ -82,12 +82,14 @@ export const createApiKey = (data) => {
       throw new Error(`Scope name is not part of the pre-defined scopes: ${scope.name}`);
     }
 
-    if (hasText(scope.domains) && !Array.isArray(scope.domains)) {
-      throw new Error(`Scope domains should be an array: ${scope.domains}`);
-    }
-    for (const domain of scope.domains) {
-      if (!isValidUrl(domain)) {
-        throw new Error(`Invalid domain: ${domain}`);
+    if (scope.domains) {
```
atopper, line 85: Looks like this changes the behaviour (_depending on `spacecat-shared-utils/hasText` I suppose_).  If `scope.domains` is `null/undefined` the call will not throw an error like it did before (right?).
Is this intentional?

blefebvre: This code was added by us recently for this feature - not in use yet by others. There was a logic issue that this PR is correcting: `scope.domains` is optional, and in the case where it’s not set the for loop was throwing an error (“scope.domains is not Iterable”). It is not a text property so `hasText` was not the correct check, either.

packages/spacecat-shared-http-utils/src/auth/handlers/scoped-api-key.js
```diff
@@ -0,0 +1,69 @@
...
+    const apiKeyFromHeader = headers['x-api-key'];
+    if (!hasText(apiKeyFromHeader)) {
+      return null;
+    }
+
+    // Keys are stored by their hash, so we need to hash the key to look it up
+    const hashedKey = hashWithSHA256(apiKeyFromHeader);
+    const apiKeyEntity = await dataAccess.getApiKeyByHashedKey(hashedKey);
+
+    if (!apiKeyEntity) {
```
swetabar, line 42: Do we need to check for empty braces `{}` as well? If the DB returns `{}` in case the hashedKey is not found?

blefebvre: I had assumed `getApiKeyByHashedKey` would return `null` if the API key entity is not found, similar to:

```
export const getImportJobByID = async (dynamoClient, config, log, id) => {
  const item = await dynamoClient.getItem(
    config.tableNameImportJobs,
    { id },
  );
  return item ? ImportJobDto.fromDynamoItem(item) : null;
};
```

.. but see that this is not the case for `getApiKeyByHashedKey`. I think this should be updated to return null when not found for consistency — WDYT @swetabar?

packages/spacecat-shared-http-utils/src/auth/auth-wrapper.js
```diff
@@ -43,7 +44,15 @@ export function authWrapper(fn, opts = {}) {
     }
 
     try {
-      await authenticationManager.authenticate(request, context);
+      const authInfo = await authenticationManager.authenticate(request, context);
+
+      // Add a helper function to the context for checking scoped API keys.
+      // authInfo is available at context.attributes.authInfo.
+      if (!context.auth) {
```
solaris007, line 51: use `isObject` or even `isNonEmptyObject`

packages/spacecat-shared-http-utils/src/auth/check-scopes.js
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Check if the given AuthInfo has the requested scopes.
+ * @param {Array<string>} scopes - The scopes required for the request
+ * @param {AuthInfo} authInfo - Authentication state for the current request
+ * @param {Logger} log - Logger
+ * @return {{hasScopes: boolean, reason?: string}}
+ */
+export function checkScopes(scopes, authInfo, log) {
+  if (!authInfo) {
```
solaris007, line 21: use `isObject` or even `isNonEmptyObject`

packages/spacecat-shared-http-utils/src/auth/check-scopes.js
```diff
@@ -0,0 +1,43 @@
...
+
+  if (missingScopes.length > 0) {
+    log.error(`API key with ID: ${authInfo.getProfile()?.api_key_id} does not have required scopes. It's missing: ${missingScopes.join(',')}`);
+    return {
+      hasScopes: false,
+      reason: `API key is missing the [${missingScopes.join(',')}] scope(s) required for this resource`,
+    };
+  }
+
+  return { hasScopes: true };
```
solaris007, line 42: instead of returning complex objects, i'd suggest this method throws an error, with the reason in the error message.

solaris007 (approved):
looks great, just minor comments

packages/spacecat-shared-http-utils/src/auth/auth-info.js
blefebvre: Will go with `Missing required scopes` on the controller side (api-service), while logging the detail to Error level. cc @swetabar

packages/spacecat-shared-http-utils/src/auth/check-scopes.js
blefebvre: Description now includes updated sample code:

```
try {
  auth.checkScopes(['imports.write']);
} catch (error) {
  throw new ErrorWithStatusCode('API key is missing required scopes', 401);
}
```

PR #313, opened 2024-08-02
chore: fixing unit tests

rpapani:
mistakenly used `getMilliseconds` instead of `getTime`, so the test was flaky


ekremney (approved):
thanks!

PR #325, opened 2024-08-15
fix: refactor hashed api key and query to retrieve API key

swetabar:
## Related Issues
[SITES-23633](https://jira.corp.adobe.com/browse/SITES-23633)

- Refactor hashedKey to hashedApiKey for consistency with spacecat-api-service
- Modify the unit tests spacecat-shared-http-utils to remove dependency on spacecat-data-shared methods
- Modify the query - `getApiKeyByHashedApiKey` to return the first element of an array

### DIFF
--- a/packages/spacecat-shared-http-utils/test/auth/auth-wrapper.test.js
+++ b/packages/spacecat-shared-http-utils/test/auth/auth-wrapper.test.js
@@ -55,7 +55,7 @@ describe('auth wrapper', () => {
       dataAccess: {},
     };
     mockApiKey = createApiKey({
-      hashedKey: '372c6ba5a67b01a8d6c45e5ade6b41db9586ca06c77f0ef7795dfe895111fd0b',
+      hashedApiKey: '372c6ba5a67b01a8d6c45e5ade6b41db9586ca06c77f0ef7795dfe895111fd0b',
       name: 'Test API key name',
       scopes: [
         {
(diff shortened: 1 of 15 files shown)


packages/spacecat-shared-http-utils/test/auth/auth-wrapper.test.js
```diff
@@ -54,16 +55,20 @@ describe('auth wrapper', () => {
       },
       dataAccess: {},
     };
-    mockApiKey = createApiKey({
-      hashedKey: '372c6ba5a67b01a8d6c45e5ade6b41db9586ca06c77f0ef7795dfe895111fd0b',
-      name: 'Test API key name',
-      scopes: [
+    mockApiKey = {
+      getId: () => '1C4ED8DE-8ECD-42E1-9812-AF34082FB1B4',
+      getHashedKey: () => '372c6ba5a67b01a8d6c45e5ade6b41db9586ca06c77f0ef7795dfe895111fd0b',
```
blefebvre, line 60: Just confirming: These mocked functions (`getHashedKey` in particular) will be removed in the next PR?

swetabar: Yes - correct. The next set of modifications will be for `spacecat-shared-utils`, so that it uses the refactored getHashedApiKey from `spacecat-shared-data-access`.

PR #326, opened 2024-08-15
feat: introduce url inspect in google client

dzehnder:
In context of [SITES-24087](https://jira.corp.adobe.com/browse/SITES-24087), introducing URL inspect to the Google Client

### DIFF
--- a/packages/spacecat-shared-google-client/src/index.js
+++ b/packages/spacecat-shared-google-client/src/index.js
@@ -20,6 +20,7 @@ import {
   isValidDate, isValidUrl,
   resolveCustomerSecretsName,
 } from '@adobe/spacecat-shared-utils';
+import { fetch as httpFetch } from './utils.js';
 
 export default class GoogleClient {
   static async createFrom(context, baseURL) {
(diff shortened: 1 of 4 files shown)


packages/spacecat-shared-google-client/src/index.js
```diff
@@ -148,6 +149,34 @@ export default class GoogleClient {
     }
   }
 
+  async urlInspect(url) {
+    if (!isValidUrl(url)) {
+      throw new Error('Error inspecting URL: Invalid URL format');
```
iuliag, line 154: It may be helpful logging the url received as input.

packages/spacecat-shared-google-client/src/index.js
```diff
@@ -148,6 +149,34 @@ export default class GoogleClient {
...
+        Authorization: `Bearer ${this.authClient.credentials.access_token}`,
+      },
+      body: JSON.stringify({
+        inspectionUrl: url,
+        siteUrl: this.siteUrl,
+      }),
+    });
+
+    if (!response.ok) {
+      throw new Error('Error inspecting URL');
```
iuliag, line 174: It may be helpful logging which url had an error while inspecting and what status was returned.

packages/spacecat-shared-google-client/src/index.js
```diff
@@ -148,6 +149,34 @@ export default class GoogleClient {
...
+    }
+
+    return response.json();
```
iuliag, line 177: Since parsing the response may throw, it's probably helpful if we catch that early and log a parsing error, before we bubble up.

packages/spacecat-shared-google-client/test/index.test.js
```diff
@@ -371,4 +371,55 @@ describe('GoogleClient', () => {
...
+
+    afterEach(() => {
+      nock.cleanAll();
+      sinon.restore();
+    });
+
+    it('should inspect a valid URL', async () => {
+      const url = 'https://example.com/page';
+      const apiEndpoint = 'https://searchconsole.googleapis.com';
+      const mockResponse = { inspectionResult: 'some result' };
```
iuliag, line 388: Unit tests also play the role of documentation, so it would be useful to have here a response that has the schema of the actual response this API will return.

iuliag (approved):

PR #329, opened 2024-08-15
feat: [oppty] Pages with high click amount on non-interacting element

rpapani:
TODO: Unit tests


packages/spacecat-shared-rum-api-client/src/functions/rageclick.js
```diff
@@ -0,0 +1,179 @@
...
+        // eslint-disable-next-line no-param-reassign
+        rageClickInstances[url].pageViews = pageData[url].pageViews;
+        // eslint-disable-next-line no-param-reassign
+        rageClickInstances[url].samples = pageData[url].samples;
+      }
+    }
+  }
+}
+
+function getRageClickOpporunities(rageClickInstances) {
```
ekremney, line 92: typo here

packages/spacecat-shared-rum-api-client/src/functions/rageclick.js
```diff
@@ -0,0 +1,179 @@
...
+      (acc, metric) => acc + metric.percentage,
+      0,
+    ) / opportunity.metrics.length;
+    opportunity.trackedPageKPIValue = avgRageClickPercentage;
+    opportunities.push(opportunity);
+  }
+  return opportunities;
+}
+
+function handler(bundles) {
```
ekremney, line 126: _personal opinion and not blocking_: I don't know how much memory overhead it would add but using functional js interfaces `map`, `filter` etc. would increase the readability

ekremney (approved):

packages/spacecat-shared-rum-api-client/src/functions/rageclick.js
rpapani: as I was building a global structure `rageClickInstances`, felt like map wasn't best fit and running multiple operations of map/filter, will iterate the array more than once, so I thought its better to create the structure I want with single iteration.

PR #336, opened 2024-08-19
fix: add try-catch block to authentication manager

swetabar:
## Related Issues

Add try-catch block to authentication-manager to try and figure out the issue with ScopedApiKeyHandler

### DIFF
--- a/packages/spacecat-shared-http-utils/src/auth/authentication-manager.js
+++ b/packages/spacecat-shared-http-utils/src/auth/authentication-manager.js
@@ -45,8 +45,13 @@ export default class AuthenticationManager {
     for (const handler of this.handlers) {
       this.log.debug(`Trying to authenticate with ${handler.name}`);
 
-      // eslint-disable-next-line no-await-in-loop
-      const authInfo = await handler.checkAuth(request, context);
+      let authInfo;
+      try {
+        // eslint-disable-next-line no-await-in-loop
+        authInfo = await handler.checkAuth(request, context);
+      } catch (error) {
+        this.log.error(`Failed to authenticate with ${handler.name}: ${error.message}`);
+      }
 
       if (isObject(authInfo)) {
         this.log.info(`Authenticated with ${handler.name}`);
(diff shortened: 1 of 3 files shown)


packages/spacecat-shared-http-utils/src/auth/authentication-manager.js
```diff
@@ -45,8 +45,13 @@ export default class AuthenticationManager {
...
       this.log.debug(`Trying to authenticate with ${handler.name}`);
 
-      // eslint-disable-next-line no-await-in-loop
-      const authInfo = await handler.checkAuth(request, context);
+      let authInfo;
+      try {
+        // eslint-disable-next-line no-await-in-loop
+        authInfo = await handler.checkAuth(request, context);
+      } catch (error) {
+        this.log.error(`Failed to authenticate with ${handler.name}: ${error.message}`);
```
blefebvre, line 53: To log the stack trace as well:
```suggestion
        this.log.error(`Failed to authenticate with ${handler.name}:`, error);
```

packages/spacecat-shared-http-utils/src/auth/handlers/scoped-api-key.js
```diff
@@ -46,6 +46,7 @@ export default class ScopedApiKeyHandler extends AbstractHandler {
       this.log(`No API key entity found in the data layer for the provided API key: ${apiKeyFromHeader}`, 'error');
       return null;
     }
+    this.log(`API Key entity: ${apiKeyEntity}`, 'debug');
```
blefebvre, line 49: Should this object be stringified?

blefebvre (approved):
Minor item with the log statement.

Is an additional test case needed to hit the `catch` block?

packages/spacecat-shared-http-utils/src/auth/handlers/scoped-api-key.js
swetabar: Removing these logs as we might not need them anymore.

swetabar:
> Minor item with the log statement.
> 
> Is an additional test case needed to hit the `catch` block?

Yes. Added the additional test case to fix the test errors.

PR #337, opened 2024-08-19
fix: experimentation IT tests failing intermittently

rpapani:
- tests are using the `dataAccess.getSites()[0]` depending on the order sites are returned, this could be anything, so I've used the getSiteByBaseURL to get the same site object always
- `removes organization` it is removing the first organization, so most likely this is removing randomly, so getSites isn't returning the sites from this removed organization, which was causing the failure. So, I've moved the `removes organization` to the last in the file


solaris007 (approved):

solaris007:
merging to unblock - thanks @rpapani

PR #343, opened 2024-08-22
fix: add property to include specific urls in the site config

dzehnder:
Upon several naming discussions, we decided to change the name from product detail pages so something more generic.


iuliag (approved):
👍🏻 
No test needs to change when renaming this property?

dzehnder:
> 👍🏻 No test needs to change when renaming this property?

the change is tested via the config schema

PR #356, opened 2024-09-03
fix: [Importer] Specify the list of supported features

bhellema:
Fixing up some documentation, and define a constant to specify what features the importer supports.

## Related Issues
[SITES-24836](https://jira.corp.adobe.com/browse/SITES-24836)

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/importer/import-job.js
+++ b/packages/spacecat-shared-data-access/src/models/importer/import-job.js
@@ -10,160 +10,162 @@
  * governing permissions and limitations under the License.
  */
 
+/**
+ * @typedef {import('../../').ImportJob} ImportJob
+ */
+
 import {
-  hasText, isIsoDate, isValidUrl, isObject, isString, isNumber, isInteger,
+  hasText, isIsoDate, isNumber, isInteger, isValidUrl, isString, isObject,
 } from '@adobe/spacecat-shared-utils';
 import { Base } from '../base.js';
-
-export const ImportJobStatus = {
-  RUNNING: 'RUNNING',
-  COMPLETE: 'COMPLETE',
-  FAILED: 'FAILED',
-};
+import { ImportJobStatus, ImportOptions } from './import-constants.js';
 
 /**
- * Creates a new ImportJob object.
- *
- * @param {Object} importJobData - The data for the ImportJob object.
- * @returns {ImportJob} The new ImportJob object.
+ * Creates a new ImportJob.
+ * @param data - The data for the ImportJob object.
+ * @returns {Readonly<ImportJob>} The new ImportJob object.
  */
-const ImportJob = (data) => {
-  const self = Base(data);
+const ImportJobEntry = (data) => {
+  const importJob = Base(data);
 
-  self.getBaseURL = () => self.state.baseURL;
-  self.getHashedApiKey = () => self.state.hashedApiKey;
-  self.getOptions = () => self.state.options;
-  self.getStartTime = () => self.state.startTime;
-  self.getEndTime = () => self.state.endTime;
-  self.getDuration = () => self.state.duration;
-  self.getStatus = () => self.state.status;
-  self.getUrlCount = () => self.state.urlCount;
-  self.getSuccessCount = () => self.state.successCount;
-  self.getFailedCount = () => self.state.failedCount;
-  self.getImportQueueId = () => self.state.importQueueId;
-  self.getInitiatedBy = () => self.state.initiatedBy;
+  // generate get methods for all properties of our base object
+  importJob.getBaseURL = () => importJob.state.baseURL;
+  importJob.getHashedApiKey = () => importJob.state.hashedApiKey;
+  importJob.getOptions = () => importJob.state.options;
+  importJob.getStartTime = () => importJob.state.startTime;
+  importJob.getEndTime = () => importJob.state.endTime;
+  importJob.getDuration = () => importJob.state.duration;
+  importJob.getStatus = () => importJob.state.status;
+  importJob.getUrlCount = () => importJob.state.urlCount;
+  importJob.getSuccessCount = () => importJob.state.successCount;
+  importJob.getFailedCount = () => importJob.state.failedCount;
+  importJob.getImportQueueId = () => importJob.state.importQueueId;
+  importJob.getInitiatedBy = () => importJob.state.initiatedBy;
 
   /**
-     * Updates the end time of the ImportJob.
-     * @param {string} endTime - The new end time.
-     * @returns {ImportJob} The updated ImportJob object.
-     */
-  self.updateEndTime = (endTime) => {
-    if (!isIsoDate(endTime)) {
+   * Updates the state of the ImportJob.
+   * @param key - The key to update.
+   * @param value - The new value.
+   * @param validator - An optional validation function to use before updating the value.
+   * @returns {ImportJob} The updated ImportJob object.
+   */
+  const updateState = (key, value, validator) => {
+    if (validator && typeof validator === 'function') {
+      validator(value);
+    }
+
+    importJob.state[key] = value;
+    importJob.touch();
+
+    return importJob;
+  };
+
+  /**
+   * Updates the end time of the ImportJob.
+   * @param {string} endTime - The new end time.
+   */
+  importJob.updateEndTime = (endTime) => updateState('endTime', endTime, (value) => {
+    if (!isIsoDate(value)) {
       throw new Error(`Invalid end time during update: ${endTime}`);
     }
-
-    self.state.endTime = endTime;
-    self.touch();
-
-    return self;
-  };
+  });
 
   /**
-     * Updates the duration of the ImportJob.
-     * @param {number} duration - The new duration.
-     * @returns {ImportJob} The updated ImportJob object.
-     */
-  self.updateDuration = (duration) => {
-    if (!isNumber(duration)) {
-      throw new Error(`Invalid duration during update: ${duration}`);
+   * Updates the duration of the ImportJob.
+   * @param {number} duration - The new duration.
+   */
+  importJob.updateDuration = (duration) => updateState('duration', duration, (value) => {
+    if (!isNumber(value)) {
+      throw new Error(`Invalid duration during update: ${value}`);
     }
-
-    self.state.duration = duration;
-    self.touch();
-
-    return self;
-  };
+  });
 
   /**
-     * Updates the status of the ImportJob.
-     * @param {string} status - The new status.
-     * @returns {ImportJob} The updated ImportJob object.
-     */
-  self.updateStatus = (status) => {
-    if (!Object.values(ImportJobStatus).includes(status)) {
-      throw new Error(`Invalid Import Job status during update: ${status}`);
+   * Updates the status of the ImportJob.
+   * @param {string} status - The new status.
+   */
+  importJob.updateStatus = (status) => updateState('status', status, (value) => {
+    if (!Object.values(ImportJobStatus).includes(value)) {
+      throw new Error(`Invalid Import Job status during update: ${value}`);
     }
-
-    self.state.status = status;
-    self.touch();
-
-    return self;
-  };
+  });
 
   /**
    * Updates the Url count of the ImportJob
    * @param {number} urlCount - The new url count.
-   * @returns {ImportJob} The updated ImportJob object.
    */
-  self.updateUrlCount = (urlCount) => {
-    if (!isInteger(urlCount)) {
+  importJob.updateUrlCount = (urlCount) => updateState('urlCount', urlCount, (value) => {
+    if (!isInteger(value)) {
       throw new Error(`Invalid url count during update: ${urlCount}`);
     }
-
-    self.state.urlCount = urlCount;
-    self.touch();
-
-    return self;
-  };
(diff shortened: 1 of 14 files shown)


packages/spacecat-shared-data-access/src/models/importer/import-job.js
```diff
@@ -21,148 +21,150 @@ export const ImportJobStatus = {
...
-     * @returns {ImportJob} The updated ImportJob object.
-     */
-  self.updateEndTime = (endTime) => {
-    if (!isIsoDate(endTime)) {
-      throw new Error(`Invalid end time during update: ${endTime}`);
+   * Updates the state of the ImportJob.
+   * @param key - The key to update.
+   * @param value - The new value.
+   * @param validator - An optional validation function to use before updating the value.
+   */
```
swetabar, line 55: We could add `@returns {ImportJob}` here.

packages/spacecat-shared-data-access/src/models/importer/import-job.js
```diff
@@ -21,148 +21,150 @@ export const ImportJobStatus = {
...
-     * Updates the failed count of the ImportJob.
-     * @param {number} failedCount - The new failed count.
-     * @returns {ImportJob} The updated ImportJob object.
-     */
-  self.updateFailedCount = (failedCount) => {
-    if (!isInteger(failedCount)) {
-      throw new Error(`Invalid failed count during update: ${failedCount}`);
+   * Updates the failed count of the ImportJob.
+   * @param {number} failedCount - The new failed count.
+   * @returns {{ImportJob}} The updated ImportJob object.
```
swetabar, line 120: Maybe we can remove line 120 from here as we will include it in line 55?

packages/spacecat-shared-data-access/src/models/importer/import-job.js
```diff
@@ -21,148 +21,150 @@ export const ImportJobStatus = {
   FAILED: 'FAILED',
 };
 
+export const ImportOptions = {
+  ENABLE_JAVASCRIPT: 'enableJavascript',
```
swetabar, line 25: We check for `pageLoadTimeout` in the content-scraper : https://github.com/adobe/spacecat-content-scraper/blob/main/src/handlers/abstract-handler.js#L270. Is that something we could like to add in as well?

packages/spacecat-shared-data-access/src/models/importer/import-url.js
```diff
@@ -21,78 +21,72 @@ export const ImportUrlStatus = {
...
 
   /**
-   * Updates the status of the ImportUrl
+   * Updates the state of the ImportJob.
+   * @param key - The key to update.
+   * @param value - The new value.
+   * @param validator - An optional validation function to use before updating the value.
+   * The validator can return false to indicate that the value isn't worth throwing an exception,
+   * but continue to use the previous value.
    */
```
swetabar, line 47: We could add `@returns ImportUrl` here.

packages/spacecat-shared-data-access/src/models/importer/import-url.js
```diff
@@ -21,78 +21,72 @@ export const ImportUrlStatus = {
...
-  self.setStatus = (status) => {
-    if (!Object.values(ImportUrlStatus).includes(status)) {
-      throw new Error(`Invalid Import URL status during update: ${status}`);
+  const updateState = (key, value, validator) => {
+    if (validator && typeof validator === 'function') {
+      // a validator can return true or false to indicate if the value is valid
+      // however if a validator throws an error, it is considered critical and invalid.
+      if (!validator(value)) {
```
swetabar, line 52: Do we need to consider a similar check for import-job?

swetabar (commented):
Looks good, Ben! Thanks for the validation checks~
I added a few comments / questions.

packages/spacecat-shared-data-access/src/models/importer/import-job.js
bhellema: Is the pageLoadTimeout a user provided property or one of our default properties that is automatically added by us?

packages/spacecat-shared-data-access/src/models/importer/import-url.js
bhellema: It doesn't seem to be the same use case from what I saw.  In the ImportUrl scenario the use of hasText, was used to validate the input, but does not throw an error.  In the case of ImportJob, any missing or invalid data did throw an error, so the use case to continue if the value was false wasn't there.  So I kept the existing behaviour not to shake the tree to much.

packages/spacecat-shared-data-access/src/models/importer/import-job.js
swetabar: It has a default value if not provided by the user. But, its value could be provided by the user - based on the current helix-importer-ui tool.

packages/spacecat-shared-data-access/src/models/importer/import-job.js
```diff
@@ -10,161 +10,163 @@
...
-  self.getStartTime = () => self.state.startTime;
-  self.getEndTime = () => self.state.endTime;
-  self.getDuration = () => self.state.duration;
-  self.getStatus = () => self.state.status;
-  self.getUrlCount = () => self.state.urlCount;
-  self.getSuccessCount = () => self.state.successCount;
-  self.getFailedCount = () => self.state.failedCount;
-  self.getImportQueueId = () => self.state.importQueueId;
-  self.getInitiatedBy = () => self.state.initiatedBy;
+const ImportJobEntry = (data) => {
```
blefebvre, line 28: Is this change worth it? Do all the JSDocs (like line 26 above) need to be adjusted to use this new name?

packages/spacecat-shared-data-access/src/models/importer/import-job.js
```diff
@@ -10,161 +10,163 @@
...
- * @returns {ImportJob} The new ImportJob object.
+ * @param {Object} data - The data for the ImportJob object.
+ * @returns {Readonly<ImportJob>} The new ImportJob object.
+ * @throws {Error} If any of the required fields are invalid.
  */
 export const createImportJob = (data) => {
+  // Define a list of data type validators for each import option
+  const ImportOptionTypeValidator = {
+    [ImportOptions.ENABLE_JAVASCRIPT]: (value) => {
+      if (![true, false, 'true', 'false'].includes(value)) {
```
blefebvre, line 152: Can `isBoolean` be used instead? Should never be a string.

packages/spacecat-shared-data-access/src/models/importer/import-job.js
```diff
@@ -177,17 +179,29 @@ export const createImportJob = (data) => {
...
+      .filter((key) => !Object.values(ImportOptions).includes(key));
+
+    if (invalidOptions.length > 0) {
+      throw new Error(`Invalid options: ${invalidOptions}`);
+    }
+
+    // validate each option for it's expected data type
+    Object.keys(newState.options).forEach((key) => {
+      if (ImportOptionTypeValidator[key]) {
+        ImportOptionTypeValidator[key](data.options[key]);
```
blefebvre, line 201: Nice to have this validation for `options`.

packages/spacecat-shared-data-access/src/models/importer/import-url.js
```diff
@@ -10,89 +10,82 @@
...
- *
- * @param {Object} importUrlData
- * @returns {ImportUrl}
+ * Creates a new ImportUrl object.
+ * @param {Object} data - The data for the ImportUrl object.
+ * @return {ImportUrl} The new ImportUrl object.
  */
-const ImportUrl = (data) => {
-  const self = Base(data);
+const ImportUrlEntry = (data) => {
```
blefebvre, line 26: Same question here. What does the `Entry` suffix buy us?

packages/spacecat-shared-data-access/src/service/import-job/accessPatterns.js
```diff
@@ -40,12 +45,12 @@ export const getImportJobsByDateRange = async (dynamoClient, config, log, startD
...
 
 /**
- * Get Import Job by ID
+ * Get Import Job by ID.
  * @param {DynamoClient} dynamoClient
  * @param {Object} config
  * @param {Logger} log
  * @param {string} id
- * @returns {Promise<ImportJobDto> | null}
+ * @return {ImportJob | null}
```
blefebvre, line 53: Technically it does return a Promise.

packages/spacecat-shared-data-access/src/service/import-job/accessPatterns.js
```diff
@@ -56,12 +61,12 @@ export const getImportJobByID = async (dynamoClient, config, log, id) => {
...
 
 /**
- * Get Import jobs by status
+ * Get Import jobs by status.
  * @param {DynamoClient} dynamoClient
  * @param {Object} config
  * @param {Logger} log
  * @param {string} status
- * @returns {Promise<ImportJobDto[]>}
+ * @return {Array<ImportJob>} - The ImportJobs with the given status.
```
blefebvre, line 69: Does the removal of `Promise` for these async functions affect anything downstream?

blefebvre (commented):
Nice to have the import options validation in place. Few comments to consider.

packages/spacecat-shared-data-access/src/models/importer/import-job.js
bhellema: I was wondering about how the values are being initialized when the request comes in.  Is there a jackson layer that is converting the body of the json into the correct types?  What if someone sends in a { enableJavascript: "true" } vs { enableJavascript: true }.

blefebvre: That happens in https://github.com/adobe/helix-shared/tree/main/packages/helix-shared-body-data

packages/spacecat-shared-data-access/src/models/importer/import-url.js
bhellema: I was finding that the IDE wasn't able to resolve the ImportUrl type, if this function was called ImportUrl, it didn't know to look at the type definition.  

Being that the ImportUrlEntry is essentially a helper function that isn't exposed from this module, I just added `Entry`, where it maybe more accurate to call it `construct` function maybe?

blefebvre: I see. Perhaps the `Builder` suffix would work?

packages/spacecat-shared-data-access/src/models/importer/importer.js
```diff
@@ -0,0 +1,18 @@
...
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+import { ImportJobStatus, ImportUrlStatus, ImportOptions } from './import-constants.js';
+
+export {
```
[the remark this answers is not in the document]
bhellema: I was trying to isolate (separation of concerns) the import service operations to the rest of the project by providing our own barrel file.  Essentially it make it cleaner and more readable when it comes to others importing them.  That said, the main barrel file in this project is the index.js up one level, but providing multiple barrel files allows us to provide a nicer import in the index.js file. 

We could then simply update the index.js to `export * './models/importer/importer.js';`, rather than doing the following `import { ImportJobStatus, ImportUrlStatus, ImportOptions } from './models/importer/importer.js';` and then `export { ImportJobStatus, ImportUrlStatus, ImportOptions }`, this then allows us one central location for what we want to expose.  

What's your thoughts on this approach?

packages/spacecat-shared-data-access/src/service/import-job/accessPatterns.js
bhellema: I guess eh, I'll update that my bad.

packages/spacecat-shared-data-access/src/service/import-job/accessPatterns.js
bhellema: I think that it was wrong for me to remove it.

packages/spacecat-shared-data-access/src/models/importer/importer.js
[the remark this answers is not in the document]
blefebvre: Good with that. I hadn't heard the term "barrel file" before.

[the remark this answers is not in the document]
bhellema: Oh I think I came across another interesting use case.  In the api-service code base there is a unit test that mocks the createImportJob API and it digs into the code like this:
```
import { createImportJob } from '@adobe/spacecat-shared-data-access/src/models/importer/import-job.js';
```
By exporting the createImportJob from the src/models/importer/index.js (<-- i would rename the importer.js to index.js), we can export this module for testing and mocking it, but not export it from the main index.js barrel file.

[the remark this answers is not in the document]
blefebvre: > we can export this module for testing and mocking it

I do prefer the readability of your import over the absolute file path used currently 👍

PR #366, opened 2024-09-13
feat: update redirects

chicharr:
Feature to update redirects spreadsheet for an Edge Delivery Services website.
Receives a list of new redirects, validates them, ensuring they have the right format, and that they will not duplicate or create cycles with the existing redirect rules, and appends them to the existing rules in the spreadsheet.

### DIFF
--- a/packages/spacecat-shared-content-client/src/clients/content-client.js
+++ b/packages/spacecat-shared-content-client/src/clients/content-client.js
@@ -170,7 +170,7 @@ const removeRedirectLoops = (currentRedirects, newRedirects, log) => {
     }
   });
   if (newRedirects.length !== noCycleRedirects.length) {
-    this.log.info(`Removed ${newRedirects.length - noCycleRedirects.length} redirect loops`);
+    log.info(`Removed ${newRedirects.length - noCycleRedirects.length} redirect loops`);
   }
   return noCycleRedirects;
 };
--- a/packages/spacecat-shared-content-client/src/clients/content-client.js
+++ b/packages/spacecat-shared-content-client/src/clients/content-client.js
@@ -158,7 +158,7 @@ const removeRedirectLoops = (currentRedirects, newRedirects, log) => {
   const noCycleRedirects = [];
   currentRedirects.forEach((r) => redirectsGraph.addEdge(r.from, r.to));
   if (hasCycle(redirectsGraph)) {
-    throw new Error('Redirect cycle detected');
+    throw new Error('Redirect cycle detected in current redirects');
   }
   newRedirects.forEach((r) => {
     redirectsGraph.addEdge(r.from, r.to);
--- a/packages/spacecat-shared-content-client/src/clients/content-client.js
+++ b/packages/spacecat-shared-content-client/src/clients/content-client.js
@@ -12,6 +12,7 @@
 
 import { createFrom as createContentSDKClient } from '@adobe/spacecat-helix-content-sdk';
 import { hasText, isObject } from '@adobe/spacecat-shared-utils';
+import { Graph, hasCycle } from 'graph-data-structure';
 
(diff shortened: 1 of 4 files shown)


packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -97,6 +98,68 @@ const validateMetadata = (metadata) => {
...
+    .forEach((redirectRule) => redirectsSet.add(redirectRule));
+
+  const newRedirectsClean = [];
+  newRedirects.forEach((redirectRule) => {
+    const { from, to } = redirectRule;
+    const strRedirectRule = `${from}:${to}`;
+    if (!redirectsSet.has(strRedirectRule)) {
+      redirectsSet.add(strRedirectRule);
+      newRedirectsClean.push(redirectRule);
+    } // TODO: report duplicated entry??
```
alinarublea, line 149: Here I would log the duplicate entry

packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -97,6 +98,68 @@ const validateMetadata = (metadata) => {
...
+  });
+  return newRedirectsClean;
+};
+
+const detectRedirectLoops = (currentRedirects, newRedirects) => {
+  const redirectsGraph = new Graph();
+  [...currentRedirects, ...newRedirects].forEach((r) => redirectsGraph.addEdge(r.from, r.to));
+  if (hasCycle(redirectsGraph)) {
+    // TODO: report cycle path??
+    throw new Error('Redirect cycle detected');
```
alinarublea, line 159: In this case I see we are not adding any of the entries. Could we check after each edge if there is a cycle and just not add that specific edge ?

alinarublea (approved):
Except the small comments all the rest looks good to me

packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -97,6 +98,80 @@ const validateMetadata = (metadata) => {
...
+
+    if (redirect.from === redirect.to) {
+      throw new Error('Redirect from and to paths must be different');
+    }
+  }
+};
+
+const removeDuplicatedRedirects = (currentRedirects, newRedirects, log) => {
+  const redirectsSet = new Set();
+  currentRedirects.map(({ from, to }) => `${from}:${to}`)
```
dipratap, line 139: Can be simplified to
```
const redirectsSet = new Set(
    currentRedirects.map(({ from, to }) => `${from}:${to}`)
);
```

packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -97,6 +98,80 @@ const validateMetadata = (metadata) => {
...
+
+    if (!hasText(redirect.from)) {
+      throw new Error('Redirect must have a valid from path');
+    }
+
+    if (!hasText(redirect.to)) {
+      throw new Error('Redirect must have a valid to path');
+    }
+
+    if (!redirect.from.startsWith('/')) {
```
dipratap, line 123: Can we use a regex to validate the from and to endpoints?

packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -192,4 +269,50 @@ export default class ContentClient {
...
+    if (cleanNewRedirects.length === 0) {
+      this.log.info('No valid redirects to update');
+      return;
+    }
+    const noCycleRedirects = removeRedirectLoops(currentRedirects, cleanNewRedirects, this.log);
+    if (noCycleRedirects.length === 0) {
+      this.log.info('No valid redirects to update');
+      return;
+    }
+    if (noCycleRedirects.length !== cleanNewRedirects.length) {
```
dipratap, line 307: Should we log the removed loops in the removeRedirectLoops method itself?

dipratap (approved):

PR #368, opened 2024-09-13
feat: SITES-24835:[Importer] Tracing: logging infrastructure to support tracing an import job across all Lambdas

codevoyage:
### PR Context: Distributed Tracing with JobId for Bulk Import Jobs

This PR introduces a unified tracking mechanism across multiple Lambda invocations during bulk import jobs. By adding a `jobId` that is automatically injected into log entries, debugging and tracking become more straightforward.

**Flow Overview**  
Each import job triggers several Lambda functions. Every log entry in the bulk import flow needs to include a consistent `jobId`, enabling comprehensive tracking across all execution points. The `jobId` is added via the `logWrapper`, making it available without manual intervention.

**Wrapper Function for JobId Injection**  
The logWrapper reads the `jobId` from SQS messages and injects it into log statements. This enhances the logs with the necessary context, simplifying tracing and troubleshooting with platforms like Coralogix.

Example log entry:
```json
{
  '0': 'Content Processor received a message. Processing results: {"processingType":"","scrapeResults":[]}',
  jobId: 'a07a9c23-c17f-4c13-8241-40bbc397d676'
}
```

This approach ensures consistent logging, reducing manual effort and the potential for errors.

**Next Steps**  
Once merged, the `logWrapper`, when included in Lambdas, will ensure all relevant logs across Lambdas are tagged with the `jobId`, improving traceability for import jobs.

Relevant Issues:
- **[SITES-24835](https://jira.corp.adobe.com/browse/SITES-24835)** - Distributed Tracing: Logging Infrastructure for Import Jobs Across Lambdas

---

## Related Issues
https://jira.corp.adobe.com/browse/SITES-24835


packages/spacecat-shared-utils/src/index.js
```diff
@@ -38,6 +38,8 @@ export {
 export { sqsWrapper } from './sqs.js';
 export { sqsEventAdapter } from './sqs.js';
 
+export { logWrapper } from './log-wrapper.js';
```
blefebvre, line 41: Is `jobLogWrapper` too verbose? This wrapper is pretty specific to event-driven function invocation involving messages that include a job ID.

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -0,0 +1,55 @@
...
+
+        return fn(message, context);
+      }
+
+      const { jobId } = message;
+
+      // Enhance the log object to include jobId in each log statement
+      context.contextualLog = {
+        info: (...args) => {
+          log.info({ jobId, ...args });
```
blefebvre, line 38: Can you share a screenshot from your testing of how the resulting log entries are rendered on Coralogix? 

Is leveraging this functionality to put together the "complete" log stream as simple as searching for `jobId=<UUID>` in Coralogix? 🤞

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -0,0 +1,55 @@
...
+ * @returns {function(object, object): Promise<Response>} A new function that, when invoked,
+ *  will automatically include the `jobId` in all log statements within the context.
+ */
+export function logWrapper(fn) {
+  return async (message, context) => {
+    const { log } = context;
+
+    if (log) {
+      if (!message || !message.jobId) {
+        log.debug('Missing jobId, hence it will not be included in log messages.');
```
blefebvre, line 27: Wondering about the implication of making this an `error` level log instead, and then alerting on it? Will this wrapper be used (in the 3 initial target Lambdas) where `jobId` will not be present in the message?

```suggestion
        log.debug('Unable to determine jobId from the provided message. As a result, jobId will not be included in log messages.');
```

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -0,0 +1,55 @@
...
+ * @param {function} fn - The original function that will be wrapped.
+ * @returns {function(object, object): Promise<Response>} A new function that, when invoked,
+ *  will automatically include the `jobId` in all log statements within the context.
+ */
+export function logWrapper(fn) {
+  return async (message, context) => {
+    const { log } = context;
+
+    if (log) {
+      if (!message || !message.jobId) {
```
blefebvre, line 26:
```suggestion
      if (!isObject(message) || !hasText(message.jobId)) {
```

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -0,0 +1,55 @@
...
+        log.debug('Missing jobId, hence it will not be included in log messages.');
+        context.contextualLog = log;
```
blefebvre, line 28: Is `jobLog` no good? It is really job specific. I may have missed a discussion with @bhellema on this.

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -0,0 +1,55 @@
...
+        },
+        error: (...args) => {
+          log.error({ jobId, ...args });
+        },
+        debug: (...args) => {
+          log.debug({ jobId, ...args });
+        },
+        warn: (...args) => {
+          log.warn({ jobId, ...args });
+        },
```
blefebvre, line 48: While less frequently used, there are additional log levels. For completeness sake should we provide support for them as well?

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -0,0 +1,55 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * Wraps the provided function to enhance the logging functionality by including the
+ * `jobId` in each log statement.
+ *
+ * @param {function} fn - The original function that will be wrapped.
+ * @returns {function(object, object): Promise<Response>} A new function that, when invoked,
+ *  will automatically include the `jobId` in all log statements within the context.
```
blefebvre, line 19: Not necessarily, as they could still use `context.log` if desired. 

> When wrapped with this function, a new logger is available as context.contextualLog.

... or perhaps `context.jobLog` which I think describes the purpose better.

packages/spacecat-shared-utils/src/log-wrapper.js
blefebvre: On second read, and seeing the duplicated `return fn(message, context);` line, I think it might improve readability to change this guard to check for the positive (e.g. `if (isObject(message) && hasText(message.jobId)) {`) and then include an `else` which sets contextualLog to log, just above the single return statement.

packages/spacecat-shared-utils/test/log-wrapper.test.js
```diff
@@ -0,0 +1,127 @@
...
+
+// Helper function to check if a key-value pair exists in a JSON object
+function containsKeyValue(obj, key, value) {
+  return Object.prototype.hasOwnProperty.call(obj, key) && obj[key] === value;
+}
+
+const mockFnFromSqs = sinon.spy();
+let mockContext;
+
+describe('logWrapper', () => {
```
blefebvre, line 37:
```suggestion
describe('logWrapper tests', () => {
```

packages/spacecat-shared-utils/test/log-wrapper.test.js
```diff
@@ -0,0 +1,127 @@
...
+
+  it('should not wrap log methods when jobId is missing', async () => {
+    const wrappedFn = logWrapper(mockFnFromSqs);
+
+    await wrappedFn({ message: 'This contains no jobId' }, mockContext);
+
+    mockContext.contextualLog.info('Test log');
+    const logArgs = mockContext.log.info.getCall(0).args[0];
+    expect(containsKeyValue(logArgs, 'jobId', undefined)).to.be.false;
+    expect(logArgs).to.equal('Test log');
```
blefebvre, line 80: Could check the log.debug spy here as well to ensure it logs the issue.

packages/spacecat-shared-utils/test/log-wrapper.test.js
```diff
@@ -0,0 +1,127 @@
...
+    const wrappedFn = logWrapper(mockFnFromSqs);
+
+    await wrappedFn({}, mockContext);
+
+    mockContext.contextualLog.info('Info log');
+    mockContext.contextualLog.error('Error log');
+    mockContext.contextualLog.debug('Debug log');
+    mockContext.contextualLog.warn('Warn log');
+
+    expect(containsKeyValue(mockContext.log.info.getCall(0).args[0], 'jobId', undefined)).to.be.false;
```
blefebvre, line 122: Are these `containsKeyValue` checks false because the object doesn't have `jobId`, or because its value does not equal `undefined`? I think a more specific check might be needed here.

blefebvre (changes requested):
Open to discussion on naming, as I missed a few days when this was being worked on.

codevoyage:
> Open to discussion on naming, as I missed a few days when this was being worked on.

Hi @blefebvre,

@bhellema and I discussed the naming convention on Slack and decided to go with logWrapper. This choice allows for adding more properties to the log in the future, ensuring it is not too specific to jobs.

You can find our discussion here:: https://cq-dev.slack.com/archives/C06HYS3AQUF/p1725992056209769?thread_ts=1725989731.295479&cid=C06HYS3AQUF

packages/spacecat-shared-utils/src/index.js
codevoyage: Addressed in the reply to comment above ^^^^

packages/spacecat-shared-utils/src/log-wrapper.js
codevoyage: See this discussion: [Slack thread](https://cq-dev.slack.com/archives/C06HYS3AQUF/p1725990669391249?thread_ts=1725989731.295479&cid=C06HYS3AQUF).

I changed the name from jobLog to contextualLog, as we may want to add more properties in the future that provide context, not necessarily related to jobs.

packages/spacecat-shared-utils/src/log-wrapper.js
codevoyage: These are the log levels and methods supported by `console`. Which ones do we want to support? So far, I’ve only seen `info`, `debug`, and `error` being used.

```
console.log()
console.info()
console.warn()
console.error()
console.debug()
console.trace()
console.group()
console.groupEnd()
console.groupCollapsed()
console.table()
console.time()
console.timeEnd()
console.timeLog()
console.assert()
console.clear()
console.count()
console.countReset()
console.dir()
console.dirxml()
console.profile()
console.profileEnd()
console.timeStamp()
```

codevoyage: @blefebvre, which log levels should we support? So far, I’ve seen `info`, `debug`, and `error` being used, but it could be extended to include more levels like so:
```js
context.contextualLog = {
  info: (...args) => {
    log.info({ jobId, ...args });
  },
  error: (...args) => {
    log.error({ jobId, ...args });
  },
  debug: (...args) => {
    log.debug({ jobId, ...args });
  },
  warn: (...args) => {
    log.warn({ jobId, ...args });
  },
  trace: (...args) => {
    log.trace({ jobId, ...args });
  },
  log: (...args) => {
    log.log({ jobId, ...args });
  },
  group: (...args) => {
    log.group({ jobId, ...args });
  },
  groupEnd: () => {
    log.groupEnd();
  },
  groupCollapsed: (...args) => {
    log.groupCollapsed({ jobId, ...args });
  },
  assert: (condition, ...args) => {
    log.assert(condition, { jobId, ...args });
  },
  count: (...args) => {
    log.count({ jobId, ...args });
  },
  countReset: (...args) => {
    log.countReset({ jobId, ...args });
  },
  time: (...args) => {
    log.time({ jobId, ...args });
  },
  timeEnd: (...args) => {
    log.timeEnd({ jobId, ...args });
  },
  timeLog: (...args) => {
    log.timeLog({ jobId, ...args });
  },
};
```

packages/spacecat-shared-utils/src/log-wrapper.js
codevoyage: The wrapper is used in some files within the content-processor and content-scraper Lambdas that are unrelated to imports. We should likely exclude those invocations from triggering alerts.

packages/spacecat-shared-utils/src/log-wrapper.js
codevoyage: It doesn’t appear as a key, but rather as jobId: 'f91afda0-afc8-467e-bfa3-fdbeba3037e8' as part of the message, which can be included in the search query.”

packages/spacecat-shared-utils/src/log-wrapper.js
codevoyage: I’ve added the trace level to the contextual logger. I haven’t seen usage for the other functions so far, but we can update the logger if the need arises later. What do you think?

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -11,45 +11,46 @@
...
  */
 export function logWrapper(fn) {
   return async (message, context) => {
     const { log } = context;
 
     if (log) {
-      if (!message || !message.jobId) {
-        log.debug('Missing jobId, hence it will not be included in log messages.');
-        context.contextualLog = log;
+      if (message && 'jobId' in message) {
```
blefebvre, line 27:
```suggestion
      if (isObject(message) && 'jobId' in message) {
```

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -11,45 +11,46 @@
...
+          debug: (...args) => {
+            log.debug({ jobId, ...args });
+          },
+          warn: (...args) => {
+            log.warn({ jobId, ...args });
+          },
+          trace: (...args) => {
+            log.trace({ jobId, ...args });
+          },
+        };
```
blefebvre, line 47: I remembered where I had seen the `silly` level: https://github.com/adobe/helix-log/blob/main/src/log.js#L961

(not saying we need it in this impl; just didn't want you to think I was crazy!)

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -11,45 +11,46 @@
...
-        log.debug('Missing jobId, hence it will not be included in log messages.');
-        context.contextualLog = log;
+      if (message && 'jobId' in message) {
+        const { jobId } = message;
 
-        return fn(message, context);
+        // Enhance the log object to include jobId in all log statements
+        context.contextualLog = {
+          info: (...args) => {
+            log.info({ jobId, ...args });
```
blefebvre, line 33: I have concerns about how this `...rest` and then `...spread` approach changes the resulting log format. Take the following simple function for example:

```
function logTest(...args) { console.log({jobId: "12345", ...args}); }
```

If I invoke the existing JS logger with multiple parameters, they are concatenated together to form a single log line:

```
> console.log("Here is my Log", "with something else (maybe a stacktrace)", "and an object:", {foo: "property"})
Here is my Log with something else (maybe a stacktrace) and an object: { foo: 'property' }
```

If I send the same params to our new logger, simplified in the `logTest` function above, I would be surprised that it produces a very different output:

```
> logTest("Here is my Log", "with something else (maybe a stacktrace)", "and an object:", {foo: "property"})
{
  '0': 'Here is my Log',
  '1': 'with something else (maybe a stacktrace)',
  '2': 'and an object:',
  '3': { foo: 'property' },
  jobId: '12345'
}
```

I think this is partially due to how we are taking the `...rest` and then spreading its values onto an object literal, which adds the index-like property names.

At this point, given no clear way to add a `key` (or `marker` in Splunk terms) to each entry, what if we simplified things to just include a prefix on each log entry containing the jobId? Modifying my simple logger above:

```
function logTest2(...args) { console.log('[jobId=12345]', ...args); }
```

Sample invocation and output:

```
> logTest2("Here is my Log", "with something else (maybe a stacktrace)", "and an object:", {foo: "property"})
[jobId=12345] Here is my Log with something else (maybe a stacktrace) and an object: { foo: 'property' }
```

This is much closer to the existing logging format expected from a logger, and does not have any degradation in terms of query-ability of the resulting log entry. WDTY?

packages/spacecat-shared-utils/src/log-wrapper.js
codevoyage: ESLint reported a circular dependency issue with the isObject function, so I did not include it.

packages/spacecat-shared-utils/src/log-wrapper.js
codevoyage: Interesting. The repository appears to be the log forwarder, as indicated by the forwarding logic found here: [coralogix.js](https://github.com/adobe/helix-log/blob/main/src/coralogix.js). To align with this behaviour, I can add the fatal, verbose, and silly log levels.

packages/spacecat-shared-utils/src/log-wrapper.js
codevoyage: Sounds good to me. Just to clarify, if additional "markers" need to be added to the logger, would they appear like this: [jobId=12345, api-key=xyz, other-property=abc], or something like this: [jobId=12345] [api-key=xyz] [other-property=abc]? What are you envisioning?

codevoyage: @blefebvre I have updated the "marker" format.

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -11,40 +11,42 @@
...
+ * logging and returns the result of the original function.
+ *
+ * `context.contextualLog` will include logging methods with `jobId` prefixed, or fall back
+ * to the existing `log` object if no `jobId` is provided.
  */
 export function logWrapper(fn) {
   return async (message, context) => {
     const { log } = context;
 
     if (log) {
```
blefebvre, line 35: The typical pattern for these wrappers is to only add to context if the property in question is not already set. A small change: 
```suggestion
    if (log && !context.contextualLog) {
```

packages/spacecat-shared-utils/src/log-wrapper.js
```diff
@@ -11,40 +11,42 @@
...
-      if (message && 'jobId' in message) {
+      if (typeof message === 'object' && message !== null && 'jobId' in message) {
```
blefebvre, line 36: Even if its null you can just use the `in` operator, so I think this will suffice for readability:
```suggestion
      if (typeof message === 'object' && 'jobId' in message) {
```

blefebvre (approved):
👍 with a couple minor suggestions.

packages/spacecat-shared-utils/src/log-wrapper.js
blefebvre: > if additional "markers" need to be added to the logger

I think the format you described would work fine, to a point. Eventually it would be great to figure out how we can send custom "keys" to Coralogix, or if that is even available with the log API we are using.

PR #375, opened 2024-09-19
feat(data-cruncher): port queries to the data cruncher from RUM Explorer

ekremney:
This PR ports the existing queries to the `data-cruncher` from https://github.com/adobe/helix-website/

Main idea behind the port is to use the same definitions for the metrics and concepts.

So far, this PR contains the ports for the following queries:
1. CWV
2. Experiment
3. High Inorganic High Bounce Rate

more ports will follow...

---

This PR uses a copied version of the `cruncher.js` from https://github.com/adobe/helix-website/ library. The plan is to extract the `cruncher.js` into a shared library soon, so that the way raw RUM data crunched is the same in all places.

Thanks @karlpauls for the contributions! 💯


packages/spacecat-shared-rum-api-client/src/common/cruncher.js
ramboz, line 1: I'd be tempted to go one step further and extract each class to individual files

ramboz (approved):
Looks good to me

packages/spacecat-shared-rum-api-client/.eslintignore
```diff
@@ -0,0 +1 @@
+src/common/cruncher.js
```
solaris007, line 1: why are we excluding this?

packages/spacecat-shared-rum-api-client/.nycrc.json
```diff
@@ -10,5 +10,8 @@
   "all": true,
   "include": [
     "src/**/*.js"
+  ],
+  "exclude": [
```
solaris007, line 14: why are we excluding this?

packages/spacecat-shared-rum-api-client/src/common/cruncher.js
ekremney: done 👍

packages/spacecat-shared-rum-api-client/.eslintignore
ekremney: no more exclusion. we are inclusive now ❤️

packages/spacecat-shared-rum-api-client/.nycrc.json
ekremney: no more exclusion. we are inclusive now ❤️

PR #381, opened 2024-09-24
fix: SITES-25634 [Importer] Expose hasCustomImportJs and hasCustomHeaders as two new fields on a import job

bhellema:


packages/spacecat-shared-data-access/test/unit/service/import-job/index.test.js
```diff
@@ -51,195 +50,95 @@ describe('Import Job Tests', () => {
...
+
+          const result = await exportedFunctions.getImportJobsByDateRange(
+            mockDynamoClient,
+            TEST_DA_CONFIG,
+            mockLog,
+            '2024-05-27T14:26:00.000Z',
+            '2024-06-02T14:26:00.000Z',
+          );
+
+          // expect that 2 results were in the db, and that the results were not modified
```
blefebvre, line 96: > and that the results were not modified
Does this warrant another `expect`?

blefebvre (approved):
Looks like changes will have to follow on the api-service side to set these props correctly — right?

swetabar (approved):

bhellema:
> Looks like changes will have to follow on the api-service side to set these props correctly — right?

Correct: https://github.com/adobe/spacecat-api-service/pull/509

PR #384, opened 2024-09-24
fix: Add a unit test for imports.write_all_domains scope

swetabar:
## Related Issues

Add a unit test for imports.write_all_domains scope

### DIFF
--- a/packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
+++ b/packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
@@ -59,6 +59,11 @@ describe('ApiKey Model tests', () => {
       expect(apiKey.getScopes()).to.deep.equal([{ name: 'imports.write_all_domains', domains: [] }]);
     });
 
+    it('creates an ApiKey object for a user with scope - imports.write', () => {
+      const apiKey = createApiKey({ ...validApiKey, scopes: [{ name: 'imports.write', domains: ['https://adobe.com', 'https://test.com'] }] });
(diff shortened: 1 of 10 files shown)


packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
```diff
@@ -54,6 +54,11 @@ describe('ApiKey Model tests', () => {
       expect(apiKey.getCreatedAt()).is.not.empty;
     });
 
+    it('creates an ApiKey object for a user with scope - imports.write_all_domains', () => {
+      const apiKey = createApiKey({ ...validApiKey, scopes: [{ name: 'imports.write_all_domains', domains: [] }] });
+      expect(apiKey.getScopes()).to.deep.equal([{ name: 'imports.write_all_domains', domains: [] }]);
```
bhellema, line 59: Too bad that the scopeNames array isn't exported, such that we could use a constant here.  

Do we not want to verify that the domains provided in the createApiKey are actually still in the apiKey?

swetabar: I can add another test with the API scope`'imports.write'` to verify that the domain created is still in the apiKey.

bhellema: I think it's beneficial to have these tests if someone were to modify the createApiKey function that somehow lost the domains, at least we would catch these issues.

swetabar: Done.

bhellema (approved):

PR #385, opened 2024-09-26
fix: Modify the scope name to imports.all_domains

swetabar:
## Related Issues
https://jira.corp.adobe.com/browse/SITES-22756

Reference PR: https://github.com/adobe/spacecat-api-service/pull/511/files

### DIFF
--- a/packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
+++ b/packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
@@ -55,8 +55,8 @@ describe('ApiKey Model tests', () => {
     });
 
     it('creates an ApiKey object for a user with scope - imports.all_domains', () => {
-      const apiKey = createApiKey({ ...validApiKey, scopes: [{ name: 'imports.all_domains', domains: [] }] });
-      expect(apiKey.getScopes()).to.deep.equal([{ name: 'imports.all_domains', domains: [] }]);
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
```diff
@@ -54,9 +54,9 @@ describe('ApiKey Model tests', () => {
       expect(apiKey.getCreatedAt()).is.not.empty;
     });
 
-    it('creates an ApiKey object for a user with scope - imports.write_all_domains', () => {
-      const apiKey = createApiKey({ ...validApiKey, scopes: [{ name: 'imports.write_all_domains', domains: [] }] });
-      expect(apiKey.getScopes()).to.deep.equal([{ name: 'imports.write_all_domains', domains: [] }]);
+    it('creates an ApiKey object for a user with scope - imports.all_domains', () => {
+      const apiKey = createApiKey({ ...validApiKey, scopes: [{ name: 'imports.all_domains', domains: [] }] });
```
blefebvre, line 58: To be valid, this key should have the `imports.write` scope as well, right?

blefebvre (approved):

packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
swetabar: Yes. Should that check be here?
Or in spacecat-api-service?

swetabar: Added the additional scope to the test.

PR #392, opened 2024-10-02
feat: Expire URLs automatically after set period of time

blefebvre:
Adds an `expiresAt` field to import-url entities which will be leveraged by Dynamo to automatically remove items after their TTL has elapsed.

## Related Issues

- [SITES-23373](https://jira.corp.adobe.com/browse/SITES-23373)

### DIFF
--- a/packages/spacecat-shared-data-access/docs/schema.json
+++ b/packages/spacecat-shared-data-access/docs/schema.json
@@ -1089,6 +1089,10 @@
       },
       "NonKeyAttributes": [
         {
+          "AttributeName": "expiresAt",
+          "AttributeType": "S"
+        },
+        {
           "AttributeName": "file",
           "AttributeType": "S"
         },
--- a/packages/spacecat-shared-data-access/docs/schema.json
+++ b/packages/spacecat-shared-data-access/docs/schema.json
@@ -1090,7 +1090,7 @@
       "NonKeyAttributes": [
         {
           "AttributeName": "expiresAt",
-          "AttributeType": "S"
+          "AttributeType": "N"
         },
         {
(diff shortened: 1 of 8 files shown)


packages/spacecat-shared-data-access/docs/schema.json
```diff
@@ -1088,6 +1088,10 @@
         }
       },
       "NonKeyAttributes": [
+        {
+          "AttributeName": "expiresAt",
+          "AttributeType": "S"
```
bhellema, line 1093: Should this not be an N to be a number - as the value should the epoch time.

packages/spacecat-shared-data-access/src/models/importer/import-url.js
```diff
@@ -31,6 +33,8 @@ const ImportUrl = (data) => {
   self.getPath = () => self.state.path;
   // Resulting path and filename of the imported .docx file
   self.getFile = () => self.state.file;
+  // Expiration date for the URL
+  self.getExpiresAt = () => self.state.expiresAt;
```
bhellema, line 37: I don't know if we should be exposing a getter for this property.  It doesn't buy us anything in our business logic, it's simply a property that should be stamped into the db when a row is created and no one should really read it.

bhellema (approved):

packages/spacecat-shared-data-access/src/models/importer/import-url.js
blefebvre: As discussed offline, will keep the accessor to facilitate serialization and avoid having to "reach in" to the underlying state.

PR #411, opened 2024-10-24
feat: introduce urlNumber and totalUrlCount to the import-url entity

blefebvre:
Relates to: Race condition in the content-processor due to GSI's eventual consistency model

## Related Issues

- [SITES-26245](https://jira.corp.adobe.com/browse/SITES-26245)


blefebvre:
I had hoped we could reuse these model objects, but I see now that this won't be necessary. Thanks @bhellema — closing.

PR #417, opened 2024-10-31
feat: support FIFO queues with the SQS sendMessage helper

blefebvre:
Add support for passing a `MessageGroupId` to the `sqsClient.send(..)` method to enable the use of FIFO queues.

Without this property we see the following error in the logs:

```
...failed with error: MissingParameter: The request must contain the parameter MessageGroupId.
```

## Related Issues

- [SITES-26422](https://jira.corp.adobe.com/browse/SITES-26422) Migrate all import queues to FIFO SQS queues

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22599,7 +22599,7 @@
     },
     "packages/spacecat-shared-data-access": {
       "name": "@adobe/spacecat-shared-data-access",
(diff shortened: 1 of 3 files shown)


swetabar (approved):
LGTM!

PR #422, opened 2024-11-05
SITES-26591 [Import Assistant] Expand GPT client to use chat endpoint

atopper:
## Related Issues
[SITES-26591](https://jira.corp.adobe.com/browse/SITES-26591) [Import Assistant] Expand GPT client to use chat endpoint

- Required for https://jira.corp.adobe.com/browse/SITES-26526

## Changes
- `fetchCapabilityExecution` : `fetch` endpoint is now deprecated, replaced with `fetchCapabilityExecution` with exactly the same parameters, errors and response.  It still hits the Firefall _/v2/capability_execution/job_ end point.
- `fetchChatCompletion` : is introduced.  It uses the Firefall _/v2/chat/completions_ endpoint.
  - It takes a `prompt` parameter (_like fetchCapabilityExecution()_)
  - It does not poll (_it just waits for the Firefall response_)
  - It takes a second `options` object parameter which, for now, accepts:
    - `model` property: indicates which llmModel to use.  It defaults to `gpt-4-turbo`.
      - `"gpt-4-vision"` is required to handle imageUrls
    - `imageUrls` array property: each image is added to the call using the `"type": "image_url"` content type (see [docs](https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=adobeds&title=tutorial+-+using+firefall+generative+APIs)).
    - `responseFormat` property: indicates the format of the AI call response.  `json_object` is supported by "gpt-35-turbo-1106" and "gpt-4-turbo"


packages/spacecat-shared-gpt-client/README.md
```diff
@@ -58,7 +60,8 @@ async function fetchInsights(prompt) {
       log: console,
     });
 
-    const insights = await client.fetch(prompt);
+    // Internally, use Firefall's .../capability_execution/job endpoint.
```
atopper, line 63: Almost talked myself out of this comment since it is an implementation detail, but thought it might help some dev's with Firefall experience.

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -161,23 +163,113 @@ export default class FirefallClient {
     return jobStatusResponse;
   }
 
-  async fetch(prompt) {
+  /**
+   * Fetches data from Firefall Chat Completion API.
+   * @param prompt The text prompt to provide to Firefall
+   * @param options The options for the call:
+   *                - imageUrls: An array of URLs of the images to provide to Firefall
+   * @returns {Promise<*>}
```
atopper, line 171: Actually returns the response.  I'll update this to:
```
@returns {string} - Firefall response
```

packages/spacecat-shared-gpt-client/README.md
```diff
@@ -68,6 +71,38 @@ async function fetchInsights(prompt) {
...
+#### Via Chat Completions endpoint
+
+```javascript
+async function fetchInsights(prompt) {
+  try {
+    const client = FirefallClient.createFrom({
+      env: {
+        FIREFALL_API_ENDPOINT: 'https://api.firefall.example.com',
+        FIREFALL_API_KEY: 'yourApiKey',
+        FIREFALL_API_CAPABILITY_NAME: 'yourCapabilityName',
```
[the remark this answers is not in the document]
arumsey: Can this example use a different function name than `fetchInsights`?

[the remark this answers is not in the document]
arumsey: How would one specify the model they want to use with `fetchChatCompletion`? Is that part of the env or can each fetch request specify a different model?

packages/spacecat-shared-gpt-client/README.md
```diff
@@ -68,6 +71,38 @@ async function fetchInsights(prompt) {
...
+        IMS_CLIENT_CODE: 'yourClientCode',
+        IMS_CLIENT_SECRET: 'yourClientSecret',
+      },
+      log: console,
+    });
+    const options = {
+      imageUrls: ['data:image/png;base64,iVBORw0KGgoAAAA...=']
+    };
+
+    // // Internally, use Firefall's .../chat/completions endpoint.
```
arumsey, line 95: Like above I think this can be added as a short description above the code block. Also, Firefall is an implementation detail that users of the gpt client don't really need to know about. There are a bunch of `FIREFALL_*` env options however so I guess users of this client do need to know that Firefall is being used.

packages/spacecat-shared-gpt-client/README.md
```diff
@@ -68,6 +71,38 @@ async function fetchInsights(prompt) {
...
+    const insights = await client.fetchChatCompletion(prompt, { options });
+    console.log('Insights:', insights);
```
arumsey, line 97: Change to be something other than insights in ordeer to differentiate from capabilities endpoint?

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -161,23 +163,113 @@ export default class FirefallClient {
...
+        content: [
+          {
+            type: 'text',
+            text: prompt,
+          },
+        ],
+      };
+      const body = {
+        llm_metadata: {
+          model_name: this.config.FIREFALL_API_CAPABILITY_NAME,
```
arumsey, line 192: I think we should add a different config property for model name. `FIREFALL_API_MODEL_NAME`. Will the model name be validated against models we want to support? How can I request a response format of `json_object`?

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -161,23 +163,113 @@ export default class FirefallClient {
...
+          llm_type: 'azure_chat_openai',
+        },
+        messages: [
+          {
+            role: 'system',
+            content: 'You are a helpful assistant designed to output JSON.',
```
arumsey, line 198: This system role is only need when a response format of `json_object` is requested.

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -161,23 +163,113 @@ export default class FirefallClient {
...
+          },
+          userRole,
+        ],
+      };
+      if (hasImageUrls) {
+        if (!Array.isArray(options?.imageUrls)) {
+          throw new Error('imageUrls must be an array.');
+        }
+        // Warn if model might not handle images.
+        if (!this.config.capabilityName.includes('vision') && this.log?.warn) {
```
arumsey, line 208: If images are provided can we just override the model name to always be gpt-4-vision? How much knowledge of Open AI models do we want users of the client to understand vs just letting the client figure out the best model to use?

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -89,15 +96,10 @@ export default class FirefallClient {
...
+  async #submitPrompt(prompt, body, path) {
     const apiAuth = await this.#getApiAuth();
 
-    const body = JSON.stringify({
-      input: prompt,
-      capability_name: this.config.capabilityName,
-    });
-
-    const url = createUrl(`${this.config.apiEndpoint}/v2/capability_execution/job`);
+    const url = createUrl(`${this.config.apiEndpoint}${path}`);
```
arumsey, line 102: Should the firefall-client export an array of valid Firefall api endpoints that can be used with the config? Otherwise every user of the client will need to maintain their own set of endpoints.

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -89,15 +96,10 @@ export default class FirefallClient {
     this.log.debug(`${message}: took ${duration}ms`);
   }
 
-  async #submitJob(prompt) {
+  async #submitPrompt(prompt, body, path) {
```
arumsey, line 99: Where do prompt and body get combined? Since the body will be different depending on the Firefall path used I would think that prompt would have to be combined into the body before calling `#submitPrompt`. If that is the case `body` could remain an object and only be stringified right before calling `httpFetch`.

packages/spacecat-shared-gpt-client/README.md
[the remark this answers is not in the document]
atopper: Example from tests:  `FIREFALL_API_CAPABILITY_NAME: 'gpt4_32k_completions_capability',`
I thought Capability mapped to Model internally, but can't find what I read that gave me that impression.  Perhaps that also answers your first question above.
Shall I accept a `llm_metadata` or `model_name` (_hardcoding llm_type?_) in the options?
Default:
`"model_name": "gpt-4-turbo"`
Thoughts?

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
atopper: I assumed it would always return a JSON object.  I believe that is the only requirement right now.  If others need another response, should we anticipate and allow it right now?  Or allow a future enhancement for it?

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
atopper: Nice.  I wasn't sure about this check.  Without the right model, the images are ignored.  But I didn't want to lock us into specific models, as they'll change (maybe one will come out using `images` instead of `vision`).  Allowing any model and allowing Firefall to reject them would be more future proof.
Or default to `"gpt-4-turbo"` and switch to `"gpt-4-vision"` if there are images?  Any allowances to use other models to use fewer credits or more advanced responses?
What do you think?

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
arumsey: This isn't the HTTP response format, it is the message format. We are already using both formats with the import assistant.

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
arumsey: If you want to stay open ended and less opinionated then you probably should do any validation then since you won't know what model name to validate against. If image urls are provided just add them to the messsages payload and pss it along to Firefall.

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
atopper: Completely correct.  Nice.
`Prompt` is already in the body at this point, put together as required by each endpoint.

packages/spacecat-shared-gpt-client/README.md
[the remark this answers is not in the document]
arumsey: Somehow the client will need to build up a base Firefall payload object that messsages can be added to.

```
{
  llm_metadata: {
    model_name: 'gpt-4-turbo',
    llm_type: 'azure_chat_openai',
  },
  response_format: {
    type: 'json_object',
  },
  messages: [
    {
      role: 'system',
      content: 'You are a helpful assistant designed to output JSON.',
    },
  ],
}
```

`model_name` should be supplied to the client
`response_format` could also be supplied and can either be undefined or `json_object`.

Only when a json response object is supplied then the system message will need to be provided.

You can read more about JSON mode here: https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=adobeDS&title=Tutorial+-+Using+Firefall+Generative+APIs#TutorialUsingFirefallGenerativeAPIs-ChatCompletionsV2:JSONModesupport

Also, only 2 models currently support json mode. ` "gpt-35-turbo-1106", "gpt-4-turbo"`

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
arumsey: IIUC, every user of the firefall client must provide their own Firefall endpoint URL.

eg. `FIREFALL_API_ENDPOINT: 'https://api.firefall.example.com',`

AFAIK, there are only 2 possible URLs so why can't the firefall client manage them?

```
Stage: [firefall-stage.adobe.io](http://firefall-stage.adobe.io/)
Production: [firefall.adobe.io](http://firefall.adobe.io/)
```

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
atopper: I'm adding `model` and `responseFormat` to options.
```
      const options = {
        model: 'gpt-4-turbo',
        responseFormat: 'json_object',
      };
```

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
arumsey: If the endpoint changes any users of the client would have to be be updated anyway. Any string could still be used as the FIREFALL_API_ENDPOINT but exporting the default URL values would just be helpful.

I was thinking of exporting a frozen firefall config object.

```
export const firefallEndpointConfig = Object.freeze({
  stage: 'http://firefall-stage.adobe.io',
  prod: 'http://firefall.adobe.io',
});
```

Where? I guess it could be a static object of the `FirefallClient` class.

solaris007: the URL should be controlled via env var from the AWS secrets relevant to the dev/stage/prod environment.

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -161,23 +167,125 @@ export default class FirefallClient {
...
+   * Fetches data from Firefall Chat Completion API.
+   * @param prompt The text prompt to provide to Firefall
+   * @param options The options for the call:
+   *                - imageUrls: An array of URLs of the images to provide to Firefall
+   * @returns {Object} - AI response
+   */
+  async fetchChatCompletion(prompt, options = {}) {
+    const getBody = () => {
+      const { imageUrls, responseFormat, model: llmModel = 'gpt-4-turbo' } = options || {};
+      const hasImageUrls = imageUrls && imageUrls.length > 0;
```
solaris007, line 180: as `imageUrls` is accessed from the parent `options` the destructuring and validation should occur outside of the `getBody` function for efficiency reasons.

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -161,23 +167,125 @@ export default class FirefallClient {
...
+            type: 'text',
+            text: prompt,
+          },
+        ],
+      };
+
+      if (hasImageUrls) {
+        if (!Array.isArray(imageUrls)) {
+          throw new Error('imageUrls must be an array.');
+        }
```
solaris007, line 195: see previous comment

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -161,23 +167,125 @@ export default class FirefallClient {
...
+        options.imageUrls.forEach((imageUrl) => {
+          userRole.content.push({
+            type: 'image_url',
```
solaris007, line 198: should use constants or config object for all such static references

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
atopper: I'm reading that as letting the AWS secrets control the endpoint, and not having any exported config's specifying the possible endpoints.
So I'll remove them.  Let me know if I misunderstood.

arumsey (approved):

PR #428, opened 2024-11-07
feat: support queries to generate self-service api-keys

swetabar:
## Related Issues

[SITES-25915](https://jira.corp.adobe.com/browse/SITES-25915)

Supporting the following queries with this PR:
* `getApiKeysByImsUserIdAndImsOrgId`
* `getApiKeyById`
* `updateApiKey`

Sample apiKey entry in the DB:
```
{
  id: '321456987efgh',
  hashedApiKey: 'axspbgh5649075',
  name: 'test-name',
  imsUserId: 'test-ims-user-id',
  imsOrgId: 'test-ims-org-id',
  createdAt: '2024-10-29T14:26:00.000Z'
  expiresAt: '2025-05-29T14:26:00.000Z',
  scopes: [{
        name: 'imports.write',
        domains: ['https://www.test.com'],
      }],
}
```


packages/spacecat-shared-data-access/src/index.d.ts
```diff
@@ -27,6 +27,12 @@ export declare const ImportUrlStatus: {
   readonly FAILED: 'FAILED';
 };
 
+// packages/spacecat-shared-data-access/src/models/api-key/api-key-constants.js
+export declare const ApiKeyStatus: {
+  readonly ACTIVE: 'ACTIVE';
+  readonly INACTIVE: 'INACTIVE';
+};
+
 // TODO: introduce AuditType interface or Scores interface
```
atopper, line 36: Intentional TODO?

packages/spacecat-shared-data-access/src/index.d.ts
```diff
@@ -839,6 +865,16 @@ export interface DataAccess {
   createNewApiKey: (
       apiKeyData: object,
   ) => Promise<ApiKey>;
+  updateApiKey: (
+      apiKey: ApiKey,
+  ) => Promise<ApiKey>;
+  getApiKeyByImsUserIdAndImsOrgId: (
+      imsUserId: string,
+      imsOrgId: string,
+  ) => Promise<ApiKey | null>;
```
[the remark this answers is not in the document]
swetabar: Good point. Typo for the interface. I'll fix it.

packages/spacecat-shared-data-access/src/models/api-key/api-key-constants.js
```diff
@@ -0,0 +1,20 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * Api Key Status Types
+ * Any changes to this object needs to be reflected in the index.d.ts file as well.
+ */
+export const ApiKeyStatus = {
```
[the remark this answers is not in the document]
swetabar: The key is still considered `ACTIVE` in that case. It is updated to `INACTIVE` when keys are deleted.

packages/spacecat-shared-data-access/src/service/api-key/accessPatterns.js
```diff
@@ -42,10 +43,68 @@ export const getApiKeyByHashedApiKey = async (hashedApiKey, dynamoClient, log, c
  * @param {DynamoClient} dynamoClient
  * @param {Object} config
  * @param {Logger} log
- * @returns {Promise<ApiKeyDto>}
+ * @returns {Promise<ApiKey>}
  */
 export const createNewApiKey = async (apiKey, dynamoClient, config, log) => {
-  const item = ApiKeyDto.toDynamoItem(apiKey);
-  await dynamoClient.putItem(config.tableNameApiKeys, item, log);
+  const item = createApiKey(apiKey);
```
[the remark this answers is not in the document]
swetabar: I think we were not using this method before. I refactored it to work similar to how import-job works now for consistency.

packages/spacecat-shared-data-access/src/models/api-key/api-key-constants.js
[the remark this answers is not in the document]
blefebvre: Just thinking out loud: would we be any worse off if we didn't have the `status` column? We'll have to check the `deletedAt`, `revokedAt`, and `expiresAt` anyways to determine if a key is still valid, so `status` seems potentially unnecessary.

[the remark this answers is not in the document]
swetabar: Yes. I could remove status for the api-key.

packages/spacecat-shared-data-access/docs/schema.json
```diff
@@ -1209,11 +1209,49 @@
           "AttributeName": "revokedAt",
           "AttributeType": "S"
         },
+        {
+          "AttributeName": "deletedAt",
+          "AttributeType": "S"
+        },
+        {
+          "AttributeName": "status",
```
blefebvre, line 1217: Status is still present here.

packages/spacecat-shared-data-access/src/dto/api-key.js
```diff
@@ -10,15 +10,16 @@
...
 export const ApiKeyDto = {
 
   /**
      * Converts an ApiKey object into a DynamoDB item.
      * @param apiKey
-     * @returns {{createdAt: *, name: *, imsUserId, scopes, revokedAt, key: *,
-     *          imsOrgId: *, expiresAt: *}}
+     * @returns {{createdAt: string, name: string, imsUserId: string,
+     * scopes: array<object>, revokedAt: string, deletedAt: string,
+     * hashedApiKey: *, imsOrgId: *, expiresAt: *}}
```
blefebvre, line 22: Please add types.

packages/spacecat-shared-data-access/src/models/api-key/api-key.js
```diff
@@ -40,8 +40,37 @@ const ApiKey = (data) => {
   self.getCreatedAt = () => self.state.createdAt;
   self.getExpiresAt = () => self.state.expiresAt;
   self.getRevokedAt = () => self.state.revokedAt;
+  self.getDeletedAt = () => self.state.deletedAt;
   self.getScopes = () => self.state.scopes;
 
+  /**
```
blefebvre, line 46: Would this model be the right place for an `isValid()` method, which could check all the internal properties to ensure that a key is still valid?

blefebvre: Or `isActive()`, alternatively?

swetabar: We could do that in the service. Do we need it in spacecat-shared?

blefebvre: If we do so in the service than anywhere that re-uses this model will need to implement something similar. I see it as a direct concern of the model itself, "am I still valid", and think that it makes sense to have that logic live alongside the model. 

I also like the pattern of "asking" the entity itself, are you still valid: `if (apiKeyInstance.isValid()) {` — nice and clean.

packages/spacecat-shared-data-access/src/models/api-key/api-key.js
```diff
@@ -43,6 +43,28 @@ const ApiKey = (data) => {
...
+   * Checks if the apiKey is valid.
+   * @returns {boolean} True if the apiKey is valid, false otherwise
+   */
+  self.isValid = () => {
+    const now = new Date();
+
+    if (self.state.deletedAt && new Date(self.state.deletedAt) > now) {
```
blefebvre, line 53: Is this comparison correct? If deletedAt is > now, that would mean it was deleted in the future — right? And therefore still valid?

packages/spacecat-shared-data-access/src/models/api-key/api-key.js
```diff
@@ -43,6 +43,28 @@ const ApiKey = (data) => {
...
+      return false;
+    }
+
+    if (self.state.revokedAt && new Date(self.state.revokedAt) > now) {
+      return false;
+    }
+
+    if (self.state.expiresAt && new Date(self.state.expiresAt) > now) {
+      return false;
+    }
```
blefebvre, line 63: Same issue with these two, I think.

If an expiry is set when the key is created (as it must) than it will always be in the future, which would mean that this check fails and the key is not considered valid.

packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
```diff
@@ -149,5 +149,42 @@ describe('ApiKey Model tests', () => {
...
+
+    it('returns true if deletedAt, revokedAt, and expiresAt are not set', () => {
+      apiKey.state.deletedAt = null;
+      apiKey.state.revokedAt = null;
+      apiKey.state.expiresAt = null;
+      expect(apiKey.isValid()).to.be.true;
+    });
+
+    it('returns false if deletedAt is after the current time', () => {
+      apiKey.updateDeletedAt(new Date(Date.now() + 10000).toISOString());
```
blefebvre, line 169: Can concrete values please be used for these test cases so they are easier for a human to reason about?

ie. if a key expires in 2025 and it is 2024, it should still be valid.

packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
```diff
@@ -149,5 +149,42 @@ describe('ApiKey Model tests', () => {
...
+      apiKey.updateRevokedAt(new Date(Date.now() + 10000).toISOString());
+      expect(apiKey.isValid()).to.be.false;
+    });
+
+    it('returns false if expiresAt is after the current time', () => {
+      apiKey.updateExpiresAt(new Date(Date.now() + 10000).toISOString());
+      expect(apiKey.isValid()).to.be.false;
+    });
+
+    it('returns true if deletedAt, revokedAt, and expiresAt are before the current time', () => {
```
blefebvre, line 183: This is the core of the issue, I think. 

- if expiresAt is in the future: still valid
- if deletedAt or revokedAt are in the past: no longer valid

packages/spacecat-shared-data-access/src/models/api-key/api-key.js
swetabar: Yes. Correct.

swetabar: It will be deleted in future. So, the key is still valid.

packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
blefebvre: In other words: expiresAt is the only property we should _expect_ to be in the future. If ANY of these properties are set to a date before `now`, the key should not be considered valid.

packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
swetabar: Done.

packages/spacecat-shared-data-access/test/unit/models/api-key.test.js
```diff
@@ -166,24 +166,24 @@ describe('ApiKey Model tests', () => {
...
       expect(apiKey.isValid()).to.be.false;
     });
 
     it('returns true if deletedAt, revokedAt, and expiresAt are before the current time', () => {
-      apiKey.updateDeletedAt(new Date(Date.now() - 10000).toISOString());
-      apiKey.updateRevokedAt(new Date(Date.now() - 10000).toISOString());
-      apiKey.updateExpiresAt(new Date(Date.now() - 10000).toISOString());
+      apiKey.updateDeletedAt('2025-05-27T14:26:00.000Z');
+      apiKey.updateRevokedAt('2025-05-27T14:26:00.000Z');
+      apiKey.updateExpiresAt('2025-05-27T14:26:00.000Z');
```
blefebvre, line 186: This test will start failing next May :) 

Can we bump the year to 3025 please?

blefebvre (approved):
Minor item about dates in test cases.

PR #437, opened 2024-11-12
feat: enable internal workspace reporting

alinarublea:
To be used for #project-success-studio-for-acom (adobe.com domains, channel is private but could become public in the future)


packages/spacecat-shared-data-access/test/unit/models/site/config.test.js
```diff
@@ -183,7 +185,6 @@ describe('Config Tests', () => {
       const slackConfig = dynamoItem.slack;
       const slackMentions = dynamoItem.handlers[404].mentions.slack;
       expect(slackConfig.channel).to.equal('channel1');
-      expect(slackConfig.workspace).to.equal('workspace1');
```
iuliag, line 186: `expect(isInternal).to.equal(false);`?

iuliag (approved):
Could you also please link in the description the feature/ticket/thread where this is used?

solaris007 (approved):

PR #442, opened 2024-11-15
fix: resolving url when site has bot protection

alinarublea:
business.adobe.com has bot protection and we cannot resolve the audit url to run the broken-backlinks audit

### DIFF
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -88,7 +88,7 @@ async function composeAuditURL(url) {
   const resp = await fetch(urlWithScheme, {
     method: 'GET',
     headers: {
-      'User-Agent': 'curl/7.88.1', // Set the same User-Agent
+      'User-Agent': 'curl/7.88.1',
     },
   });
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -85,7 +85,12 @@ function composeBaseURL(domain) {
 
 async function composeAuditURL(url) {
   const urlWithScheme = prependSchema(url);
-  const resp = await fetch(urlWithScheme);
+  const resp = await fetch(urlWithScheme, {
+    method: 'GET',
+    headers: {
+      'User-Agent': 'curl/7.88.1',
```
ekremney, line 91: adding curl to the user agent feels like we are announcing that we are a bot 😅 if this helps to circumvent any bot detection system, let's collect their names so that we can tell our customers not to use those so called "bot detection systems"

alinarublea: I understand, but for now how can we run the audits ?

dzehnder (approved):

PR #443, opened 2024-11-15
fix: SITES-26948 [Import Assistant] Allow calls to Firefall to use specified ims org id

atopper:
The context's IMS_CLIENT_ID value is used in 2 places:
- IMSClient: IMS Token fetch (as an IMS Client Id)
- GPTClient: Firefall call as the `x-gw-ims-org-id` header value

This PR is to allow the `x-gw-ims-org-id` to use a specified, optional value.

## Related Issues
https://jira.corp.adobe.com/browse/SITES-26948

### DIFF
--- a/packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
+++ b/packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
@@ -50,7 +50,7 @@ export default class FirefallClient {
 
     const {
       FIREFALL_API_ENDPOINT: apiEndpoint,
-      IMS_CLIENT_ID: imsOrg,
+      IMS_CLIENT_ID: imsClientId,
       FIREFALL_IMS_ORG_ID: firefallImsOrgId,
       FIREFALL_API_KEY: apiKey,
       FIREFALL_API_POLL_INTERVAL: pollInterval = 2000,
@@ -70,7 +70,7 @@ export default class FirefallClient {
       apiKey,
       capabilityName,
       imsClient,
-      imsOrg: firefallImsOrgId || imsOrg,
+      imsOrg: firefallImsOrgId || imsClientId,
       pollInterval,
     }, log);
   }
@@ -188,7 +188,6 @@ export default class FirefallClient {
    *          - imageUrls: An array of URLs of the images to provide to Firefall
    *          - model: LLM Model to use (default: gpt-4-turbo).  Use 'gpt-4-vision' with images.
    *          - responseFormat: The response format to request from Firefall (accepts: json_object)
-   *          - imsOrgId: The IMS Org id to send with Firefall prompts (optional)
    * @returns {Object} - AI response
    */
   async fetchChatCompletion(prompt, options = {}) {
--- a/packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
+++ b/packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
@@ -107,7 +107,14 @@ export default class FirefallClient {
     this.log.debug(`${message}: took ${duration}ms`);
   }
 
-  async #submitPrompt(body, path) {
+  /**
+   * Submit a prompt to the Firefall API.
+   * @param body The body of the request.
+   * @param path The Firefall API path.
+   * @param imsOrgId An optional IMS Org ID, to override the config default.
+   * @returns {Promise<unknown>}
+   */
+  async #submitPrompt(body, path, imsOrgId = undefined) {
     const apiAuth = await this.#getApiAuth();
 
     const url = createUrl(`${this.config.apiEndpoint}${path}`);
@@ -115,7 +122,7 @@ export default class FirefallClient {
       'Content-Type': 'application/json',
       Authorization: `Bearer ${apiAuth}`,
       'x-api-key': this.config.apiKey,
-      'x-gw-ims-org-id': this.config.imsOrg,
+      'x-gw-ims-org-id': imsOrgId ?? this.config.imsOrg,
     };
 
     this.log.info(`URL: ${url}, Headers: ${JSON.stringify(headers)}`);
@@ -181,10 +188,16 @@ export default class FirefallClient {
    *          - imageUrls: An array of URLs of the images to provide to Firefall
    *          - model: LLM Model to use (default: gpt-4-turbo).  Use 'gpt-4-vision' with images.
    *          - responseFormat: The response format to request from Firefall (accepts: json_object)
+   *          - imsOrgId: The IMS Org id to send with Firefall prompts (optional)
    * @returns {Object} - AI response
    */
   async fetchChatCompletion(prompt, options = {}) {
-    const { imageUrls, responseFormat, model: llmModel = 'gpt-4-turbo' } = options || {};
+    const {
+      imageUrls,
+      responseFormat,
+      imsOrgId,
+      model: llmModel = 'gpt-4-turbo',
+    } = options || {};
     const hasImageUrls = imageUrls && imageUrls.length > 0;
 
     const getBody = () => {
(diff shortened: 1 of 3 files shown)


packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -107,15 +107,22 @@ export default class FirefallClient {
...
 
-  async #submitPrompt(body, path) {
+  /**
+   * Submit a prompt to the Firefall API.
+   * @param body The body of the request.
+   * @param path The Firefall API path.
+   * @param imsOrgId An optional IMS Org ID, to override the config default.
+   * @returns {Promise<unknown>}
+   */
+  async #submitPrompt(body, path, imsOrgId = undefined) {
```
arumsey, line 117: Instead of of an extra parameter maybe the third parameter should be an object of overrides?

arumsey: Where does `this.config.imsOrg` gets its IMS Org value from?

arumsey: Could this logic be moved to `FirefallClient.createFrom`?

```
    const {
      FIREFALL_API_ENDPOINT: apiEndpoint,
      IMS_CLIENT_ID: imsClientId,
      IMS_ORG: imsOrg,
      FIREFALL_API_KEY: apiKey,
      FIREFALL_API_POLL_INTERVAL: pollInterval = 2000,
      FIREFALL_API_CAPABILITY_NAME: capabilityName = 'gpt4_32k_completions_capability',
    } = context.env;

   const firefallImsOrg = imsOrg || imsClientId;

    return new FirefallClient({
      apiEndpoint,
      apiKey,
      capabilityName,
      imsClient,
      imsOrg: firefallImsOrg,
      pollInterval,
    }, log);
```

atopper: > Where does this.config.imsOrg gets its IMS Org value from?
From `IMS_CLIENT_ID: imsOrg,` line 53.

atopper: > Could this logic be moved to FirefallClient.createFrom?
`createFrom` and `context` seems to dealing with ENV variables (from AWS secrets) and Spacecat injected values.
Since our IMS_ORG used in the FF call isn't part of that, it doesn't seem to fit in `createFrom()`.

If a GPTClient called `fetchChatCompletion()` twice, it might want to use 2 different IMS ORG ids, but has to use the same context.

atopper: Are you picturing this:
```
async #submitPrompt(body, path, overrides = {}) {
 ...
  const headers = {
      'Content-Type': 'application/json',
      Authorization: `Bearer ${apiAuth}`,
      'x-api-key': this.config.apiKey,
      'x-gw-ims-org-id': imsOrgId ?? this.config.imsOrg,
      ...overrides,
    }
```
I am thinking the use here of "_this.config.imsOrg (i.e. IMS_CLIENT_ID)_" as an IMS ORG ID is actually incorrect (or at least misleading) and it would be the only exception ever made.  I could make it overrides if you want.

arumsey: > > Could this logic be moved to FirefallClient.createFrom?
> > `createFrom` and `context` seems to dealing with ENV variables (from AWS secrets) and Spacecat injected values.
> > Since our IMS_ORG used in the FF call isn't part of that, it doesn't seem to fit in `createFrom()`.
> 
> If a GPTClient called `fetchChatCompletion()` twice, it might want to use 2 different IMS ORG ids, but has to use the same context.

Is this an actual issue or a hypothetical issue? If the ims org is coming from the api key which is part of the context why would you want to change the ims org during the same request?

arumsey: > Are you picturing this:
> 
> ```
> async #submitPrompt(body, path, overrides = {}) {
>  ...
>   const headers = {
>       'Content-Type': 'application/json',
>       Authorization: `Bearer ${apiAuth}`,
>       'x-api-key': this.config.apiKey,
>       'x-gw-ims-org-id': imsOrgId ?? this.config.imsOrg,
>       ...overrides,
>     }
> ```
> 
> I am thinking the use here of "_this.config.imsOrg (i.e. IMS_CLIENT_ID)_" as an IMS ORG ID is actually incorrect (or at least misleading) and it would be the only exception ever made. I could make it overrides if you want.

I am now preferring the possibility of needing any overrides at this level. I think it just makes the API more confusing having different places where header values can be set.

atopper: Hypothetical I suppose.
Since this is shared in all of spacecat, I'd prefer not to assume how some other service would use it.  I guess they could create a new GPTClient if need be.
Why do you think it should be embedded in the client, instead of just during the call where it is used?  The latter feels more flexible but if you are sure, I can change it.

arumsey: We should get some more historical context form @solaris007. At first glance it seems confusing to allow only img org to be set two different ways. When creating a `FirefallClient` I would expect all the `config` values it needs would come from the context. It would be nice if all the `FirefallClient` methods worked the same way. `fetchChatCompletion` shouldn't need to set it img org differently than `fetchCapabilityExecution` does. So the question that needs to be anwered is why does `fetchCapabilityExecution` use the ims client id as its ims org? Is this because there was no ims org to use? Is this a bug?

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -51,6 +51,7 @@ export default class FirefallClient {
     const {
       FIREFALL_API_ENDPOINT: apiEndpoint,
       IMS_CLIENT_ID: imsOrg,
```
arumsey, line 53: Can this be changed to `imsClientId` to be less confusing?

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -181,10 +188,15 @@ export default class FirefallClient {
    *          - imageUrls: An array of URLs of the images to provide to Firefall
    *          - model: LLM Model to use (default: gpt-4-turbo).  Use 'gpt-4-vision' with images.
    *          - responseFormat: The response format to request from Firefall (accepts: json_object)
+   *          - imsOrgId: The IMS Org id to send with Firefall prompts (optional)
```
arumsey, line 191: Not needed anymore.

packages/spacecat-shared-gpt-client/test/clients/firefall-client.test.js
```diff
@@ -246,6 +246,7 @@ describe('FirefallClient', () => {
         imageUrls: [base64ImageUrl],
         model: 'gpt-4-turbo',
         responseFormat: 'json_object',
+        imsOrgId: 'ImsOrgId',
```
arumsey, line 249: Not needed?

arumsey (approved):

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
atopper: Um.  Ya.  That would be more accurate.
Would I change line 73 too?
Feels like I'm starting to step on existing code but...

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -69,7 +70,7 @@ export default class FirefallClient {
       apiKey,
       capabilityName,
       imsClient,
-      imsOrg,
+      imsOrg: firefallImsOrg || imsOrg,
```
arumsey, line 73:
```suggestion
      imsOrg: firefallImsOrg || imsClientID,
```

arumsey (approved):

PR #446, opened 2024-11-17
fix: exclude bot traffic from page views and opportunities ctr calculations

rpapani:
https://jira.corp.adobe.com/browse/SITES-25839


packages/spacecat-shared-rum-api-client/test/common/rum-bundler-client.test.js
```diff
@@ -102,6 +162,7 @@ describe('Rum bundler client', () => {
     const containsCorrectData = result.every((item) => item.events.length === 1 && item.events[0].checkpoint === 'good');
 
     expect(result.length).to.equal(dates.length);
+    // eslint-disable-next-line no-unused-expressions
```
rpapani, line 165: somehow `eslint` is blocking me from committing, so I've to add these, even though this is not related to my change.

rpapani:
Thank you @ekremney that's good to have rum-distiller already taking care of this. I still want to have this change in the rum-api-client so that anyone using this for example rage click, will have this temporarily, when we move to rum-distiller we get this by default. Would it be ok to merge this PR then?

ekremney (approved):

PR #541, opened 2025-01-15
fix: improve sqs event adapter error handling

solaris007:
This is to prevent ops alerts when no messages are in the payload.


solaris007:
merging to prevent ops alerts

PR #566, opened 2025-01-30
feat: add utility api for forms metrics

nit23uec:

## Related Issues

Thanks for contributing!


nit23uec:
> @nit23uec i am not sure `spacecat-shared-utils` is the right place for forms-specific utils, as the tools herein are rather generic and use-case independent. not that i have a better idea currently :)

Thanks for the review @solaris007 . These methods are expected to be used atleast in audit worker and reporting worker. So, it would be useful to park them at a common location to avoid duplication. 
Should I create a new package (spacecat-shared-forms) for this use case?

solaris007 (approved):

PR #577, opened 2025-02-05
fix(sites-28856): update shared-data schema to include type in options

bhellema:
To include support for crosswalk imports, we need to define `type` property in our options field.  The value for type can either be 'xwalk' or 'doc'.  The schema needs to be updated to support this new property.

image

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/import-job/import-job.schema.js
+++ b/packages/spacecat-shared-data-access/src/models/import-job/import-job.schema.js
@@ -15,6 +15,7 @@
 import {
   isInteger,
   isIsoDate,
+  isNonEmptyObject,
   isNumber,
   isObject,
   isValidUrl,
@@ -41,7 +42,7 @@ const ImportOptionTypeValidator = {
     }
   },
   [ImportJob.ImportOptions.DATA]: (value) => {
-    if (!isObject(value)) {
+    if (value && !isNonEmptyObject(value)) {
       throw new Error(`Invalid value for ${ImportJob.ImportOptions.DATA}: ${value}`);
     }
   },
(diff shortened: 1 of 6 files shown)


packages/spacecat-shared-data-access/src/models/import-job/import-job.schema.js
```diff
@@ -35,6 +36,11 @@ const ImportOptionTypeValidator = {
       throw new Error(`Invalid value for ${ImportJob.ImportOptions.PAGE_LOAD_TIMEOUT}: ${value}`);
     }
   },
+  [ImportJob.ImportOptions.TYPE]: (value) => {
+    if (!isString(value) || !Object.values(ImportJob.ImportOptionTypes).includes(value)) {
```
solaris007, line 40: the `isString` check is not required here i think

blefebvre (approved):
👍

packages/spacecat-shared-data-access/src/models/import-job/import-job.schema.js
```diff
@@ -35,6 +35,16 @@ const ImportOptionTypeValidator = {
       throw new Error(`Invalid value for ${ImportJob.ImportOptions.PAGE_LOAD_TIMEOUT}: ${value}`);
     }
   },
+  [ImportJob.ImportOptions.TYPE]: (value) => {
+    if (!Object.values(ImportJob.ImportOptionTypes).includes(value)) {
+      throw new Error(`Invalid value for ${ImportJob.ImportOptions.TYPE}: ${value}`);
+    }
+  },
+  [ImportJob.ImportOptions.DATA]: (value) => {
+    if (!isObject(value)) {
```
solaris007, line 44: is the value nullable? if so i'd check for `if (value && !isNonEmptyObject(value))` - would not necessarily store an empty object for nothing.

PR #585, opened 2025-02-06
feat: introduce fetch config for site to use custom headers

dzehnder:
Introduce new fetchConfig to be able to add custom headers, like User-Agent, per site. 

We need this as the fetch operation does not work for all sites with the same User-Agent.


ekremney (approved):
👍

alinarublea (approved):

solaris007 (approved):

PR #590, opened 2025-02-10
feat(rum-api-client): admin key support

ekremney:

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-rum-api-client/README.md
+++ b/packages/spacecat-shared-rum-api-client/README.md
@@ -12,18 +12,22 @@ npm install @adobe/spacecat-shared-rum-api-client
 
 ## Usage
 
-#### Creating and instance from Helix UniversalContext
+#### Creating an instance from Helix UniversalContext
 
 ```js
-const context = {}; // Your AWS Lambda context object
+// The context must include an 'env' property so that the client can use RUM_ADMIN_KEY if needed.
+const context = { env: process.env };
 const rumApiClient = RUMAPIClient.createFrom(context);
-
 ```
 
-#### From constructor
+#### Using the constructor
 
 ```js
-const rumApiClient = new RUMAPIClient();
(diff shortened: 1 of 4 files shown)


rpapani (approved):

solaris007 (commented):
IIUC if not present, the rum domain key will be retrieved for every query. after retrieval it is not cached in the instance, which would retrieve it multiple times if the same instance is used for multiple queries for the same domain.

furthermore, it would be super cool if the retrieval of the domain key could be public API. e.g. the auto-onboarding feature and others could benefit from retrieving the key during onboarding and store it for later re-use.

rpapani:
> the rum domain key will be retrieved for every query.

Good point, its better to store it in the secrets like we did manually. 

> furthermore, it would be super cool if the retrieval of the domain key could be public API. e.g. the auto-onboarding feature and others could benefit from retrieving the key during onboarding and store it for later re-use.

All the audits will be using RUM API Client, so if the domain key is missing, this will be handled by RUM API client. Is there any requirement in auto-onboarding to access the domain key, with out using the RUM API client? If not, I think its better to keep under the RUM API Client.

solaris007:
> All the audits will be using RUM API Client, so if the domain key is missing, this will be handled by RUM API client. Is there any requirement in auto-onboarding to access the domain key, with out using the RUM API client? If not, I think its better to keep under the RUM API Client.

@rpapani i'd like to see the ability to retrieve the RUM domain key for the following reasons:
- the RUM key may be needed by other systems (e.g. agents) that do not use the RUM domain key
- we may want to expose the key in UIs (e.g. BO)
- RUM client should not know how to store things as customer secrets (separation of concerns)

rpapani:
> RUM client should not know how to store things as customer secrets (separation of concerns)

hmm. I see. Can we do this as part of auto-onboard command?

solaris007 (approved):
Good stuff :)

PR #579, opened 2025-02-05
Splunk client

langswei:
Splunk client to give the option for CDN logs as a data source.

Please ensure your pull request adheres to the following guidelines:
- [x] make sure to link the related issues in this description
- [x] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes

## Related Issues
#578

PR #591, opened 2025-02-10
feat: util for reading .query static files, modeled after getPrompt

langswei:
In #579 comments, @solaris007 suggested handling queries as static files, external to the javascript source.  The getPrompt helper function was mentioned as a model.  This PR will create an updated release of the spacecat-shared-utils package which can be referenced by #579.

### DIFF
--- a/packages/spacecat-shared-utils/package.json
+++ b/packages/spacecat-shared-utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@adobe/spacecat-shared-utils",
-  "version": "1.30.2",
+  "version": "1.30.3",
   "description": "Shared modules of the Spacecat Services - utils",
   "type": "module",
   "engines": {
--- a/packages/spacecat-shared-utils/package.json
+++ b/packages/spacecat-shared-utils/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@adobe/spacecat-shared-utils",
-  "version": "1.30.3",
+  "version": "1.30.2",
   "description": "Shared modules of the Spacecat Services - utils",
   "type": "module",
   "engines": {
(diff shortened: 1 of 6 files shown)


langswei:
Was I supposed to update the version in package.json myself, or is that automated?

nitinja (approved):

packages/spacecat-shared-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@adobe/spacecat-shared-utils",
-  "version": "1.30.2",
+  "version": "1.30.3",
```
solaris007, line 3: the version should only be increased by semantic release during CI/CD

packages/spacecat-shared-utils/src/helpers.js
```diff
@@ -131,3 +131,23 @@ export async function getPrompt(placeholders, filename, log = console) {
...
+ *
+ * @param {Object} placeholders - A JSON object containing values to replace in the query content.
+ * @param {String} filename - The filename of the query file.
+ * @param {Object} log - The logger
+ * @returns {Promise<string|null>} - A promise that resolves to a string with the query content,
+ * or null if an error occurs.
+ */
+export async function getQuery(placeholders, filename, log = console) {
+  try {
+    const queryContent = await fs.readFile(`./static/queries/${filename}.query`, { encoding: 'utf8' });
```
solaris007, line 147: these two lines are a duplication of `getPrompt` with the exception of path segments. You could create an internal function used by both `getQuery` and `getPrompt` with the outer functions only adjusting the relative file path (e.g. `queries/${fileName}.query`.

packages/spacecat-shared-utils/package.json
langswei: reverted

packages/spacecat-shared-utils/src/helpers.js
langswei: @solaris007 Done, please review https://github.com/adobe/spacecat-shared/pull/591/commits/54860d8a7ed27ace3d1b3f7d62add0a3d6acc06d and approve if looks good.

solaris007 (approved):
LGTM :) Thank you!

PR #608, opened 2025-02-17
feat: Implement dependency check for handlers

herzog31:
This PR allows to express dependencies between handlers. For example, the audit for structured data requires the top pages of a site to be scraped by the `scrape-top-pages` handler. This handler itself relies on the `import-top-pages` handler.

* Implement handler dependencies as previously defined in https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-data-access/src/models/configuration/configuration.schema.js#L35-L39.
* When trying to enable a handler with un-met dependencies, an error is thrown which lists the handlers that need to be activated first.
* Consumer of the API already handle errors properly:
    * https://github.com/adobe/spacecat-api-service/blob/1c7f73a9fabc5f7ceac54d503b96646c8a26e09f/src/controllers/sites-audits-toggle.js
    * https://github.com/adobe/spacecat-api-service/blob/1c7f73a9fabc5f7ceac54d503b96646c8a26e09f/src/support/slack/commands/onboard.js
    * https://github.com/adobe/spacecat-api-service/blob/1c7f73a9fabc5f7ceac54d503b96646c8a26e09f/src/support/slack/commands/toggle-site-audit.js
* Updated unit tests.

## Related Issues
https://jira.corp.adobe.com/browse/SITES-28718


packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js
```diff
@@ -146,12 +146,58 @@ class Configuration extends BaseModel {
     const siteId = site.getId();
     if (this.isHandlerEnabledForSite(type, site)) return;
 
+    const deps = this.isHandlerDependencyMetForSite(type, site);
+    if (deps !== true) {
+      throw new Error(`Cannot disable handler ${type} for site ${siteId} because of missing dependencies: ${deps}`);
```
alinarublea, line 151: cannot enable ?

alinarublea (approved):
Only small logs comment the rest looks good to me

packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js
herzog31: Good catch. This is fixed now :)

packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js
```diff
@@ -146,12 +146,58 @@ class Configuration extends BaseModel {
...
+   * Check if all dependencies for a handler of given type are met for the given org.
+   *
+   * @param {string} type handler type
+   * @param {object} org org object
+   * @returns true if all dependencies are met, array with missing dependencies otherwise
+   */
+  isHandlerDependencyMetForOrg(type, org) {
+    const handler = this.getHandler(type);
+
+    if (!handler || !handler.dependencies) return true;
```
solaris007, line 167: could use `!isNonEmptyArray(handler?.dependencies)`

packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js
```diff
@@ -146,12 +146,58 @@ class Configuration extends BaseModel {
...
+
+    const unmetDependencies = handler.dependencies
+      .filter(({ handler: depHandler }) => !this.isHandlerEnabledForOrg(depHandler, org))
+      .map(({ handler: depHandler }) => depHandler);
+
+    return unmetDependencies.length === 0 ? true : unmetDependencies;
```
solaris007, line 173: could use `isNonEmptyArray`

solaris007 (approved):
LGTM, just cosmetics

iuliag:
`import-top-pages` is not a handler, it's currently an import configured at site-level (https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-data-access/src/models/site/config.js#L21), hence this implementation won't cover for the cases where the prerequisite is an import.

herzog31:
@iuliag that's a good point. How should we handle this case? Shall we extend the dependency schema for it, or handle it via the new audit steps? @alinarublea @solaris007 wdyt?

solaris007:
@herzog31 i would go with stepped audit for now

PR #619, opened 2025-02-21
feat: add spacecat user-agent for fetch requests

AndreiAlexandruParaschiv:
- add default user agent for tracking-fetch requests used in audits

### DIFF
--- a/packages/spacecat-shared-utils/src/tracing-fetch.js
+++ b/packages/spacecat-shared-utils/src/tracing-fetch.js
@@ -119,8 +119,8 @@ export async function tracingFetch(url, options) {
     return adobeFetch(url, {
       ...options,
       headers: {
-        ...options?.headers,
         'User-Agent': DEFAULT_USER_AGENT,
+        ...options?.headers,
       },
     });
   }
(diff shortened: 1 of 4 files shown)


packages/spacecat-shared-utils/src/tracing-fetch.js
```diff
@@ -115,10 +116,22 @@ export async function tracingFetch(url, options) {
   const parentSegment = AWSXRay.getSegment();
 
   if (!parentSegment) {
-    return adobeFetch(url, options);
+    return adobeFetch(url, {
+      ...options,
+      headers: {
+        ...options?.headers,
+        'User-Agent': DEFAULT_USER_AGENT,
```
alinarublea, line 123: I would reverse the order here, and in case there is a user-agent set the fetch should use that one instead of the default

AndreiAlexandruParaschiv: good point 👍

AndreiAlexandruParaschiv:
closing in favor or  :https://github.com/adobe/spacecat-shared/pull/620

PR #626, opened 2025-02-25
feat: import config schema

solaris007:
Add schema to site/org config detailing the supported import types and their configuration. Adds utility methods to enable/disable imports.

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/config.js
+++ b/packages/spacecat-shared-data-access/src/models/site/config.js
@@ -18,13 +18,13 @@ export const IMPORT_TYPES = {
   TOP_PAGES: 'top-pages',
 };
 
-const IMPORT_DESTINATIONS = {
+export const IMPORT_DESTINATIONS = {
   DEFAULT: 'default',
 };
 
-const IMPORT_SOURCES = {
+export const IMPORT_SOURCES = {
   AHREFS: 'ahrefs',
-  GSC: 'gsc',
+  GSC: 'google',
 };
 
 const IMPORT_BASE_KEYS = {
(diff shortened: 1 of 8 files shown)


alinarublea (approved):

packages/spacecat-shared-data-access/src/models/site/config.js
```diff
@@ -136,6 +198,47 @@ export const Config = (data = {}) => {
...
+    state.imports = state.imports.map(
+      (imp) => (imp.type === type ? { ...imp, enabled: false } : imp),
+    );
+
+    validateConfiguration(state);
+  };
+
+  self.getImportConfig = (type) => state.imports?.find((imp) => imp.type === type);
+
+  self.isImportEnabled = (type) => {
```
dzehnder, line 237: adding the `enabled` property and this function requires some changes in the jobs dispatcher as well. As we did not have that property before, the jobs dispatcher purely checks for the presence of the config: https://github.com/adobe/spacecat-jobs-dispatcher/blob/main/src/handlers/imports.js#L26-L31

packages/spacecat-shared-data-access/src/models/site/config.js
```diff
@@ -12,13 +12,75 @@
...
+  TOP_PAGES: 'top-pages',
+};
+
+const IMPORT_DESTINATIONS = {
+  DEFAULT: 'default',
+};
+
+const IMPORT_SOURCES = {
+  AHREFS: 'ahrefs',
+  GSC: 'gsc',
```
iuliag, line 27: The existing imports have the source `google` in the config for GSC.

PR #630, opened 2025-02-26
fix: import-schema backwards compatibility

solaris007:
The import schema update introduced via #626 accidentally set the newly introduce `enabled` property as required. this caused already existing import configurations in the database to be invalid, causing outages. The `enabled` flag is now optional. Relevant unit and IT tests have been enhanced.


rpapani (approved):
Looks good to me

PR #632, opened 2025-02-26
fix: improves logs for FF client failures

igor-grubic:
We need a way to see why Firefall is refusing some requests. Current error message is swallowed and not delivered to consumer.
This passes message across when available.

## Related Issues
https://jira.corp.adobe.com/browse/ASSETS-47371

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
+++ b/packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
@@ -137,7 +137,7 @@ export default class FirefallClient {
       let message = '';
       try {
         const errorBody = await response.json();
-        message = `Job submission failed with status code ${response.status} and message: ${errorBody.message}`;
+        message = `Job submission failed with status code ${response.status} and body: ${JSON.stringify(errorBody)}`;
       } catch {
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
```diff
@@ -134,7 +134,14 @@ export default class FirefallClient {
     });
 
     if (!response.ok) {
-      throw new Error(`Job submission failed with status code ${response.status}`);
+      let message = '';
+      try {
+        const errorBody = await response.json();
+        message = `Job submission failed with status code ${response.status} and message: ${errorBody.message}`;
+      } catch {
+        message = `Job submission failed with status code ${response.status}`;
```
solaris007, line 142: should in this case add the raw body

solaris007 (approved):

packages/spacecat-shared-gpt-client/src/clients/firefall-client.js
igor-grubic: 👍  removed `json()` in favour of `text()`

PR #633, opened 2025-02-26
fix(override-base-url): add overrideBaseURL in site.config.fetchConfig

iuliag:
Implements: https://jira.corp.adobe.com/browse/SITES-29423

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/config.js
+++ b/packages/spacecat-shared-data-access/src/models/site/config.js
@@ -49,6 +49,7 @@ export const IMPORT_TYPE_SCHEMAS = {
     type: Joi.string().valid(IMPORT_TYPES.TOP_PAGES).required(),
     ...IMPORT_BASE_KEYS,
     geo: Joi.string(),
+    limit: Joi.number().integer().min(1).max(2000),
   }),
 };
 
@@ -85,6 +86,7 @@ export const configSchema = Joi.object({
   ),
   fetchConfig: Joi.object({
     headers: Joi.object().pattern(Joi.string(), Joi.string()),
+    overrideBaseURL: Joi.string().uri(),
   }).optional(),
   handlers: Joi.object().pattern(Joi.string(), Joi.object({
     mentions: Joi.object().pattern(Joi.string(), Joi.array().items(Joi.string())),
--- a/packages/spacecat-shared-data-access/test/unit/models/site/config.test.js
+++ b/packages/spacecat-shared-data-access/test/unit/models/site/config.test.js
@@ -139,6 +139,7 @@ describe('Config Tests', () => {
         headers: {
           'User-Agent': 'custom-agent',
         },
+        overrideBaseURL: 'https://example.com',
       };
       config.updateFetchConfig(fetchConfig);
       expect(config.getFetchConfig()).to.deep.equal(fetchConfig);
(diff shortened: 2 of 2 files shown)


packages/spacecat-shared-data-access/src/models/site/config.js
```diff
@@ -85,6 +86,7 @@ export const configSchema = Joi.object({
   ),
   fetchConfig: Joi.object({
     headers: Joi.object().pattern(Joi.string(), Joi.string()),
+    overrideBaseURL: Joi.string().uri(),
```
rpapani, line 89: will this also work for import config or should we've a separate config for imports?

iuliag: I'm going to open a PR in the import-worker to use the new override provided at site.config.fetchConfig level

rpapani: thank you. So, there'll be separate PR for broken-backlinks audit to leverage the overrideUrl right? the `url` introduced in https://jira.corp.adobe.com/browse/SITES-29345 isn't required any more right?

iuliag: Yes, I'm updating all the repos (see PR list in https://jira.corp.adobe.com/browse/SITES-29423?focusedId=46432145&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-46432145), while I figure out why nock doesn't work as expected in the unit tests here.

solaris007 (approved):

ekremney:
`overrideBaseURL` in [SITES-29423](https://jira.corp.adobe.com/browse/SITES-29423) is a great idea. However implementing a `resolveFinalURL` method which does all the redirect following logic intertwined with `overrideBaseURL` might create confusion.

I see two shortcomings of `resolveFinalURL` method:
1. http calls are error prone (ie bot detection) won't work all the time
2. `resolveFinalURL` logic is different for all consumers (imports audits etc)

hence, I suggest to use `overrideBaseURL` field with `getBaseURL` method (with a `useOverride` param which is false by default) and let the url resolving to consumers. Examples:

no `overrideBaseURL` set:
```
baseURL: https://spacecat.com
overrideBaseURL: null

getBaseURL() => https://spacecat.com
getBaseURL(useOverride=true) => https://spacecat.com
```

`overrideBaseURL` set:
```
baseURL: https://spacecat.com
overrideBaseURL: https://spacedog.com

getBaseURL() => https://spacecat.com
getBaseURL(useOverride=true) => https://spacedog.com
```

whatever the consumer downstream (import, audit etc) can decide which one to use @iuliag

solaris007:
@ekremney not sure i can support the proposal. while I understand where you're coming from, the `getBaseURL` has almost ID character and i see cases where developers have `overrideBaseURL=>true` set but later treat the baseURL as before. Also, if possible, auto-generated accessors should not be overridden. It may also lead to the temptation to use `setBaseURL` for "override" URLs.

ekremney:
> Also, if possible, auto-generated accessors should not be overridden.

@solaris007 got it, similar fashion a `getBaseURLOverride()` would do the job as well. Then the consumer would have to check if there is an manual override set.

rpapani:
> Then the consumer would have to check if there is an manual override set.

Rather than each consumer handling base or override, it'll be great, if the `resolveUrl` takes care of this.

PR #635, opened 2025-02-27
feat: Broken internal links validate and update

AnaniMarius:

## Related Issues

Thanks for contributing!


packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -349,4 +373,40 @@ export default class ContentClient {
...
+
+    this.log.info(`Updating page links for ${this.site.getId()} and path ${path}`);
+
+    const docPath = this.#resolveDocPath(path);
+    const document = await this.rawClient.getDocument(docPath);
+
+    const updateLinkPromises = brokenLinks.map(async (brokenLink) => {
+      let response;
+      // the if else might not be necessary except for logging
+      if (this.contentSource.type === CONTENT_SOURCE_TYPE_DRIVE_GOOGLE) {
```
alinarublea, line 393: I think we dont need the if here

packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -137,6 +136,31 @@ const validateRedirects = (redirects) => {
   }
 };
 
+export const validateURLs = (urls) => {
```
alinarublea, line 139: I think we can re-use this in the case of validateRedirects as well, and just pass the different regexp as parameter

packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -349,4 +354,31 @@ export default class ContentClient {
 
     this.#logDuration('updateRedirects', startTime);
   }
+
+  async updateBrokenInternalLinks(path, brokenLinks) {
```
alinarublea, line 358: I was thinking again here and I think it would be better if we receive in the beginning just one brokenLink

PR #636, opened 2025-02-28
configurable threshold update.

anikethrevale:

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/configuration/configuration.schema.js
+++ b/packages/spacecat-shared-data-access/src/models/configuration/configuration.schema.js
@@ -32,6 +32,8 @@ const handlerSchema = Joi.object().pattern(Joi.string(), Joi.object(
       orgs: Joi.array().items(Joi.string()),
     }),
     enabledByDefault: Joi.boolean().required(),
+    movingAvgThreshold: Joi.number().min(0).optional(),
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-data-access/src/models/configuration/configuration.schema.js
```diff
@@ -32,6 +32,8 @@ const handlerSchema = Joi.object().pattern(Joi.string(), Joi.object(
       orgs: Joi.array().items(Joi.string()),
     }),
     enabledByDefault: Joi.boolean().required(),
+    movingAvgThreshold: Joi.number().min(0).optional(),
+    percentageChangeThreshold: Joi.number().min(0).optional(),
```
alinarublea, line 36: I think the minimal here should be at least 1

iuliag (commented):
Could you also update the TS type? see https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-data-access/src/models/site/index.d.ts#L63 
And update the relevant unit tests? see https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-data-access/test/unit/models/site/config.test.js#L555

alinarublea (approved):

PR #657, opened 2025-03-06
feat: tracing fetch timeout

solaris007:
Introduce capability to set timeout for requests using `tracingFetch`. Default timeout is set to 10s.

### DIFF
--- a/packages/spacecat-shared-utils/src/tracing-fetch.js
+++ b/packages/spacecat-shared-utils/src/tracing-fetch.js
@@ -9,7 +9,7 @@
  * OF ANY KIND, either express or implied. See the License for the specific language
  * governing permissions and limitations under the License.
  */
-import { Request } from '@adobe/fetch';
+import { Request, timeoutSignal, AbortError } from '@adobe/fetch';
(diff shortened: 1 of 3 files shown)


packages/spacecat-shared-utils/src/tracing-fetch.js
```diff
@@ -145,21 +190,22 @@ export async function tracingFetch(url, options) {
...
-    }
+  subSegment.addAnnotation('timeout_ms', timeout);
 
-    setSubSegmentFlagsByStatusCode(subSegment, response.status);
+  try {
+    // Create a new request with the signal
+    const requestWithSignal = new Request(request, { signal });
+
+    // Use the same fetchWithTimeout function but catch errors to handle subsegment
+    const response = await fetchWithTimeout(requestWithSignal, null, signal);
```
ekremney, line 200: how about passing an empty object instead of `null`? Spread operation on null might fail

`fetchWithTimeout(requestWithSignal, {}, signal)`

ekremney (approved):

PR #664, opened 2025-03-14
feat(rum-api-client): human readable error messages for the failed bundler api calls

ekremney:
This PR introduces human readable error messages for the failed RUM Bundler API calls.

For example:

```
Error: Query 'cwv' failed. Opts: {"domain":"www.domain.com","interval":7,"granularity":"hourly"}. Reason: Error fetching data from https://bundles.aem.page/bundles/www.domain.com/2025/03/13/14?domainkey=redacted: Request to https://bundles.aem.page/bundles/www.domain.com/2025/03/13/14?domainkey=redacted failed with status 500
```

### DIFF
--- a/packages/spacecat-shared-rum-api-client/src/common/rum-bundler-client.js
+++ b/packages/spacecat-shared-rum-api-client/src/common/rum-bundler-client.js
@@ -22,6 +22,19 @@ const ONE_DAY = ONE_HOUR * HOURS_IN_DAY;
 
 const CHUNK_SIZE = 31;
 
+function sanitizeURL(url) {
+  try {
+    const parsedUrl = new URL(url);
+    if (parsedUrl.searchParams.has('domainkey')) {
+      parsedUrl.searchParams.set('domainkey', 'redacted');
+    }
+    return parsedUrl.toString();
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-rum-api-client/src/common/rum-bundler-client.js
```diff
@@ -192,11 +206,26 @@ async function fetchBundles(opts, log) {
...
+      } catch (error) {
+        throw new Error(`Error fetching data from ${sanitizeURL(url)}: ${error.message}`);
+      }
+    }));
+
+    const bundles = await Promise.all(responses.map(async (response) => {
+      try {
+        return await response.json();
+      } catch (error) {
+        throw new Error(`Error parsing JSON from ${sanitizeURL(response.url)}: ${error.message}`);
```
rpapani, line 226: @ekremney does this bubble up the error to the caller? Can we record all the failed urls and return the list of failed urls along with fetched bundles? Or can we introduce `ignoreErrors` option and ignore the errors and return the fetched bundles only? I don't want to fail the entire RUM bundles request if we can't get some bundles.. in almost all of the scenarios we're ok to calculate the insights we need with available data.

lenaorobei:
@ekremney any chance you can look into Ravi's suggestion in the PR?

PR #669, opened 2025-03-20
feat: jwt auth handler (WIP)

solaris007:
add new auth handler that evaluates a JWT signed by spacecat.


packages/spacecat-shared-http-utils/src/auth/handlers/ims.js
```diff
@@ -83,9 +77,9 @@ export default class AdobeImsHandler extends AbstractHandler {
   }
 
   async #validateToken(token, config) {
-    const decoded = await decodeJwt(token);
-    if (config.name !== decoded.as) {
-      throw new Error(`Token not issued by expected idp: ${config.name} != ${decoded.as}`);
+    const claims = await decodeJwt(token);
```
alinarublea, line 80: I feel confused, decodeJwt could be also renamed

alinarublea (approved):

packages/spacecat-shared-http-utils/src/auth/handlers/ims.js
solaris007: the method is imported from `jose`

PR #672, opened 2025-03-30
fix: adding forms constants at common place

05rahulsingh:

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-utils/src/formcalc.js
+++ b/packages/spacecat-shared-utils/src/formcalc.js
@@ -11,7 +11,9 @@
  */
 
 const DAILY_PAGEVIEW_THRESHOLD = 200;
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-utils/src/formcalc.js
```diff
@@ -11,7 +11,9 @@
  */
 
 const DAILY_PAGEVIEW_THRESHOLD = 200;
-const CR_THRESHOLD_RATIO = 0.2;
+const CR_THRESHOLD_RATIO = 0.3;
+//audit interval
+const INTERVAL = 15;
```
nit23uec, line 16: Let's update the default value of the methods to consume interval from here.
And then we can remove the arg from worker and reporter

nit23uec (approved):

PR #682, opened 2025-04-08
fix: enhance form calculations for view and submit

05rahulsingh:

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-utils/src/formcalc.js
+++ b/packages/spacecat-shared-utils/src/formcalc.js
@@ -66,7 +66,7 @@ function hasHighPageViews(pageViews) {
 }
 
 function hasLowerConversionRate(formSubmit, formViews) {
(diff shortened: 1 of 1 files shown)


packages/spacecat-shared-utils/src/formcalc.js
```diff
@@ -66,7 +66,7 @@ function hasHighPageViews(pageViews) {
 }
 
 function hasLowerConversionRate(formSubmit, formViews) {
-  return formSubmit / formViews < CR_THRESHOLD_RATIO;
+  return formSubmit > 0 && formSubmit / formViews < CR_THRESHOLD_RATIO;
```
nit23uec, line 69: we should not skip the form submit = 0 case.

05rahulsingh: done
https://github.com/adobe/spacecat-shared/pull/684

PR #710, opened 2025-04-28
feat(xray): non-aws ie local friendly xray instrumentation

ekremney:
currently, we are instrumenting the aws clients with xray regardless of the runtime examples below. This 
- [dynamo client](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-data-access/src/service/index.js#L25)
- [sqs client](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-utils/src/sqs.js#L33)

idea is to add a new xray wrapper to the main function which is runtime aware and skip xray instrumentation if the context.runtime (comes from helix aws-adapter) is not lambda:

currently:
```
const dbClient = client || AWSXray.captureAWSv3Client(new DynamoDB());
```

proposed:
```
const dbClient = client || context.xray.instrument(new DynamoDB());
```

`fetch` instrumentation is a bit tricky though. So far, `tracingFetch` was being imported directly in the code and use like this: https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-splunk-client/src/index.js#L13

to circumvent this, `tracingFetch` will check `process.env.AWS_EXECUTION_ENV` to see if invoked from an AWS Lambda function to fallback to adobe fetch if not invoked inside of an aws lambda function

example usage:

```
export const main = wrap(run)
  // other wrappers
  .with(xrayWrapper) // <- this one
  .with(helixStatus);
  ```


davidaurelio (approved):
lgtm

PR #739, opened 2025-05-15
fix: local development friendly aws clients and sqs wrapper

ekremney:
this PR changes:
1. the way SQS wrapper extracts the events when invoked outside of an AWS context
2. the way aws * client instrumentation with xray when initialized outside of an AWS context


packages/spacecat-shared-google-client/src/index.js
```diff
@@ -10,12 +10,11 @@
  * governing permissions and limitations under the License.
  */
 
-import AWSXray from 'aws-xray-sdk';
 import { google } from 'googleapis';
 import { OAuth2Client } from 'google-auth-library';
 import { GetSecretValueCommand, SecretsManagerClient } from '@aws-sdk/client-secrets-manager';
 import {
-  composeAuditURL,
+  composeAuditURL, instrumentAWSClient,
```
solaris007, line 17: carriage return

PR #758, opened 2025-05-23
feat: add updatedBy attribute to entities and ims api to resolve profile

ravverma:
- [ ] 
[SITES-31569 [updatedBy] Keep last updated by record for entities like opportunity and suggestion](https://jira.corp.adobe.com/browse/SITES-31569)

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/index.d.ts
+++ b/packages/spacecat-shared-data-access/src/models/site/index.d.ts
@@ -165,8 +165,6 @@ export interface Site extends BaseModel {
   setIsLiveToggledAt(isLiveToggledAt: string): Site;
   setOrganizationId(organizationId: string): Site;
   toggleLive(): Site;
-  getUpdatedBy(): string;
-  setUpdatedBy(updatedBy: string): Site;
 }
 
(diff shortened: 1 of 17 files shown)


packages/spacecat-shared-data-access/src/models/site/index.d.ts
```diff
@@ -165,6 +165,8 @@ export interface Site extends BaseModel {
   setIsLiveToggledAt(isLiveToggledAt: string): Site;
   setOrganizationId(organizationId: string): Site;
   toggleLive(): Site;
+  getUpdatedBy(): string;
```
solaris007, line 168: since you added the `updatedBy` property to the base model, the type info should be added on the `BaseModel` (similar to `updatedAt`) and not on the individual extending types.

solaris007 (approved):
LGTM, minor comment

packages/spacecat-shared-data-access/src/models/site/index.d.ts
ravverma: yes make sense. I have moved to `BaseModel`

PR #771, opened 2025-05-27
fix: config test

davidaurelio:

## Related Issues

Thanks for contributing!


iuliag:
Shouldn't we have both? `organic-keywords` and `organic-keywords-non-branded`?

iuliag:
In addition to the previous PR that introduced the new config schema, it should have been added to the `DEFAULT_IMPORT_CONFIGS`, so that it doesn't need to be explicitly enabled for newly added sites.

PR #776, opened 2025-05-30
feat: adding probable cta information in a form page

vdua:

## Related Issues

Thanks for contributing!


packages/spacecat-shared-rum-api-client/src/functions/form-vitals.js
```diff
@@ -135,6 +137,42 @@ function findFormCTAForInternalNavigation(bundles, formVitals) {
...
+        .filter((e) => e.source && !e.source.match(/\bform\b/i));
+
+      return clicks.map((_) => _.source);
+    }).filter((_) => _.length > 0).flat();
+    if (sources.length > 0) {
+      return {
+        ...item,
+        cta: {
+          sources,
+          form: item.url,
```
vdua: form not required since this is same as the top url,
also need to make sources unique

PR #796, opened 2025-06-10
fix: for adobe users admin access to only admin group members

ravverma:
This PR is to provide all org's data access to only those members who are member of a dedicated IMS admin group only.

## Related Issues

Thanks for contributing!


packages/spacecat-shared-http-utils/src/auth/handlers/ims.js
```diff
@@ -135,8 +135,9 @@ export default class AdobeImsHandler extends AbstractHandler {
       const config = loadConfig(context);
       const payload = await this.#validateToken(token, config);
       const imsProfile = await context.imsClient.getImsUserProfile(token);
+      const isAdmin = await context.imsClient.isUserASOAdmin(token);
```
solaris007, line 138: this will cause an additional request to IMS for every request incoming - also, the IMS handler should be deprecated in favor of the JWT handler. the fact whether someone is an admin should be included in the JWT payload.

solaris007: also, you cannot use a newly added method from IMS client without having a new version of IMS client released first and the dependency updated in the local package's dependencies.

packages/spacecat-shared-ims-client/src/utils.js
```diff
@@ -23,6 +23,15 @@ export const IMS_VALIDATE_TOKEN_ENDPOINT = '/ims/validate_token/v1';
...
 export const IMS_ADMIN_PROFILE_ENDPOINT = '/ims/admin_profile/v3';
+export const ADMIN_GROUP_IDENT = {
+  '8C6043F15F43B6390A49401A': [
+    635541219,
+  ],
+  '908936ED5D35CC220A495CD4': [
+    879529884,
+    901092291,
+  ],
+};
```
solaris007, line 34: what do these represent? can you document? this should probably not be hidden in a utils file.

packages/spacecat-shared-http-utils/src/auth/handlers/ims.js
solaris007: this implicitly calls `getImsUserOrganizations` - which will be called again for non-adobe users further below....

solaris007 (changes requested):

packages/spacecat-shared-http-utils/src/auth/handlers/ims.js
ravverma: >this will cause an additional request to IMS for every request incoming - also, the IMS handler should be deprecated in favor of the JWT handler. the fact whether someone is an admin should be included in the JWT payload.

Agreed on the extra request to IMS , I have fixed it and agreed on "IMS client" should not aware about constitutes an ASO admin, changed that as well, now no changes in the  IMS client.

packages/spacecat-shared-ims-client/src/utils.js
ravverma: yes removed

solaris007 (approved):

PR #818, opened 2025-06-23
fix: when the form doesn't have id/classname the views/submit numbers…

vdua:
… are incorrect

## Related Issues

Thanks for contributing!


packages/spacecat-shared-rum-api-client/src/functions/form-vitals.js
```diff
@@ -33,10 +36,6 @@ function filterEvents(bundles) {
   return bundles.map((bundle) => ({
     ...bundle,
     events: bundle.events.filter((event) => {
-      if (!CHECKPOINTS.includes(event.checkpoint)) {
-        return false;
-      }
```
vdua, line 38: This is redundant since this is handled by default

PR #820, opened 2025-06-24
feat: add Athena client

akshaymagapu:
add Athena client


ekremney (approved):
could you please add jsdocs comments to query and execute functions in the index.d.ts file?

PR #826, opened 2025-06-26
feat: introduce authoring type

dzehnder:
For resolving preview urls to a site if the site is on crosswalk, we need essentially a hlx config and a delivery type. To not mix these up and properly index them, we need an authoring type on top of that.

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/site.collection.js
+++ b/packages/spacecat-shared-data-access/src/models/site/site.collection.js
@@ -81,6 +81,8 @@ class SiteCollection extends BaseCollection {
     const previewType = getAuthoringType(hostname, Site.AUTHORING_TYPES);
 
     switch (previewType) {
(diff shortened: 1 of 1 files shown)


packages/spacecat-shared-data-access/src/models/site/site.model.js
```diff
@@ -19,10 +19,10 @@ export const AEM_CS_HOST = /^author-p(\d+)-e(\d+)/i;
 /**
  * Computes external IDs based on delivery type and configuration
  */
-export const computeExternalIds = (attrs) => {
+export const computeExternalIds = (attrs, authoringTypes) => {
   const { hlxConfig, deliveryConfig } = attrs;
 
-  if (hlxConfig) {
+  if (hlxConfig && authoringTypes.DA) {
```
alinarublea, line 25: Shouldn t you check here the authoring type of the site ?

PR #835, opened 2025-07-02
fix sqs submission of scrape job

hannessolo:
The scraper SQS queue expects an array of objects with a property `url`, like `[{url: 'https://example.com'}]`. Currently the scrape job supervisor is simply submitting an array of strings.

The effect of this is that the [spacecat api to start a new scrape job](https://opensource.adobe.com/spacecat-api-service/#tag/scrape/operation/createScrapeJob) fails because the scraper tries to extract the url property from a string. See screenshot.

See an example message in the scraper here: https://github.com/adobe/spacecat-content-scraper/blob/main/test/run-sqs.test.js#L268

@danieljchuser since you wrote this file, could you help me validate and test? I'm not sure how I can test this eg. by making the spacecat API use the PR version of this library?

Screenshot 2025-07-02 at 10 55 40

### DIFF
--- a/packages/spacecat-shared-scrape-client/src/clients/scrape-job-supervisor.js
+++ b/packages/spacecat-shared-scrape-client/src/clients/scrape-job-supervisor.js
@@ -148,7 +148,7 @@ function ScrapeJobSupervisor(services, config) {
     const message = {
       processingType,
       jobId: scrapeJob.getId(),
-      urls,
+      urls: urls.map((url) => ({ url })),
       customHeaders,
       options,
     };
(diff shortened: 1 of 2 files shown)


danieljchuser:
No change required, the wrong endpoint url was used.

hannessolo:
Thanks, that indeed worked. However, I copied that URL from the API docs is https://opensource.adobe.com/spacecat-api-service/#tag/scrape/operation/createScrapeJob 

Two concerns here:
1. The API docs should be updated to use the correct URL
2. The deprecated endpoint should be turned off and removed from the codebase. Leaving it in leads to wasted effort of debugging and creating PRs, as in this case.

PR #841, opened 2025-07-08
Paid traffic analysis import

cwjwisse:

## Related Issues

Thanks for contributing!


ekremney (changes requested):

packages/spacecat-shared-rum-api-client/src/functions/traffic-analysis.js
```diff
@@ -0,0 +1,59 @@
...
+
+    return {
+      ...trafficType,
+      url,
+      weight,
+      referrer,
+      utmSource,
+      utmMedium,
+      device: bundle.userAgent,
+      events: bundle.events,
```
cwjwisse, line 50: Currently passing on all events, as we may decide to use them at a later point. The importer will iterate once over all events and decide what to use for further processing.

The amount of data we pull from OpTel won't change since filtering is applied after fetching the data.

@ekremney, wdyt?

packages/spacecat-shared-rum-api-client/src/functions/traffic-analysis.js
```diff
@@ -0,0 +1,123 @@
...
+  return bundle.events
+    .find((e) => e.checkpoint === '404')
+    ?.source || null;
+}
+
+function getReferrer(bundle) {
+  return bundle.events
+    .find((e) => e.checkpoint === 'enter')
+    ?.source || null;
+}
```
ekremney, line 46: `document.referrer` also exists in `navigate` checkpoint. Please also check the source of the navigate checkpoint if exists: https://github.com/adobe/helix-rum-enhancer/blob/main/modules/index.js#L173

ekremney (approved):
LGTM, thanks for incorporating the changes! Please also address the following comments as well:

- Test fixtures do not contain all cases for consent checkpoint. Could you please update the test fixtures in a way that we can have all possible values of consent field in the `expected.json`?

PR #851, opened 2025-07-16
feat: add support for google CWV related APIs (PSI and CRUX)

shsteimer:

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-google-client/src/index.js
+++ b/packages/spacecat-shared-google-client/src/index.js
@@ -81,6 +81,7 @@ export default class GoogleClient {
         expiryDate: secrets.expiry_date,
         siteUrl: secrets.site_url,
         baseUrl: baseURL,
+        apiKey: context.env.GOOGLE_API_KEY,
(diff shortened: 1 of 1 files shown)


packages/spacecat-shared-google-client/src/index.js
```diff
@@ -32,6 +32,7 @@ export default class GoogleClient {
       config.clientSecret,
       config.redirectUri,
     );
+    this.apiKey = config.apiKey;
```
shsteimer, line 35: This may warrant a discussion is this is appropriate way to auth with these APIs or not.

My understanding is that the oath client used for the GSC apis that exist in this module uses OAUTH on behalf of the customer. In this case, I don't think we want to do that. Both of these APIs support API key auth, subject to some rate limits. Would request us to create and manage that Key, but that I assume should be simple enough. But happy to discuss and if using auth is the preferred route, we can do that.

https://developers.google.com/speed/docs/insights/v5/get-started#APIKey
https://developer.chrome.com/docs/crux/api#crux_api_key

PR #935, opened 2025-08-26
fix: get pageId from content-page-ref if content-page-id not available

rpapani:


packages/spacecat-shared-utils/src/aem.js
```diff
@@ -23,6 +23,8 @@ export const DELIVERY_TYPES = {
   OTHER: 'other',
 };
 
+const CONTENT_API_PAGES_RESOLVE_ENDPOINT = '/adobe/pages/resolve';
```
terinmez, line 26: Hi @rpapani , I think the official endpoints are /adobe/experimental/aspm-expires-20251231/pages/resolve and /adobe/experimental/pss/pages/resolve

terinmez (changes requested):

terinmez (approved):

packages/spacecat-shared-utils/src/aem.js
```diff
@@ -203,20 +205,49 @@ export function detectAEMVersion(htmlSource, headers = {}) {
...
+    let pageId = null;
+    if (authorURL && bearerToken) {
+      // try content-page-ref
+      const contentPageRefRegex = /<meta\s+name=['"]content-page-ref['"]\s+content=['"]([^'"]*)['"]\s*\/?>/i;
+      const contentPageRefMatch = html.match(contentPageRefRegex);
+      let contentPageRef = null;
+      if (contentPageRefMatch?.[1]?.trim()) {
+        contentPageRef = contentPageRefMatch[1].trim();
+        const referenceResolveResponse = await fetch(
+          `${authorURL}${CONTENT_API_PAGES_RESOLVE_ENDPOINT}?pageRef=${contentPageRef}`,
```
sagarmiglani, line 228: This is not supported in AMS yet. @terinmez I thought we had agreed to send the contentPageRef directly to PSS, since the PSS bundle can handle it. Could you clarify if that’s still the plan?

PR #936, opened 2025-08-26
feat: adding email id for trail users

ravverma:

## Related Issues

Thanks for contributing!


packages/spacecat-shared-http-utils/src/auth/handlers/ims.js
```diff
@@ -60,6 +60,8 @@ const transformProfile = (payload) => {
   const profile = { ...payload };
 
   profile.email = payload.user_id;
```
ravverma, line 62: this is already being used in `updatedBy` so can't touch it

PR #954, opened 2025-09-10
fix: filtering and keeping top 5 form internal navigation to reduce a…

05rahulsingh:
…udit size

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-rum-api-client/src/functions/form-vitals.js
+++ b/packages/spacecat-shared-rum-api-client/src/functions/form-vitals.js
@@ -295,6 +295,7 @@ function handler(bundles) {
 
   // populate internal navigation data
   populateFormsInternalNavigation(bundles, formVitals);
+
   // filter out pages with no form vitals
   const filteredFormVitals = Object.values(formVitals).filter(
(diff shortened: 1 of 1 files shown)


packages/spacecat-shared-rum-api-client/src/functions/form-vitals.js
```diff
@@ -322,6 +323,16 @@ function handler(bundles) {
     return formVitalCopy;
   });
 
+  // keeping on top 5 by page views count internal navigation
+  Object.values(updatedFormVitals).forEach((item) => {
+    if (item.forminternalnavigation) {
+      // eslint-disable-next-line max-len
+      item.forminternalnavigation.sort((a, b) => (b.totalClicksOnPage || 0) - (a.totalClicksOnPage || 0));
```
vdua, line 330: Currently we are picking based on the pageviews. Shall we not use the same approach here for sorting

https://github.com/adobe/spacecat-audit-worker/blob/main/src/forms-opportunities/formcalc.js#L159

05rahulsingh: done

vdua (approved):

PR #964, opened 2025-09-17
[WIP]feat: make fetchAllPages=true default to prevent data truncation

Kanishkavijay39:
- Changed BaseCollection to fetch all DynamoDB pages by default
- Prevents truncation issues when data exceeds 1MB query limit
- Added fetchAllPages option to QueryOptions interface with default true
- Added comprehensive tests for default and opt-out behavior
- Implements system-wide fix as per wiki recommendation (Option 1)

This addresses the Suggestion API pagination bug and applies the fix globally to prevent similar issues across all endpoints.

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/base/base.collection.js
+++ b/packages/spacecat-shared-data-access/src/models/base/base.collection.js
@@ -252,9 +252,11 @@ class BaseCollection {
       let result = await query.go(queryOptions);
       let allData = result.data;
 
-      // if the caller requests ALL pages and we're not using limit: 1,
-      // continue to fetch until there is no cursor.
-      if (options.fetchAllPages && options.limit !== 1) {
+      // By default, fetch ALL pages unless explicitly disabled or using limit: 1
+      // This ensures complete data retrieval to prevent truncation issues
+      const shouldFetchAllPages = options.fetchAllPages !== false && options.limit !== 1;
+
+      if (shouldFetchAllPages) {
         while (result.cursor) {
           queryOptions.cursor = result.cursor;
           // eslint-disable-next-line no-await-in-loop
--- a/packages/spacecat-shared-data-access/src/models/base/index.d.ts
+++ b/packages/spacecat-shared-data-access/src/models/base/index.d.ts
@@ -34,6 +34,11 @@ export interface QueryOptions {
   limit?: number;
   order?: string;
   attributes?: string[];
+  /** 
+   * Whether to automatically fetch all pages of results. Defaults to true.
+   * Set to false to only fetch the first page (for performance optimization).
+   */
+  fetchAllPages?: boolean;
 }
 
 export interface BaseCollection<T extends BaseModel> {
--- a/packages/spacecat-shared-data-access/test/unit/models/base/base.collection.test.js
+++ b/packages/spacecat-shared-data-access/test/unit/models/base/base.collection.test.js
@@ -797,6 +797,50 @@ describe('BaseCollection', () => {
       const secondCallArgs = goStub.secondCall.args[0];
       expect(secondCallArgs).to.deep.include({ order: 'desc', cursor: 'key1' });
     });
+
+    it('fetches all pages by default (system-wide fix)', async () => {
+      const firstResult = { data: [mockRecord], cursor: 'key1' };
+      const secondRecord = { id: '2', foo: 'bar' };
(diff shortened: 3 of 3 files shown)


packages/spacecat-shared-data-access/src/models/base/base.collection.js
```diff
@@ -252,15 +252,37 @@ class BaseCollection {
...
+          pageCount += 1;
           queryOptions.cursor = result.cursor;
           // eslint-disable-next-line no-await-in-loop
           result = await query.go(queryOptions);
           allData = allData.concat(result.data);
         }
+
+        // Logging for observation of pagination behavior - can be removed if not needed
+        if (pageCount > 1) {
+          this.log.info(`Pagination completed for ${this.entityName}`, {
```
iuliag, line 273: This might lead to excessive logging to Coralogix, where we have a quota limit.
Maybe you could decrease the log level to debug (these logs would not be forwarded to Coralogix iirc, but could still be consulted via AWS CloudWatch)?

iuliag (approved):

solaris007:
alternatively, let's callers simply opt in, for example:
```js
// before
async allSitesToAudit() {
  return (await this.all({}, { attributes: ['siteId'] })).map((site) => site.getId());
}

// opt-in
async allSitesToAudit({ fetchAllPages = true } = {}) {
  const sites = await this.all({}, { attributes: ['siteId'], fetchAllPages });
  return sites.map((site) => site.getId());
}
```

packages/spacecat-shared-data-access/src/models/base/base.collection.js
```diff
@@ -252,15 +252,44 @@ class BaseCollection {
...
 
-      // if the caller requests ALL pages and we're not using limit: 1,
-      // continue to fetch until there is no cursor.
-      if (options.fetchAllPages && options.limit !== 1) {
+      // By default, fetch ALL pages unless explicitly disabled or using limit: 1
+      // This ensures complete data retrieval to prevent truncation issues
+      const shouldFetchAllPages = options.fetchAllPages !== false && options.limit !== 1;
+
+      // To see logs on CloudWatch, then I will remove
+      this.log.info(`🔍 PAGINATION CHECK: ${this.entityName} - records: ${result.data.length}, hasCursor: ${!!result.cursor}, fetchAll: ${shouldFetchAllPages}`);
```
solaris007, line 260: change to debug

Kanishkavijay39:
> * Your loop logic is correct and aligns with ElectroDB’s cursor semantics. However, flipping the default to “paginate everything” at this layer changes behavior: callers that pass limit (expecting at most N results) will now get unbounded results across pages. That’s the main risk.
> * Suggest tightening the default so you only auto-paginate when the caller didn’t specify a limit, or they explicitly opted in.
> * Also move noisy logs to debug and keep the truncation warning.

Tightened the default pagination logic
Collection.all() → Auto-paginate (no limit specified, fixes truncation)
Collection.all({}, { limit: 10 }) → Respects limit, returns max 10 records (no unbounded results)
Collection.all({}, { fetchAllPages: true }) → Explicit opt-in works
Collection.all({}, { fetchAllPages: true, limit: 10 }) → Explicit opt-in overrides limit

On truncation warnings: I removed them to reduce noise, since the new logic is much safer and users are making conscious choices (either setting limits or opting out). If you think they’re still valuable for spotting potential data loss, I can add it again.

packages/spacecat-shared-data-access/src/models/base/base.collection.js
Kanishkavijay39: Hi @solaris007 , I just tested this flow with api service. The query I ran was:
query i runned: fields @timestamp, @message
| filter @message like "PAGINATION CHECK" 
| sort @timestamp desc
| limit 20

But I’m still not able to see the log. Even when I add the same string in the message filter in CloudWatch Log Insights, nothing shows up.
Do you have any idea what could be causing this?
[Since now I rolledback but didn't get why my logs was not visible]
api service PR: https://github.com/adobe/spacecat-api-service/pull/1269/files

Kanishkavijay39: done.

solaris007: debug logs only show up in cloudwatch, not coralogix

PR #967, opened 2025-09-18
fix: need to switch to /adobe/pages because experimental endpoints do not exist in AMS

terinmez:
fix: need to switch to /adobe/pages because experimental endpoints do not exist in AMS

### DIFF
--- a/packages/spacecat-shared-utils/src/aem-content-api-utils.js
+++ b/packages/spacecat-shared-utils/src/aem-content-api-utils.js
@@ -12,7 +12,8 @@
 
 import { fetch } from './adobe-fetch.js';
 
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-utils/src/aem-content-api-utils.js
```diff
@@ -12,7 +12,8 @@
 
 import { fetch } from './adobe-fetch.js';
 
-const CONTENT_API_PREFIX = '/adobe/experimental/aspm-expires-20251231';
+export const CONTENT_API_PREFIX = '/adobe';
+// export const CONTENT_API_PREFIX = '${CONTENT_API_PREFIX}';
```
rpapani, line 16: you can get rid of the commented line

rpapani (approved):

PR #985, opened 2025-09-25
feat: Add Query for fetching RUM engagement metrics

dipratap:

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-rum-api-client/README.md
+++ b/packages/spacecat-shared-rum-api-client/README.md
@@ -437,6 +437,28 @@ An example response:
 ]
 ```
 
+### engagement
+
+Calculates engagement metrics for all pages from RUM data. A page view is considered engaged if there has been at least some user interaction (click events) or significant content has been viewed (4 or more viewmedia or viewblock events).
+Ref. - https://github.com/adobe/rum-distiller/blob/22f8b3caa6d700f4d1cbe29a94b7da34b9d50764/series.js#L89
+
+An example response:
+
+```json
+[
+  {
+    "url": "https://www.example.com/home",
+    "totalTraffic": 5000,
+    "engagementPercentage": 50
+  },
+  {
+    "url": "https://www.example.com/about",
+    "totalTraffic": 2000,
+    "engagementPercentage": 40
+  }
+]
+```
+
 ## Linting
 Lint the codebase using:
 ```
--- a/packages/spacecat-shared-rum-api-client/src/functions/engagement.js
+++ b/packages/spacecat-shared-rum-api-client/src/functions/engagement.js
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2025 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


packages/spacecat-shared-rum-api-client/src/functions/user-engagement.js
```diff
@@ -0,0 +1,60 @@
...
+ * governing permissions and limitations under the License.
+ */
+
+const hasEngagement = (bundle) => {
+  const clickEngagement = bundle.events.filter((evt) => evt.checkpoint === 'click').length > 0
+    ? bundle.weight
+    : 0;
+  const contentEngagement = bundle.events
+    .filter((evt) => evt.checkpoint === 'viewmedia' || evt.checkpoint === 'viewblock')
+    .length > 3
```
prithipalpatwal, line 19: why greater than 3?

packages/spacecat-shared-rum-api-client/src/functions/user-engagement.js
```diff
@@ -0,0 +1,60 @@
...
+ * or significant content has been viewed (4 or more viewmedia or viewblock events).
+ * Ref. - https://github.com/adobe/rum-distiller/blob/22f8b3caa6d700f4d1cbe29a94b7da34b9d50764/series.js#L89
+ *
+ * @param {Array} bundles - The RUM bundles to calculate engagement metrics for.
+ * @returns {Array} An array of engagement metrics for each page.
+ */
+function handler(bundles) {
+  const urlsData = {};
+  bundles.forEach((bundle) => {
+    const engagementTraffic = hasEngagement(bundle) ? bundle.weight : 0;
```
prithipalpatwal, line 37: The ternary operator isn’t needed here, as hasEngagement() already returns either bundle.weight or 0.

packages/spacecat-shared-rum-api-client/src/functions/user-engagement.js
```diff
@@ -0,0 +1,60 @@
...
+      ) * 100;
+    }
+  });
+
+  return Object.values(urlsData);
+}
+
+export default {
+  handler,
+  checkpoints: ['click', 'viewmedia', 'viewblock'],
```
prithipalpatwal, line 59: Should we include "viewblock" as part of engagement? It mainly reflects visibility rather than active engagement. Should we also add "scroll" and "viewfragment", or keep engagement for active actions only?

packages/spacecat-shared-rum-api-client/src/functions/user-engagement.js
dipratap: These are the standard values decided by RUM team to measure user engagement. Any changes will cause deviation from what we see on RUM.
https://github.com/adobe/rum-distiller/blob/22f8b3caa6d700f4d1cbe29a94b7da34b9d50764/series.js#L89

packages/spacecat-shared-rum-api-client/src/functions/user-engagement.js
dipratap: Sure, updated.

PR #994, opened 2025-10-01
fix: remove all debug logs from data access

martinst06:
Removing all debug logs from data access

## Related Issues

EPIC: [SITES-34661](https://jira.corp.adobe.com/browse/SITES-34661)

### DIFF
--- a/packages/spacecat-shared-data-access/src/service/index.js
+++ b/packages/spacecat-shared-data-access/src/service/index.js
@@ -32,14 +32,20 @@ const createRawClient = (client = undefined) => {
   });
 };
 
-const createElectroService = (client, config) => {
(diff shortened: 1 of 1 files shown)


packages/spacecat-shared-data-access/src/service/index.js
```diff
@@ -32,20 +32,14 @@ const createRawClient = (client = undefined) => {
...
-    log.debug(JSON.stringify(event, null, 4));
-  };
-  /* c8 ignore end */
 
   return new Service(
     EntityRegistry.getEntities(),
     {
       client,
       table,
-      logger,
```
ekremney, line 48: keep the logger parameter. It could be that service logging other levels like info

PR #1469, opened 2026-03-25
feat: http utils read only org ims gate

ravverma:
## Summary

Adds read-only org / read-only admin behavior in `@adobe/spacecat-shared-http-utils`: shared LaunchDarkly feature flag, IMS validation gate for read-only orgs, JWT claims support, route helpers, S2S wrapper updates, and `readOnlyAdminWrapper` for LD + read/write route enforcement.

Coordinates with **spacecat-auth-service**: [adobe/spacecat-auth-service#514](https://github.com/adobe/spacecat-auth-service/pull/514).

## Changes

- **`constants.js`** — `FF_READ_ONLY_ADMIN` (`FT_LLMO-3008`) as single source for LD key
- **`handlers/ims.js`** — After profile/orgs, evaluate LD for first org; when flag is on, log (warn) and treat as unauthorized (client sees generic failure; existing catch logs detail)
- **`read-only-admin-wrapper.js`** — `LaunchDarklyClient.createFrom(context)`; 403 for disallowed methods when RO admin + flag on
- **`route-utils.js`** — Route read/write classification for wrappers
- **`handlers/jwt.js`**, **`auth-info.js`**, **`s2s-wrapper.js`** — Read-only admin claims / capability alignment
- **Exports** — `index.js` + `index.d.ts` for new APIs
- **Tests** — `ims`, `read-only-admin-wrapper`, `route-utils` (esmock for LaunchDarkly)
- **Dependencies** — `@adobe/spacecat-shared-launchdarkly-client`; dev `esmock`; `package-lock.json` updated

## Consumer notes

- Bump `@adobe/spacecat-shared-http-utils` in services that use IMS / admin wrappers after release.
- Align rollout with auth-service **#514** (flag + behavior).

## Test plan

- [x] `npm run lint -w packages/spacecat-shared-http-utils`
- [ ] `npm test -w packages/spacecat-shared-http-utils` (run in CI / locally)


solaris007 (changes requested):
Hey @ravverma,

## Strengths

- **Correct fail-mode asymmetry.** The IMS handler fails open (LD unavailable = allow auth, preserving existing behavior) while the `readOnlyAdminWrapper` fails closed (LD unavailable = deny RO admin access). This is the right default for each trust boundary.
- **JWT mutual exclusion guard** (`jwt.js:44-47`). Rejecting tokens that carry both `is_admin` and `is_read_only_admin` prevents privilege confusion from malformed or tampered tokens. Clean and explicit.
- **Deny-by-default for unmapped routes.** When `resolveRouteCapability` returns null, `action` is undefined (not `'read'`), so unmapped routes are blocked for RO admins. Correct security posture, tested explicitly.
- **Clean route-utils extraction.** Lifting `matchRoute`, `resolveRouteCapability`, and `guardNonEmptyRouteCapabilities` out of `s2s-wrapper.js` into a shared module is well-scoped refactoring that immediately pays for itself.
- **Thorough test coverage.** Both fail-open (IMS: null LD client, LD throws, missing ident) and fail-closed (wrapper: null LD client, LD throws, empty tenants) paths are individually tested. The esmock approach for isolating LD is clean.

## Issues

### Critical (Must Fix)

**1. IMS gate blocks ALL users in a flagged org, not just read-only admins** (`ims.js:199-202`)

The `isReadOnlyAdminEnabled` check runs before `isUserASOAdmin` and before scope assignment. When the flag is enabled for an org, every user in that org - regular customers, non-admin users, even ASO admins - is blocked with `throw new Error('Unauthorized')`. The test "returns null for ASO admin when org FF is enabled" confirms this.

If the intent is to force flagged orgs onto the JWT auth path (via auth-service #514), this needs explicit documentation in the code. If the intent is only to gate read-only admin access, the check needs to move after admin determination.

**Why it matters:** Enabling this flag for an org silently breaks IMS-based access for all users in that org across all routes. The function name `isReadOnlyAdminEnabled` is misleading about the blast radius.

**Suggested fix:** Either (a) rename to `isOrgBlockedFromImsAuth` and add a clear comment explaining this is intentional and that auth-service #514 provides the alternative JWT path, or (b) move the check to only affect users who should be read-only admins.

### Important (Should Fix)

**2. `readOnlyAdminWrapper` silently passes through RO admins when `routeCapabilities` is omitted** (`read-only-admin-wrapper.js:86`)

When `routeCapabilities` is `undefined` or `null`, the `isObject()` check is false, so the entire write-blocking logic is skipped. An RO admin passes through with no restrictions - only an audit log entry. The wrapper's name and JSDoc promise write-blocking, but omitting `routeCapabilities` silently disables enforcement.

Compare with `guardNonEmptyRouteCapabilities`, which catches `{}` but not `undefined`. A consumer who calls `readOnlyAdminWrapper(fn)` without options gets audit logging but zero authorization enforcement.

**Suggested fix:** Make `routeCapabilities` required - throw at creation time if it is falsy, matching the empty-object guard pattern.

**3. First-org-only flag evaluation is fragile for multi-org users** (`ims.js:117`, `read-only-admin-wrapper.js:43`)

Both `isReadOnlyAdminEnabled` and `evaluateFeatureFlag` evaluate the flag against only `organizations[0]` / `tenantIds[0]`. A user belonging to multiple IMS orgs where only some are read-only-flagged will get inconsistent behavior depending on array ordering, which is not guaranteed to be stable.

**Suggested fix:** Evaluate the flag against all orgs. For the IMS gate (block path), return true if any org is flagged. For the wrapper, resolve the specific org from context. Alternatively, document this as a known limitation with rationale.

**4. Misleading JSDoc says "fail-closed" but IMS handler is fail-open** (`ims.js:107`)

The JSDoc on `isReadOnlyAdminEnabled` says "Fail-closed: returns false when the LD client is unavailable." Returning false means "not read-only, allow through" - that is fail-open with respect to the blocking behavior. The wrapper's `evaluateFeatureFlag` has the same comment, but there false results in a 403 (actually fail-closed). The comment is correct for the wrapper but wrong for the IMS handler.

**Suggested fix:** Change to "Fail-open: returns false (allowing authentication) when the LD client is unavailable or evaluation errors."

**5. Audit log lacks user identity** (`read-only-admin-wrapper.js:97`)

The log entry `[ro-admin-audit] RO admin accessed: GET /sites` records method and path but not who performed the action. For a role with reduced privileges, knowing WHO accessed WHAT is essential for incident investigation.

**Suggested fix:** Include the user's email or org ID from `authInfo`, e.g. `log.info({ tag: 'ro-admin-audit', method, path: suffix, org: tenantIds[0] })`.

**6. Log injection surface via `context.pathInfo.suffix`** (`read-only-admin-wrapper.js:92,97`)

`suffix` comes from the HTTP request path. If the logging framework does not sanitize newlines, an attacker can inject fake log lines (e.g., `GET /sites%0A[ro-admin-audit] RO admin accessed: DELETE /sites/victim-id`), polluting the audit trail that is the primary observability for this feature.

**Suggested fix:** Use structured log fields instead of string interpolation: `log.info({ tag: 'ro-admin-audit', method, suffix })`, or sanitize `suffix` before interpolation.

**7. Product-specific flag key in shared library** (`constants.js:13`)

The flag key `FT_LLMO-3008` is LLMO-product-specific, but this code lives in `spacecat-shared-http-utils`, consumed by all SpaceCat services. This couples the shared auth layer to LLMO's feature management namespace.

**Suggested fix:** Either accept the flag key as a configuration parameter (via context or wrapper options) or rename the flag to something product-neutral like `FT_READ_ONLY_ORG`.

### Minor (Nice to Have)

**8. Em-dash in JWT log message** (`jwt.js:45`): `'Token has both is_admin and is_read_only_admin — rejecting'` uses an em-dash. The codebase convention uses regular dashes.

**9. Capability string parsing contract is undocumented** (`read-only-admin-wrapper.js:88`): `capability?.split(':').pop()` assumes `'resource:action'` format but this is not documented. A bare value like `'readonly'` would be blocked since it does not equal `'read'`.

**10. Information disclosure in 403 responses**: The two distinct messages ("Read-only admin access is not enabled" vs "Read-only admin users cannot perform write operations") reveal feature flag state and role classification. Consider returning a single generic "Forbidden" message and logging the specific reason server-side only.

## Recommendations

- **Document the two-gate architecture.** The interaction between the IMS handler gate (blocks direct IMS auth for RO orgs) and the wrapper gate (enforces read-only on JWT-authenticated RO admins) is the key design decision. Neither the code comments nor the PR description explain this end-to-end flow clearly.
- **Consider a shared `evaluateLDFlag(context, orgIdent, flagKey)` utility.** The LD client creation + org suffix + try/catch pattern is now in two places with different fail modes. A single function with a `failDefault` parameter would reduce the surface for bugs.
- **Coordinate rollout documentation with auth-service #514.** The blast radius of enabling `FT_LLMO-3008` needs to be crystal clear since it blocks all IMS auth for the flagged org.

## Assessment

The security design is sound - fail-mode polarity is correct, JWT mutual exclusion is good, deny-by-default on unmapped routes is correct. However, Critical #1 (IMS gate blocks all users, not just RO admins) needs clarification or a code change - the naming and documentation do not match the behavior's blast radius. Important #2 (wrapper passthrough without routeCapabilities) is a misconfiguration trap that could silently grant RO admins full write access. Important #3 (first-org-only evaluation) is a real risk for multi-org users.

## Next Steps

1. Clarify or fix the IMS gate scope (Critical #1) - document if intentional, fix if not.
2. Make `routeCapabilities` required in the wrapper (Important #2).
3. Address first-org-only evaluation or document as accepted limitation (Important #3).
4. Fix the misleading JSDoc and add user identity to audit logs (Important #4, #5).
5. Minor items are optional but recommended.

packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
```diff
@@ -0,0 +1,120 @@
...
+}
+
+/**
+ * Read-only admin authorization wrapper for the helix-shared-wrap `.with()` chain.
+ *
+ * After successful authentication (authInfo already set on context by an earlier
+ * wrapper), this wrapper checks whether the authenticated user is a read-only admin.
+ * If so it:
+ *
+ * 1. Evaluates the `FT_LLMO-3008` LaunchDarkly feature flag (fail-closed).
```
solaris007: **Important:** Stale reference - still says `FT_LLMO-3008` and "fail-closed". Should be `FT_READ_ONLY_ORG`. Also, the `@param` on line 68 still marks `routeCapabilities?` as optional - should drop the `?` since it is now required.

solaris007 (changes requested):
Hey @ravverma,

Thanks for the thorough follow-up - all 10 issues from the prior review have been addressed.

## Strengths

**Previously flagged issues now addressed:**

- **Critical #1 (IMS gate scope):** `isReadOnlyAdminEnabled` renamed to `isOrgBlockedFromImsAuth` with clear JSDoc and inline comments explaining the intentional org-wide block. Well done.
- **Important #2 (routeCapabilities passthrough):** Now throws at creation time if falsy. Fail-fast at wrapper registration, not at request time - exactly right.
- **Important #3 (first-org-only):** Documented as an accepted limitation with rationale. Pragmatic given IMS org ordering instability and the fact that per-org scope filtering is a larger change.
- **Important #4 (misleading JSDoc):** Corrected to "Fail-open" with accurate behavior description.
- **Important #5 (audit log identity):** Org identity now included via structured logging. Partial fix - see Minor #3 below.
- **Important #6 (log injection):** Fully resolved. Switched from string interpolation to structured log objects (`{ tag, method, suffix, org }`).
- **Important #7 (product-specific flag key):** Renamed to `FT_READ_ONLY_ORG` / `FF_READ_ONLY_ORG`. Generic, appropriate for a shared library.
- **Minor #8 (em-dash in JWT log):** Fixed.
- **Minor #9 (capability parsing):** Documented with inline comment explaining `resource:action` format.
- **Minor #10 (info disclosure):** Both 403 paths now return generic `'Forbidden'`.

The structured logging migration across the wrapper is a nice improvement - consistent `{ tag, method, suffix, org }` shape makes these logs queryable.

## Issues

### Important (Should Fix)

**1. TypeScript definition still marks `routeCapabilities` as optional** (`index.d.ts:46-49`)

The runtime now throws if `routeCapabilities` is missing, but the `.d.ts` declaration has both `opts?` and `routeCapabilities?` as optional. TypeScript consumers will get no compile-time warning and will hit a runtime crash instead.

**Suggested fix:**
```typescript
export function readOnlyAdminWrapper(
  fn: Function,
  opts: { routeCapabilities: Record<string, string> },
): Function;
```

**2. Stale references in wrapper JSDoc** (`read-only-admin-wrapper.js:60,68`)

The function-level JSDoc still references `FT_LLMO-3008` and says "fail-closed" on line 60. The `@param` tag on line 68 still marks `routeCapabilities?` as optional. Both are stale after the fix commit.

**Suggested fix:** Update to `FT_READ_ONLY_ORG`, and remove the `?` from the `@param` to match the new required behavior.

### Minor (Nice to Have)

**3. Audit log still lacks user identity** (`read-only-admin-wrapper.js:97`): The `org` field is a good addition, but the `ro-admin-audit` entry still doesn't identify *which* RO admin user performed the action. "Someone in org X accessed GET /sites" is less useful than "user Y in org X accessed GET /sites." Worth a follow-up.

**4. Em-dashes in new ims.js text** (`ims.js:108,126`): The fix commit introduced em-dashes in the new JSDoc and inline comment. Should use regular dashes per codebase convention.

**5. LD flag key rename requires coordination**: The rename from `FT_LLMO-3008` to `FT_READ_ONLY_ORG` means the new key must exist in LaunchDarkly before deployment. In the wrapper path, a missing flag causes `evaluateFeatureFlag` to return `false`, which results in a 403 for every RO admin user. Please confirm the LD flag has been created or is planned before merge.

## Assessment

Ready to merge? **With minor fixes.**

All prior Critical and Important issues are resolved. The remaining items are small: update the TypeScript definition to match the new required contract (#1), fix the stale JSDoc (#2), and confirm the LD flag rename is coordinated (#5). The code quality and security design are solid.

## Next Steps

1. Fix the TypeScript definition (Important #1) and stale JSDoc (Important #2) - quick changes.
2. Confirm the `FT_READ_ONLY_ORG` flag exists in LaunchDarkly (Minor #5).
3. Em-dashes and audit log user identity are optional follow-ups.

solaris007 (approved):
Hey @ravverma,

All remaining items from the previous review are addressed:

## Strengths

- TypeScript definition now correctly marks both `opts` and `routeCapabilities` as required (`index.d.ts:48`) - matches the runtime contract.
- Wrapper JSDoc updated: stale `FT_LLMO-3008` replaced with `FT_READ_ONLY_ORG`, `@param` no longer marks `routeCapabilities` as optional (`read-only-admin-wrapper.js:60,68`).
- Audit log entries now include `email: authInfo.getProfile?.()?.email` across all three structured log calls - the `ro-admin-audit` trail now identifies the specific user, not just the org.
- Em-dashes in `ims.js` JSDoc and inline comment replaced with regular dashes.

Clean follow-up. No new issues.

## Assessment

Ready to merge? **Yes.**

All findings from both prior review rounds have been addressed. The security design is sound, test coverage is thorough, and the code is well-documented. Just confirm the `FT_READ_ONLY_ORG` flag exists in LaunchDarkly before deploying.

PR #1501, opened 2026-04-02
feat: LLMO-3959 changing schema to add detectCdn flag

AddyKen-ghost:
## Summary

Add support for persisting the auto-detected CDN provider inside `config.llmo.detectedCdn` on the Site model, as part of the LLMO onboarding CDN auto-detection feature (LLMO-3959).

### Changes

- **Joi schema**: Add `detectedCdn: Joi.string().valid('aem-cs-fastly', 'other').optional()` to the `llmo` config object
- **Config model**: Add `getLlmoDetectedCdn()` getter (returns `null` when absent) and `updateLlmoDetectedCdn()` setter
- **Site schema**: Remove the previously added top-level `detectedCdn` attribute — the value is now nested inside the existing `config` JSONB column, avoiding a database migration
- **Tests**: Add 4 unit tests for the new getter/setter following the existing pattern (CdnBucketConfig, CountryCodeIgnoreList)

### Three-state values

| Value | Meaning |
|-------|---------|
| `'aem-cs-fastly'` | DNS confirmed the site uses AEM CS Fastly |
| `'other'` | DNS resolved successfully but did not match AEM CS Fastly |
| `null` / absent | DNS lookup failed — inconclusive |

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/config.js
+++ b/packages/spacecat-shared-data-access/src/models/site/config.js
@@ -352,6 +352,7 @@ export const configSchema = Joi.object({
       cdnProvider: Joi.string().optional(),
       region: Joi.string().pattern(AWS_REGION_PATTERN).optional(),
     }).optional(),
+    detectedCdn: Joi.string().valid('aem-cs-fastly', 'other').optional(),
   }).optional(),
   cdnLogsConfig: Joi.object({
     bucketName: Joi.string().required(),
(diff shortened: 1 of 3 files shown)


packages/spacecat-shared-data-access/src/models/site/config.js
```diff
@@ -345,6 +352,7 @@ export const configSchema = Joi.object({
       cdnProvider: Joi.string().optional(),
       region: Joi.string().pattern(AWS_REGION_PATTERN).optional(),
     }).optional(),
+    detectedCdn: Joi.string().valid('aem-cs-fastly', 'other').optional(),
```
akshaymagapu, line 355: small nitpick:

`detectedCdn` seems generic here right?
as it is checking only if the site on AEM CS Fastly or not.

could `isAemCsFastly` works better here?

akshaymagapu (approved):

packages/spacecat-shared-data-access/src/models/site/config.js
AddyKen-ghost: yes. it's only currently supporting that but for future we will be extending the logic to detect other cdns as well. Hence the name.

PR #1526, opened 2026-04-09
fix: add method for fetching org scoped service principal token 

dipratap:


packages/spacecat-shared-ims-client/src/clients/ims-client.js
```diff
@@ -226,6 +226,38 @@ export default class ImsClient extends ImsBaseClient {
     return this.serviceAccessTokenV3;
   }
 
+  async getServiceAccessTokenOrgScopedV3(imsOrgId) {
```
nit23uec: I would explictly call it `getServicePrincipalAccessToken(imsOrgId) `
SP is always org scoped, so no need to mention that explicitly
Service Token is a different terminology in IMS  which should not be used here.

nit23uec (approved):

PR #1528, opened 2026-04-10
fix(seo-client): add prefix URL filter to getTopPages

rpapani:
## Summary

- **Problem**: `getTopPages()` was returning subdomain URLs (e.g., `blog.example.com`, `shop.example.com`) for a given domain. Downstream consumers like the meta-tags audit incorrectly map these subdomain URLs onto the base domain, leading to 404s and invalid audit results.
- **Fix**: Added a `Bw` (begins with) display filter to the SEO API call so results are scoped to the intended URL prefix. Also added client-side filtering for non-www prefixes where the API filter is unreliable.
- The `url` parameter now acts as a **prefix URL** (e.g., `https://www.example.com` or `example.com/us`), not just a plain domain.

### www prefixes (e.g., `https://www.example.com`)
The API-level `Bw` filter on the `Ur` (URL) field reliably restricts results to the correct hostname. No client-side filtering needed.

### Non-www prefixes (e.g., `https://example.com`, `example.com/us`)
The SEO API `Bw` filter is unreliable for non-www domains because it can match subdomains (e.g., `https://example.com` also matches `https://example.com.au`). To handle this:
1. The method over-fetches (2x the requested limit)
2. Applies client-side prefix filtering, keeping only URLs that match exactly or start with `{prefix}/`
3. If the filtered set cannot meet the requested limit, a warning or debug message is logged

### Unit tests
- [x] All 132 existing tests pass
- [x] New tests added for www and non-www prefix filtering
- [x] Test for subpath prefix filtering (e.g., `example.com/us`)
- [x] Test for warning when client-side filtering cannot meet limit
- [x] Test for debug message when provider has fewer matching pages
- [x] 100% code coverage maintained (lines, branches, statements, functions)

### Dev validation with real customer data
- [x] Deploy bumped seo-client to import worker dev branch
- [x] Trigger top-pages import for a site with `overrideBaseURL` (www prefix) — verify only www pages are returned, no subdomain pages
- [x] Trigger top-pages import for a site without `overrideBaseURL` (non-www prefix) — verify subdomain pages are filtered out client-side
- [x] Trigger top-pages import for a site with subpath in baseURL — verify only pages under that subpath are returned
- [x] Compare top-pages count before and after to ensure we still meet the 200-page target for www sites
- [x] Verify no 404 URLs appear in the imported top pages by spot-checking a sample of URLs

### DIFF
--- a/packages/mysticat-shared-seo-client/src/client.js
+++ b/packages/mysticat-shared-seo-client/src/client.js
@@ -158,8 +158,8 @@ export default class SeoClient {
     }
 
     // Ensure url has protocol for the Bw filter and extract hostname for the API.
-    // Input is a prefix URL, either full (https://www.veeam.com) or protocol-stripped
-    // (www.veeam.com, celestyal.com/gb).
+    // Input is a prefix URL, either full (https://www.example.com) or protocol-stripped
+    // (www.example.com, example.com/us).
     const prefixUrl = url.includes('://') ? url : `https://${url}`;
     let domain;
     try {
--- a/packages/mysticat-shared-seo-client/src/client.js
+++ b/packages/mysticat-shared-seo-client/src/client.js
@@ -157,14 +157,41 @@ export default class SeoClient {
       throw new Error(`Invalid URL: ${url}`);
     }
 
+    // Ensure url has protocol for the Bw filter and extract hostname for the API.
+    // Input is a prefix URL, either full (https://www.veeam.com) or protocol-stripped
+    // (www.veeam.com, celestyal.com/gb).
+    const prefixUrl = url.includes('://') ? url : `https://${url}`;
+    let domain;
+    try {
+      domain = new URL(prefixUrl).hostname;
+    } catch {
+      domain = url;
+    }
+
+    const isWww = prefixUrl.includes('://www.');
     const ep = ENDPOINTS.topPages;
     const epKw = ENDPOINTS.topPagesKeywords;
-    const effectiveLimit = getLimit(limit, 2000);
+
+    // For non-www prefixes, the Bw filter is unreliable (matches subdomains),
+    // so we over-fetch and filter client-side.
+    const requestLimit = !isWww
+      ? getLimit(limit * 2, 2000)
+      : getLimit(limit, 2000);
 
     const commonParams = {
-      domain: url,
+      domain,
       database: DEFAULT_DATABASE,
     };
+
+    // Build display_filter: traffic > 0, plus Bw prefix filter
+    const filters = [
+      {
+        sign: '+', field: 'Tg', op: 'Gt', value: '0',
+      },
+      {
+        sign: '+', field: 'Ur', op: 'Bw', value: prefixUrl,
+      },
+    ];
 
     // Two calls required: the SEO data provider does not offer a top-keyword-per-page
     // field in its page-level report. Sorting organic keywords by traffic and grouping
@@ -173,17 +200,15 @@ export default class SeoClient {
       this.sendRawRequest({
         type: ep.type,
         ...commonParams,
-        display_limit: effectiveLimit,
+        display_limit: requestLimit,
         export_columns: ep.columns,
-        display_filter: buildFilter([{
-          sign: '+', field: 'Tg', op: 'Gt', value: '0',
-        }]),
+        display_filter: buildFilter(filters),
         ...ep.defaultParams,
       }, ep.path),
       this.sendRawRequest({
         type: epKw.type,
         ...commonParams,
-        display_limit: effectiveLimit * 3,
+        display_limit: requestLimit * 3,
         export_columns: epKw.columns,
         ...epKw.defaultParams,
       }, epKw.path),
@@ -200,11 +225,27 @@ export default class SeoClient {
       }
     }
 
-    const pages = pageRows.map((row) => ({
+    let pages = pageRows.map((row) => ({
       url: row.Ur,
       sum_traffic: coerceValue(row.Tg, 'int'),
       top_keyword: keywordMap.get(row.Ur) ?? null,
     }));
+
+    // For non-www prefixes, apply client-side filtering since Bw is unreliable
+    if (!isWww) {
+      const filteredPages = pages.filter(
+        (page) => page.url === prefixUrl
+          || page.url.startsWith(`${prefixUrl}/`),
+      );
+
+      if (filteredPages.length < limit && pages.length >= requestLimit) {
+        this.log.warn(`[SEO] Could not meet ${limit} top pages for ${prefixUrl} after requesting ${requestLimit} (got ${filteredPages.length} matching)`);
+      } else if (filteredPages.length < limit) {
+        this.log.debug(`[SEO] Provider has only ${filteredPages.length} pages matching prefix ${prefixUrl}`);
+      }
+
+      pages = filteredPages.slice(0, limit);
+    }
 
     return {
       result: { pages },
--- a/packages/mysticat-shared-seo-client/README.md
+++ b/packages/mysticat-shared-seo-client/README.md
@@ -46,7 +46,7 @@ Returns the top organic pages for a given URL prefix, sorted by traffic.
 
 The `url` parameter acts as a prefix filter, not just a plain domain. This ensures that only pages belonging to the intended hostname (and optional path prefix) are returned. For example, passing `https://www.example.com` excludes pages from subdomains like `blog.example.com` or `shop.example.com`.
 
-- **www prefixes** (e.g., `https://www.example.com`): The method applies a server-side `Bw` (begins with) display filter on the SEMrush API call. This reliably scopes results to the correct hostname.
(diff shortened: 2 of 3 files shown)


solaris007 (changes requested):
Hey @rpapani,

Good fix for a real problem - subdomain URLs leaking into downstream audits and causing 404s is the kind of silent data-quality issue that erodes trust in the pipeline. The dual-strategy approach (server-side Bw for www, client-side filter for non-www) is pragmatic, and the documentation and test coverage are both solid. A few things to work through before this ships.

## Strengths

- **Right layer for this fix.** Filtering at the shared SEO client prevents every downstream consumer from independently dealing with subdomain pollution. (client.js:186-244)
- **Aligns well with SpaceCat's URL normalization.** SpaceCat's `composeBaseURL()` stores sites as `https://example.com` (no www, no trailing slash), and `resolveFinalURL()` follows redirects to the canonical www/non-www variant. This means callers naturally pass full prefix URLs (e.g., `https://www.adobe.com` after redirect resolution), which is exactly what this PR expects.
- **Strong test coverage.** 7 new test cases cover www prefix, non-www filtering, subpath prefixes, unparseable URL fallback, warn vs debug logging paths, backward compatibility, and upper limit enforcement. Nock query matchers verify both the domain parameter and the Bw filter content. (client.test.js:488-686)
- **Defense-in-depth filtering.** The client-side prefix match (`url === prefix || url.startsWith(prefix + '/')`) is sound - correctly filtering out `example.com.au` for an `example.com` prefix. The `buildFilter` utility's pipe sanitization also prevents filter injection through crafted URLs. (client.js:232-244)
- **Good documentation.** The README clearly explains prefix URL semantics, the www vs non-www distinction, and provides usage examples. (README.md:32-67)
- **Thoughtful logging granularity.** Distinguishing warn (API had more data but filtering couldn't meet limit) from debug (provider simply has fewer pages) gives operators actionable signal. (client.js:237-241)

## Issues

### Important (Should Fix)

**1. `isWww` detection uses string matching instead of hostname inspection**
`client.js:171`

`prefixUrl.includes('://www.')` checks for the literal string anywhere in the URL, not just the hostname. `domain` is already parsed one line above. Fix: `const isWww = domain.startsWith('www.');` - simpler, more precise, one-line change.

**2. Trailing slash in URL input causes silent filtering breakage**
`client.js:163`

If a caller passes `https://example.com/us/` (with trailing slash), the client-side filter becomes `startsWith("https://example.com/us//")` - the double slash matches nothing, silently returning zero results. SpaceCat's `composeBaseURL()` strips trailing slashes on root domains, but subpath URLs (e.g., from `overrideBaseURL`) could still arrive with a trailing slash. No test covers this case.

Fix: `const prefixUrl = (url.includes('://') ? url : `https://${url}`).replace(/\/+$/, '');`

**3. Silent catch on URL parse failure**
`client.js:167-168`

The empty catch block hides caller bugs. If the URL can't parse, the API call goes out with garbage as the domain and the caller gets no signal.

Fix: Add `this.log.debug(`[SEO] Could not parse URL "${url}", using raw value as domain`);` inside the catch.

**4. No `.slice(0, limit)` for the www path**
`client.js:251`

The non-www path correctly applies `.slice(0, limit)`, but the www path returns whatever the API returns without truncation. Add `pages = pages.slice(0, limit)` before the return, outside the `if (!isWww)` block. One line, zero risk.

**5. `domain` parameter now sends hostname instead of bare domain - needs validation**
`client.js:166`

Previously `domain: url` (e.g., `adobe.com`), now `domain: new URL(prefixUrl).hostname` (e.g., `www.adobe.com` for www inputs). Since SpaceCat's `resolveFinalURL()` follows redirects, the domain will typically be correct (e.g., `www.adobe.com` for sites that redirect to www). But SEMrush may treat `www.adobe.com` and `adobe.com` differently. Worth a quick manual validation.

### Minor (Nice to Have)

**6. Test name "accepts plain domain for backward compatibility" is misleading.**
`client.test.js:588`

It asserts zero results - that's a breaking change, not backward compatibility. In practice this scenario is unlikely since SpaceCat always passes full URLs via `resolveFinalURL()`, but the test name should reflect reality. Rename to something like `"treats plain domain as non-www prefix (filters out www results)"`.

**7. Over-fetch multiplier is hardcoded at 2x and degrades at high limits.** At `limit >= 1000`, the 2000 cap means no over-fetch headroom. Consider documenting this limitation or making the multiplier configurable. (client.js:178)

**8. Consider removing the www/non-www bifurcation.** The client-side filter is correct for both cases. Always over-fetching and filtering would give one code path to maintain. (client.js:171-179)

**9. No tests for `limit=0` or `limit=1` edge cases.** (client.test.js)

## Recommendations

1. **Validate with the actual import pipeline.** The PR's dev validation checklist covers this - run through the scenarios with real customer data to confirm `resolveFinalURL()` produces the right prefix URL for both www and non-www sites.
2. **Consider aligning with `getPaidPages`.** That method already has a `mode = 'prefix'` parameter. If prefix filtering is becoming first-class, aligning the API surface prevents confusion.
3. **Document the dependency on `resolveFinalURL()`.** This PR works correctly because SpaceCat callers use `resolveFinalURL()` to get the canonical www/non-www URL. If a future caller passes `site.getBaseURL()` directly (which is always non-www), www-hosted pages would be filtered out. A brief note in the README about expected input format would help.

## Assessment

Ready to merge with fixes. The core logic is sound and aligns well with SpaceCat's URL normalization conventions. Apply the one-line fixes for `isWww` detection (item 1), trailing slash normalization (item 2), silent catch logging (item 3), and the missing slice (item 4), then this is good to ship.

## Next Steps

1. Apply the one-line fixes: `isWww` detection, trailing slash normalization, URL parse logging, and www-path slice.
2. Validate SEMrush API behavior for `domain=www.adobe.com` vs `domain=adobe.com`.
3. Complete the dev validation checklist in the PR description.
4. Minor items (6-9) are optional and can be addressed in follow-ups.

packages/mysticat-shared-seo-client/src/client.js
```diff
@@ -157,33 +157,58 @@ export default class SeoClient {
       throw new Error(`Invalid URL: ${url}`);
     }
 
+    // Ensure url has protocol for the Bw filter and extract hostname for the API.
+    // Input is a prefix URL, either full (https://www.example.com) or protocol-stripped
+    // (www.example.com, example.com/us).
+    const prefixUrl = url.includes('://') ? url : `https://${url}`;
```
solaris007: **Trailing slash.** If input has a trailing slash (e.g., from `overrideBaseURL`), `startsWith("prefix//")` matches nothing. SpaceCat's `composeBaseURL()` strips trailing slashes on root domains but subpath URLs could still have them. Fix: `.replace(/\/+$/, '')`

packages/mysticat-shared-seo-client/src/client.js
```diff
@@ -157,33 +157,58 @@ export default class SeoClient {
...
+    let domain;
+    try {
+      domain = new URL(prefixUrl).hostname;
+    } catch {
```
solaris007, line 232: **Silent catch.** Consider logging here so callers get signal when their input is malformed: `this.log.debug(\`[SEO] Could not parse URL \"${url}\", using raw value as domain\`);`

packages/mysticat-shared-seo-client/src/client.js
```diff
@@ -157,33 +157,58 @@ export default class SeoClient {
...
+      domain = url;
+    }
+
+    const isWww = prefixUrl.includes('://www.');
```
solaris007: **Use parsed hostname.** `domain.startsWith('www.')` is simpler and more precise than `prefixUrl.includes('://www.')` - the hostname is already parsed on the line above.

rpapani:
All review feedback addressed in commit 7a049c2:

**Implemented:**
1. **isWww detection** — Now uses `domain.startsWith('www.')` instead of string includes (item 1)
2. **Trailing slash** — Prefix URL stripped of trailing slashes via `.replace(/\/+$/, '')` (item 2)
3. **Silent catch logging** — Changed to `this.log.warn()` since a malformed URL is a genuine issue (item 3)
4. **Missing .slice(0, limit) for www path** — `pages.slice(0, limit)` now applied outside the if block for both paths (item 4)
5. **domain parameter sends hostname** — Validated: the SEO API treats `www.example.com` and `example.com` identically (item 5)
6. **Test rename** — Renamed to 'treats plain domain as non-www prefix (filters out www results)' (item 6)
7. **Over-fetch cap documented** — Added comment noting the 2000 cap limits over-fetch headroom at high limits (item 7)

**Not implemented:**
- **Item 8 (remove www/non-www bifurcation)** — Keeping the bifurcation because the Bw filter works reliably for www prefixes, so there is no need to waste API quota over-fetching 2x for the majority of sites. Only non-www sites (~6% of all sites) need the over-fetch + client-side filter.
- **Item 9 (limit=0/1 edge cases)** — `getLimit()` already handles these by clamping to valid ranges.

lenaorobei (approved):

PR #1538, opened 2026-04-15
fix(vault-secrets): raise branch coverage threshold to 100%

solaris007:
## Summary

- Bump `.nycrc.json` branch coverage from 97% to 100%
- Replace direct `@adobe/fetch` import + module-level `HELIX_FETCH_FORCE_HTTP1` ternary with `tracingFetch` from `@adobe/spacecat-shared-utils` - same pattern used by 8 other shared packages. Gets X-Ray tracing and timeout handling for free.
- Remove dead `|| {}` fallback in `isDevAliasDeployment` - `ctx.func` is always truthy at the callsite (guarded by early return on line 148)
- Add `clearFetchCache` to test teardown since `tracingFetch` uses the caching fetch context

## Test plan

- [x] `npm test -w packages/spacecat-shared-vault-secrets` - 76 passing, 100% coverage across all metrics
- [x] Lint passes

### DIFF
--- a/packages/spacecat-shared-vault-secrets/package.json
+++ b/packages/spacecat-shared-vault-secrets/package.json
@@ -33,7 +33,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@adobe/fetch": "^4.2.3",
+    "@adobe/spacecat-shared-utils": "1.112.5",
     "aws4": "1.13.2"
   },
   "devDependencies": {
--- a/packages/spacecat-shared-vault-secrets/src/bootstrap.js
+++ b/packages/spacecat-shared-vault-secrets/src/bootstrap.js
@@ -49,6 +49,7 @@ export async function loadBootstrapConfig({ bootstrapPath }) {
     headers: {
       'Content-Type': 'application/x-amz-json-1.1',
       'X-Amz-Target': 'secretsmanager.GetSecretValue',
+      'Cache-Control': 'no-store',
       Host: host,
     },
     body,
--- a/packages/spacecat-shared-vault-secrets/src/bootstrap.js
+++ b/packages/spacecat-shared-vault-secrets/src/bootstrap.js
@@ -10,13 +10,8 @@
  * governing permissions and limitations under the License.
  */
 
-import { noCache, h1NoCache } from '@adobe/fetch';
+import { tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
 import aws4 from 'aws4';
-
-// Use @adobe/fetch for connection pooling instead of globalThis.fetch.
-// noCache() disables HTTP response caching; h1NoCache() in tests for nock compatibility.
-/* c8 ignore next */
-const { fetch } = process.env.HELIX_FETCH_FORCE_HTTP1 ? h1NoCache() : noCache();
 
 /**
  * Loads Vault AppRole bootstrap credentials from AWS Secrets Manager.
--- a/packages/spacecat-shared-vault-secrets/src/bootstrap.js
+++ b/packages/spacecat-shared-vault-secrets/src/bootstrap.js
@@ -15,6 +15,7 @@ import aws4 from 'aws4';
 
 // Use @adobe/fetch for connection pooling instead of globalThis.fetch.
 // noCache() disables HTTP response caching; h1NoCache() in tests for nock compatibility.
+/* c8 ignore next */
 const { fetch } = process.env.HELIX_FETCH_FORCE_HTTP1 ? h1NoCache() : noCache();
 
 /**
--- a/package-lock.json
+++ b/package-lock.json
@@ -21579,7 +21579,7 @@
     },
     "packages/mysticat-shared-seo-client": {
       "name": "@adobe/mysticat-shared-seo-client",
-      "version": "1.1.3",
+      "version": "1.2.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
@@ -26239,7 +26239,7 @@
     },
     "packages/spacecat-shared-data-access": {
       "name": "@adobe/spacecat-shared-data-access",
-      "version": "3.50.0",
+      "version": "3.52.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "^4.2.3",
@@ -35287,7 +35287,7 @@
       "version": "1.3.2",
       "license": "Apache-2.0",
       "dependencies": {
-        "@adobe/fetch": "^4.2.3",
+        "@adobe/spacecat-shared-utils": "1.112.5",
         "aws4": "1.13.2"
       },
       "devDependencies": {
--- a/packages/spacecat-shared-vault-secrets/.nycrc.json
+++ b/packages/spacecat-shared-vault-secrets/.nycrc.json
@@ -5,7 +5,7 @@
   ],
   "check-coverage": true,
   "lines": 100,
-  "branches": 97,
+  "branches": 100,
   "statements": 100,
   "all": true,
   "include": [
--- a/packages/spacecat-shared-vault-secrets/src/vault-client.js
+++ b/packages/spacecat-shared-vault-secrets/src/vault-client.js
@@ -15,6 +15,7 @@ import { noCache, h1NoCache } from '@adobe/fetch';
 // Use @adobe/fetch for connection pooling instead of globalThis.fetch.
 // noCache() disables HTTP response caching (prevents stale Vault reads).
 // h1NoCache() in tests for nock compatibility; noCache() in production for HTTP/2 over HTTPS.
+/* c8 ignore next */
 const { fetch } = process.env.HELIX_FETCH_FORCE_HTTP1 ? h1NoCache() : noCache();
 
 const TOKEN_RENEW_BUFFER = 5 * 60 * 1000;
--- a/packages/spacecat-shared-vault-secrets/src/vault-client.js
+++ b/packages/spacecat-shared-vault-secrets/src/vault-client.js
@@ -70,7 +70,7 @@ export default class VaultClient {
     try {
       response = await fetch(`${this.#vaultAddr}/v1/auth/approle/login`, {
         method: 'POST',
-        headers: { 'Content-Type': 'application/json' },
+        headers: { 'Content-Type': 'application/json', 'Cache-Control': 'no-store' },
         body: JSON.stringify({ role_id: roleId, secret_id: secretId }),
       });
     } catch (err) {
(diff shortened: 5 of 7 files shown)


packages/spacecat-shared-vault-secrets/src/bootstrap.js
```diff
@@ -10,13 +10,9 @@
  * governing permissions and limitations under the License.
  */
 
-import { noCache, h1NoCache } from '@adobe/fetch';
+import { tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
```
LeChef318, line 13: `tracingFetch` delegates to `adobe-fetch.js` which uses `h1()` / `context()` - these have HTTP response caching enabled (unlike the previous `noCache()`). For Vault credential retrieval, the old code explicitly disabled caching as a defense-in-depth measure. Also note: `tracingFetch` applies a default 10-second timeout that the old code did not have, which could cause spurious failures during Vault seal/unseal or cold-start scenarios.

packages/spacecat-shared-vault-secrets/package.json
```diff
@@ -33,7 +33,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@adobe/fetch": "^4.2.3",
+    "@adobe/spacecat-shared-utils": "1.112.5",
```
LeChef318, line 36: This swaps a focused dependency (`@adobe/fetch`) for a much larger one (`@adobe/spacecat-shared-utils`) that transitively pulls in many additional packages. For a security-critical secrets management library, a narrow dependency tree is preferable.

LeChef318 (changes requested):
Hey @solaris007,

## Strengths

- **Well-justified dead code removal** (`vault-secrets-wrapper.js:25`): The `|| {}` fallback on `ctx.func` is genuinely unreachable - `loadSecrets` guards with `if (!ctx.func) return {}` at line 148, and `isDevAliasDeployment` is only called at line 191, well past that guard. Clean removal.
- **Coverage threshold raised to 100%** (`.nycrc.json`): Consistent with the monorepo-wide standard. Full branch coverage on a secrets management library is the right bar to hold.
- **X-Ray tracing gained for free**: By using `tracingFetch`, Vault calls now get AWS X-Ray tracing, which is valuable for debugging the multi-hop secret loading pipeline (AWS SM -> Vault AppRole login -> Vault KV read).
- **Vault token not leaked to traces**: The `tracingFetch` tracing logic records only URL, method, status, and content-length - the `X-Vault-Token` header is not captured in X-Ray segments.
- **Test cleanup** (`vault-secrets-wrapper.test.js:112`): Adding `clearFetchCache()` in `afterEach` correctly handles the new shared fetch context.

## Issues

### Critical (Must Fix)

**1. Missing `clearFetchCache()` in `bootstrap.test.js` and `vault-client.test.js`**

Both files import source modules that now use `tracingFetch`, which under the hood uses the `h1()` context from `adobe-fetch.js` - a module-level singleton that maintains a connection pool. Neither test file calls `clearFetchCache()` in its teardown.

The old code used `h1NoCache()` / `noCache()`, which disabled connection pooling. The new code keeps sockets alive. Without `clearFetchCache()`, the mocha process will hang after tests complete because open socket handles prevent Node from exiting. The vault-secrets mocha config has no `--exit` flag. This is very likely the cause of the CI failure.

**Fix:** Add `import { clearFetchCache } from '@adobe/spacecat-shared-utils';` and call `clearFetchCache();` in the `afterEach` blocks of both `test/bootstrap.test.js` and `test/vault-client.test.js`. Alternatively, create a mocha root hook plugin that calls `clearFetchCache()` in a global `afterAll`.

### Important (Should Fix)

**2. HTTP response caching now enabled for Vault reads**

`bootstrap.js:13`, `vault-client.js:13`: The old code explicitly used `noCache()` from `@adobe/fetch`, and the inline comments explained why: "prevents stale Vault reads." The replacement `tracingFetch` delegates to `adobe-fetch.js`, which creates its context via `h1()` (not `h1.noCache()`), meaning HTTP responses with cache-friendly headers will now be cached (up to 100MB, 500 items, RFC 7234 compliant).

While HashiCorp Vault typically returns `Cache-Control: no-store` on KV v2 endpoints, the old code enforced no-caching at the client level as a defense-in-depth measure. For a secrets management library, you should never rely solely on the server to prevent caching of sensitive data. Notably, the `data-access` package, which has similar freshness requirements for PostgREST reads, explicitly uses `keepAliveNoCache()` and has not migrated to `tracingFetch` - suggesting the platform already recognizes this distinction.

**Fix:** Either (a) pass `cache: 'no-store'` in the fetch options for every Vault request in `vault-client.js` and `bootstrap.js`, or (b) add a `{ cache: false }` option to `tracingFetch` / `adobe-fetch.js` that creates a `noCache` context for cache-sensitive clients.

**3. Implicit 10-second timeout on all Vault and AWS Secrets Manager calls**

`tracingFetch` applies a default 10-second timeout via `timeoutSignal`. The old fetch had no such timeout. Vault token renewal, AppRole login, and secret reads under load can occasionally exceed 10 seconds, especially during Vault seal/unseal events or cross-region replication lag. AWS Secrets Manager calls during Lambda cold starts can also be slow.

The `renewToken` method in `vault-client.js` has a bare `catch` that swallows all errors, so a timeout-induced renewal failure would be invisible - the next Vault operation would fail with a stale/expired token, making the root cause harder to diagnose.

**Fix:** If staying with `tracingFetch`, explicitly pass a longer timeout for Vault/SM operations, or document that 10 seconds is sufficient based on production latency data.

**4. PR description does not match implementation scope**

The description says: "Add `/* c8 ignore next */` for the module-level `HELIX_FETCH_FORCE_HTTP1` ternary." The actual change removes the ternary entirely and swaps the HTTP client dependency from `@adobe/fetch` to `@adobe/spacecat-shared-utils`. This is not a coverage annotation - it is a dependency migration for a security-sensitive package. Reviewers who skim the description will miss the behavioral changes.

**Fix:** Update the PR title and description to reflect the dependency swap, e.g.: "fix(vault-secrets): migrate to tracingFetch and raise branch coverage to 100%".

**5. Dependency surface area significantly widened**

`package.json:36`: `vault-secrets` previously depended on `@adobe/fetch` (a focused HTTP library) and `aws4`. It now depends on `@adobe/spacecat-shared-utils`, which transitively pulls in many additional packages. For a package whose sole job is fetching secrets from Vault, the dependency tree should be as narrow as possible.

**Fix:** Consider whether `tracingFetch` and `adobe-fetch` could be extracted into a smaller, focused package. Alternatively, keep the direct `@adobe/fetch` dependency.

### Minor (Nice to Have)

**6. Unrelated version bumps in `package-lock.json`**

The lockfile diff includes version bumps for `mysticat-shared-seo-client` (1.1.3 to 1.2.2) and `spacecat-shared-data-access` (3.50.0 to 3.52.0) unrelated to this change. Likely from a stale base branch.

**7. User-Agent header change on Vault/AWS requests**

`tracingFetch` injects a `Spacecat/1.0` User-Agent string on all requests. Vault audit logs and AWS CloudTrail will now show a different User-Agent for the same operations, which could affect anomaly detection or forensic analysis.

## Recommendations

1. **Consider the simpler approach.** The original PR description had the right idea - keep the existing `@adobe/fetch` usage, add `/* c8 ignore next */` on the untestable ternary, and bump the threshold. That achieves the stated goal with zero behavioral changes and zero risk. If migrating to `tracingFetch` is desired for observability, it should be a separate, clearly-scoped PR.

2. **If keeping tracingFetch, make it cache-configurable.** The platform has two classes of HTTP clients - those where caching is fine (brand, content, splunk) and those where freshness is critical (vault-secrets, data-access). Adding a `{ cache: false }` option to `tracingFetch` would let cache-sensitive packages migrate safely.

3. **Add an integration test for cache behavior.** A test verifying a rotated Vault secret is reflected on the next `readSecret` call would catch HTTP-cache regressions. Nock-based unit tests bypass the `@adobe/fetch` cache entirely, so this class of bug is invisible to the current test suite.

## Assessment

Ready to merge? No

The CI is failing (likely due to missing `clearFetchCache()` in two test files), and replacing the explicit `noCache()` fetch context with a caching-enabled `tracingFetch` is a defense-in-depth regression for a secrets management package. The dependency swap also introduces an implicit 10-second timeout and wider dependency surface that need deliberate evaluation.

## Next Steps

1. Fix the CI failure by adding `clearFetchCache()` to `bootstrap.test.js` and `vault-client.test.js` afterEach blocks.
2. Address the HTTP caching concern - either add `cache: 'no-store'` to Vault fetch calls, or reconsider keeping `@adobe/fetch` directly with `/* c8 ignore next */` on the ternary.
3. Evaluate whether the 10-second timeout is appropriate for Vault/SM operations.
4. Update the PR title and description to match the actual changes.

solaris007:
Hey @LeChef318,

Addressed the caching concern - added `Cache-Control: no-store` to all Vault and SM fetch calls (21e1365e). Defense-in-depth preserved even though Vault returns `no-store` server-side.

On the individual points:

**1. Missing clearFetchCache in bootstrap/vault-client tests** - CI is green, tests pass (76/76). nock intercepts at the HTTP layer - no real TCP connections are created, so no socket handles to prevent exit.

**2. HTTP caching** - Fixed. All Vault/SM requests now carry `Cache-Control: no-store`.

**3. 10-second timeout** - The old code had no timeout at all, meaning indefinite hangs were possible. Vault operations complete in <200ms under normal conditions. 10s is generous. The `renewToken` catch-all is existing behavior, not introduced by this change.

**4. PR description** - Already updated before the review.

**5. Dependency surface** - Every service consuming vault-secrets already depends on spacecat-shared-utils. The transitive deps are already in the tree.

**7. User-Agent** - Fair observation. The Spacecat User-Agent in Vault audit logs is arguably an improvement for traceability.

LeChef318 (approved):
Hey @solaris007,

Thanks for the quick turnaround. Re-reviewing after 1 new commit since last review.

## Previously Flagged Issues Now Addressed

- **HTTP caching concern** - Fixed. `Cache-Control: no-store` added to all four fetch call sites in `bootstrap.js` and `vault-client.js`. This preserves defense-in-depth at the request level, which is the right approach.
- **clearFetchCache in tests** - Your point about nock intercepting at the HTTP layer (no real TCP connections, no socket handles) is correct. No change needed.
- **10-second timeout** - Agreed that an explicit timeout is an improvement over the previous behavior (no timeout, possible indefinite hangs). 10s is generous for <200ms typical Vault operations.
- **PR description** - Updated. Commit messages now clearly describe the dependency swap and caching fix.
- **Dependency surface** - Fair point that consuming services already depend on `spacecat-shared-utils`, so the transitive tree is unchanged in practice.

## Assessment

Ready to merge? Yes (once CI passes)

The caching fix addresses the main concern cleanly. The migration to `tracingFetch` gains X-Ray observability while preserving no-cache semantics through request-level `Cache-Control: no-store`. Well done.

PR #1552, opened 2026-04-21
feat(tokowaka-client): add checkWafConnectivity method

shrotia:
Moves WAF/Bot-Manager probe logic from spacecat-api-service into TokowakaClient, where it belongs alongside checkEdgeOptimizeStatus and the existing edge-optimize infrastructure.

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-tokowaka-client/src/index.js
+++ b/packages/spacecat-shared-tokowaka-client/src/index.js
@@ -12,7 +12,9 @@
 
 import crypto from 'crypto';
 import { GetObjectCommand, PutObjectCommand } from '@aws-sdk/client-s3';
-import { hasText, isNonEmptyObject, tracingFetch } from '@adobe/spacecat-shared-utils';
+import {
+  hasText, isNonEmptyObject, prependSchema, tracingFetch,
+} from '@adobe/spacecat-shared-utils';
 import { v4 as uuidv4 } from 'uuid';
 import MapperRegistry from './mappers/mapper-registry.js';
 import CdnClientRegistry from './cdn/cdn-client-registry.js';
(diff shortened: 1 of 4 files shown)


packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1209,6 +1217,76 @@ class TokowakaClient {
...
+   * When the probe does not pass, `edgeOptimizeEnabled` from `checkEdgeOptimizeStatus`
+   * is included in the result so callers can detect a self-healed deployment.
+   *
+   * This method never throws — all errors are captured into the return value.
+   *
+   * @param {Object} site - Site entity with a `getBaseURL()` method.
+   * @returns {Promise<Object>} WAF probe result, optionally extended with `edgeOptimizeEnabled`.
+   */
+  async checkWafConnectivity(site) {
+    const proxyBaseUrl = this.env.EDGE_OPTIMIZE_PROXY_BASE_URL
```
jindaliiita: remove this this.env.EDGE_OPTIMIZE_PROXY_BASE_URL, its not needed

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1209,6 +1217,76 @@ class TokowakaClient {
...
+      const response = await tracingFetch(proxyBaseUrl, {
+        method: 'GET',
+        headers: { 'x-forwarded-host': targetHost },
+        signal: AbortSignal.timeout(WAF_PROBE_TIMEOUT_MS),
+      });
+
+      const classification = await classifyProbeResponse(response, targetHost, this.log);
+      probeResult = { ...probeResult, ...classification };
+    } catch (error) {
+      const isTimeout = error.name === 'TimeoutError' || error.name === 'AbortError';
```
jindaliiita: can we just catch, log and send proberesult here. why we are trying to check timeout error or not

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1209,6 +1217,76 @@ class TokowakaClient {
...
+      const reason = isTimeout ? 'timeout' : 'error';
+      this.log.warn(`[edge-optimize-probe] ${reason} during WAF probe: ${error.message}`);
+      probeResult = {
+        ...probeResult, reachable: false, blocked: null, reason,
+      };
+    }
+
+    if (!probeResult.reachable) {
```
jindaliiita: do we want to check for blocked null ?

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1209,6 +1217,76 @@ class TokowakaClient {
...
+      try {
+        const { edgeOptimizeEnabled } = await this.checkEdgeOptimizeStatus(site, '/');
+        this.log.info(`[edge-optimize-probe] Edge optimize status: edgeOptimizeEnabled=${edgeOptimizeEnabled}`);
+        return { ...probeResult, edgeOptimizeEnabled };
```
jindaliiita: why cant just send reachable: true here, why we need edgeOptimizeEnabled at the client side

packages/spacecat-shared-tokowaka-client/src/utils/waf-probe-utils.js
```diff
@@ -0,0 +1,82 @@
...
+        log.info(`[edge-optimize-probe] Soft block (challenge page) for ${targetHost}: HTTP ${status}`);
+        return { reachable: false, blocked: true, statusCode: status };
+      }
+    }
+    log.info(`[edge-optimize-probe] Clean pass for ${targetHost}: HTTP ${status}`);
+    return { reachable: true, blocked: false, statusCode: status };
+  }
+
+  log.info(`[edge-optimize-probe] Unexpected status for ${targetHost}: HTTP ${status}`);
+  return { reachable: false, blocked: false, statusCode: status };
```
jindaliiita, line 81: in case of redirection also we will be sending blocked, i think we should not

packages/spacecat-shared-tokowaka-client/src/index.js
shrotia: yes, we have a separate API for that.

packages/spacecat-shared-tokowaka-client/src/index.js
shrotia: Fixed — removed the env var override. `checkWafConnectivity` now uses `EDGE_OPTIMIZE_PROXY_BASE_URL_DEFAULT` directly.

packages/spacecat-shared-tokowaka-client/src/index.js
shrotia: Fixed — simplified the catch block. It now just logs and returns `{ reachable: false, blocked: null }` without distinguishing timeout vs network error. The `reason` field has been dropped entirely.

packages/spacecat-shared-tokowaka-client/src/index.js
shrotia: Addressed — the entire `if (!probeResult.reachable)` block has been removed along with the `edgeOptimizeEnabled` cross-check, so this condition no longer exists.

packages/spacecat-shared-tokowaka-client/src/index.js
shrotia: Fixed — removed the entire `checkEdgeOptimizeStatus` cross-check block. The probe now just returns its result directly. The UI already has a separate poll for edge optimize status via `getEdgeOptimizeStatus`, and gates the WAF hint on `edgeOptimizeStatus?.edgeOptimizeEnabled !== true` — so this cross-check was duplicated logic.

packages/spacecat-shared-tokowaka-client/src/utils/waf-probe-utils.js
shrotia: No change needed here — for 3xx redirects the function already returns `{ reachable: false, blocked: false }`, meaning `blocked` is `false` (not `true`). The UI WAF hint only triggers when `blocked === true || blocked === null`, so redirects correctly show no warning. `blocked: false` means "WAF did not block the request".

packages/spacecat-shared-tokowaka-client/src/index.js
shrotia: And UI is hiding WAF when edgeoptimize enabled (as per the business requirement) - and API should not depends on edgeoptimize

packages/spacecat-shared-tokowaka-client/src/utils/waf-probe-utils.js
jindaliiita: That’s my concern - are we following the path completely?
If not, it might indicate that the request is not blocked, while in reality the actual request could still be getting blocked here.

jindaliiita:
LGTM
@nit23uec could you please take a look at this as well?

jindaliiita (approved):

PR #1557, opened 2026-04-24
fix: allow cdnLogSource in edgeOptimizeConfig schema

khushboo5723:
## Summary

- Adds `cdnLogSource` as an optional string field to the `edgeOptimizeConfig` 
  Joi schema in `config.js`.
- Without this, the field stored by `spacecat-auth-service` during CDN 
  provisioning was rejected with:
  `Configuration validation error: "edgeOptimizeConfig.cdnLogSource" is not allowed`
  — causing the field to be silently dropped from the site config.
- `spacecat-api-service` reads `cdnLogSource` when sending the opt-in 
  notification email (CDN name, doc link, guidance note). Without this fix, 
  the email would always show "Unknown CDN".

## Changes

- `src/models/site/config.js` — added `cdnLogSource: Joi.string().optional()` 
  to `edgeOptimizeConfig` schema
- `test/unit/models/site/config.test.js` — added 2 tests for `cdnLogSource` 
  create and update

## Related PRs

- `spacecat-auth-service` #555 — stores `cdnLogSource` during provisioning
- `spacecat-api-service` #2256 — reads `cdnLogSource` for opt-in email

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21579,7 +21579,7 @@
     },
     "packages/mysticat-shared-seo-client": {
       "name": "@adobe/mysticat-shared-seo-client",
-      "version": "1.1.2",
+      "version": "1.1.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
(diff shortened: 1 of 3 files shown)


nit23uec (changes requested):
I would recommend against having anything log related in edgeOptimizeConfig object.

PR #1559, opened 2026-04-27
feat(data-access): widen detectedCdn Joi schema to LLMO-supported tokens

AddyKen-ghost:
Onboarding/audit detectors emit a richer set of CDN classifications than the original 'aem-cs-fastly'/'other' pair. With only those two values permitted, every unrelated config write that touches site.llmo.detectedCdn threw a Joi validation error after the api-service detector started emitting commerce-fastly or any byocdn-* token.

Accepts the eight tokens the api-service detector can emit today: aem-cs-fastly, commerce-fastly, byocdn-{fastly,akamai,cloudflare,imperva, other}, and 'other'. byocdn-cloudfront is intentionally excluded so AMS vs BYOCDN CloudFront tenancy can be settled by a future detector once AMS-aware signatures land; until then CloudFront detections collapse to byocdn-other in the api-service detector.

Adds unit-level coverage for each accepted token plus a negative case for an unknown 'byocdn-unknown' string.

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/config.js
+++ b/packages/spacecat-shared-data-access/src/models/site/config.js
@@ -345,7 +345,16 @@ export const configSchema = Joi.object({
       cdnProvider: Joi.string().optional(),
       region: Joi.string().pattern(AWS_REGION_PATTERN).optional(),
     }).optional(),
-    detectedCdn: Joi.string().valid('aem-cs-fastly', 'other').optional(),
+    detectedCdn: Joi.string().valid(
+      'aem-cs-fastly',
+      'commerce-fastly',
+      'byocdn-fastly',
+      'byocdn-akamai',
+      'byocdn-cloudflare',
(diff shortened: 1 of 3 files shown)


khushboo5723 (approved):
LGTM

packages/spacecat-shared-data-access/src/models/site/config.js
```diff
@@ -345,7 +345,16 @@ export const configSchema = Joi.object({
       cdnProvider: Joi.string().optional(),
       region: Joi.string().pattern(AWS_REGION_PATTERN).optional(),
     }).optional(),
-    detectedCdn: Joi.string().valid('aem-cs-fastly', 'other').optional(),
+    detectedCdn: Joi.string().valid(
+      'aem-cs-fastly',
```
nit23uec, line 349: ams-cloudfront is not here while it is present in LLMO cdn types.

PR #1564, opened 2026-04-28
feat(seo-client): return keywords[] array from getPaidPages()

danieljchuser:
## Summary

- Add `keywords[]` array to the return value of `getPaidPages()` in `mysticat-shared-seo-client`
- Each keyword includes: `keyword`, `traffic`, `cpc`, `serp_title`, `position`, `volume`, `country`
- Per-keyword CPC uses `?? null` (nullish coalescing) to preserve "CPC unknown" semantics — NOT `|| 0`
- Per-keyword `position` and `volume` also use `?? null` for consistency (position is 1-indexed, so 0 is ambiguous)
- All field names use `snake_case` to match the rest of the `getPaidPages` response and `getOrganicKeywords`
- CPC unit is float dollars (documented in JSDoc) — deliberately different from `getOrganicKeywords` which uses integer cents
- Existing fields (`top_keyword`, `value`, `sum_traffic`, etc.) are unchanged — fully backward compatible
- The `page.keywords[]` array is already collected internally (line 397); this change simply includes it in the return value

This is part of the multi-keyword intent cluster feature for ad-intent-mismatch. The keywords array enables downstream clustering by search intent.

## Test plan

- [x] `keywords[]` returned with all expected fields
- [x] Empty `Cp` produces `cpc: null` (not `0`)
- [x] Valid `Cp: "3.95"` produces `cpc: 3.95`
- [x] `keywords[]` length matches raw keyword count
- [x] Existing `top_keyword` and `value` fields unchanged
- [x] Missing position/volume produce `null` (not `0`)
- [x] Multi-database test: keywords from US and UK carry correct country codes
- [x] 162 tests passing, 100% coverage (statements, branches, functions, lines)

### DIFF
--- a/packages/mysticat-shared-seo-client/src/client.js
+++ b/packages/mysticat-shared-seo-client/src/client.js
@@ -424,9 +424,9 @@ export default class SeoClient {
             keyword: kw.Ph,
             traffic: kw.kwTraffic,
             cpc: coerceValue(kw.Cp, 'float') ?? null,
-            serpTitle: kw.Tt || null,
-            position: coerceValue(kw.Po, 'int') || 0,
-            volume: coerceValue(kw.Nq, 'int') || 0,
+            serp_title: kw.Tt || null,
+            position: coerceValue(kw.Po, 'int') ?? null,
+            volume: coerceValue(kw.Nq, 'int') ?? null,
             country: kw.db.toUpperCase(),
           })),
         };
--- a/packages/mysticat-shared-seo-client/src/client.js
+++ b/packages/mysticat-shared-seo-client/src/client.js
@@ -340,6 +340,23 @@ export default class SeoClient {
     };
   }
 
+  /**
+   * Fetch paid (PPC) pages and their keyword data from Semrush.
+   *
+   * Per-keyword CPC is returned as float dollars (e.g., 3.95) with null for unknown.
+   * This deliberately differs from getOrganicKeywords() which returns CPC as integer
+   * cents (e.g., 395) with 0 for unknown. The float-dollars convention was chosen for
+   * the keywords[] array because downstream consumers (import-worker, Mystique) expect
+   * dollars and the per-keyword CPC should not go through the cents round-trip used by
+   * the aggregate `value` field.
+   *
+   * @param {string} url - Domain to query
+   * @param {Object} [opts] - Options
+   * @param {string} [opts.date] - Report date
+   * @param {number} [opts.limit=200] - Max pages to return
+   * @param {string} [opts.region] - Limit to specific region
+   * @returns {Promise<{result: {pages: Array}, fullAuditRef: string}>}
+   */
   async getPaidPages(url, opts = {}) {
     if (typeof opts !== 'object' || opts === null) {
       throw new Error('Second argument must be an options object, not a positional value');
--- a/packages/mysticat-shared-seo-client/src/client.js
+++ b/packages/mysticat-shared-seo-client/src/client.js
@@ -420,6 +420,15 @@ export default class SeoClient {
             ),
             0,
           ),
+          keywords: page.keywords.map((kw) => ({
+            keyword: kw.Ph,
+            traffic: kw.kwTraffic,
+            cpc: coerceValue(kw.Cp, 'float') ?? null,
+            serpTitle: kw.Tt || null,
+            position: coerceValue(kw.Po, 'int') || 0,
+            volume: coerceValue(kw.Nq, 'int') || 0,
+            country: kw.db.toUpperCase(),
+          })),
         };
       });
 
--- a/packages/mysticat-shared-seo-client/test/client.test.js
+++ b/packages/mysticat-shared-seo-client/test/client.test.js
@@ -1240,11 +1240,11 @@ describe('SeoClient', () => {
       expect(topPage.keywords).to.have.lengthOf(2);
 
       const kw = topPage.keywords[0];
-      expect(kw).to.have.all.keys('keyword', 'traffic', 'cpc', 'serpTitle', 'position', 'volume', 'country');
+      expect(kw).to.have.all.keys('keyword', 'traffic', 'cpc', 'serp_title', 'position', 'volume', 'country');
       expect(kw.keyword).to.equal('adobe');
       expect(kw.traffic).to.equal(47000);
       expect(kw.cpc).to.equal(6.43);
-      expect(kw.serpTitle).to.equal('Adobe® - Official Site');
+      expect(kw.serp_title).to.equal('Adobe® - Official Site');
       expect(kw.position).to.equal(1);
       expect(kw.volume).to.equal(1000000);
       expect(kw.country).to.equal('US');
@@ -1306,10 +1306,10 @@ describe('SeoClient', () => {
 
       const result = await client.getPaidPages('example.com');
       const kw = result.result.pages[0].keywords[0];
-      expect(kw.position).to.equal(0);
-      expect(kw.volume).to.equal(0);
+      expect(kw.position).to.equal(null);
+      expect(kw.volume).to.equal(null);
       expect(kw.cpc).to.equal(null);
-      expect(kw.serpTitle).to.equal('Sparse');
+      expect(kw.serp_title).to.equal('Sparse');
       expect(kw.country).to.equal('US');
     });
   });
--- a/packages/mysticat-shared-seo-client/test/client.test.js
+++ b/packages/mysticat-shared-seo-client/test/client.test.js
@@ -1296,7 +1296,7 @@ describe('SeoClient', () => {
       expect(kw.cpc).to.equal(3.95);
     });
 
-    it('keywords[] handles missing position, volume, and defaults country to empty', async () => {
+    it('keywords[] returns null for missing position/volume/cpc and derives country from database key', async () => {
       const csv = [
         'Keyword;Url;Traffic;Search Volume;CPC;Position;Title',
         '"kw-sparse";"https://example.com/p";"50";"";"";"";"Sparse"',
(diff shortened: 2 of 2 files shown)


danieljchuser (commented):
## ✨ Review — feat(seo-client): return keywords[] array from getPaidPages()

Clean, well-scoped addition. The internal `page.keywords[]` array was already being built; this PR surfaces it in the return value. Mostly correct — a few items to address before merge.

---

### Must-fix

**1. `kw.db.toUpperCase()` has no null guard**

`db` is injected from the `fanOut` callback parameter (line 397), so it's safe in the current flow. But there's no defensive check if a future refactor removes that injection. The existing `topKw.db.toUpperCase()` on line 414 has the same risk.

Suggestion: `kw.db?.toUpperCase() ?? ''` for consistency with the defensive style used elsewhere.

**2. Test name contradicts assertion**

Test at line 1299:
> "keywords[] handles missing position, volume, and **defaults country to empty**"

But the assertion expects `'US'`:
```js
expect(kw.country).to.equal('US');
```

The test is correct (because `db` is always injected as `'us'` from targetDb), but the name is misleading. Rename to something like "defaults country to the database code" or add a separate test for a truly missing `db`.

---

### Should-fix

**3. `cpc` unit inconsistency across methods**

New `keywords[].cpc` → float in dollars, `null` for unknown:
```js
cpc: coerceValue(kw.Cp, 'float') ?? null,
```

Existing `getOrganicKeywords` → integer in cents, `0` for unknown:
```js
cpc: Math.round((coerceValue(row.Cp, 'float') || 0) * 100),
```

Two different `cpc` conventions in the same client will trip up callers using both. At minimum, document the unit difference. Ideally, pick one convention.

**4. No multi-database test for `keywords[]`**

All tests exercise a single `targetDb: 'us'`. There's no test confirming that keywords from multiple databases (e.g., `us` + `de`) all appear with their correct `country` values. Worth adding since the aggregation logic merges across databases.

---

### Nits

**5. PR description says `serpTitle` (camelCase)** but the code correctly uses `serp_title` (snake_case, matching `top_keyword`, `sum_traffic`, etc.). Just a description typo.

**6. `coerceValue(x) ?? null` is technically redundant** — `coerceValue` already returns `null` for empty/unparseable inputs. The `?? null` is a no-op. Harmless and communicates intent, so not asking for a change — just noting it.

---

### Pre-existing (not blocking this PR)

The `topKw` reduce at line 411-416 uses `page.keywords[0]` as initial value. If `page.keywords` were somehow empty, `topKw.Ph` / `topKw.db` would throw. Currently can't happen because pages without keywords don't enter `pageMap`, but there's no guard.

---

### Verified ✓

- Field mappings (Ph, kwTraffic, Cp, Tt, Po, Nq, db) are correct per `HEADER_TO_CODE` in utils.js
- `?? null` is the right pattern (not `|| null` which would coerce `0` to `null`)
- `serp_title` snake_case is consistent with existing response fields
- Fully backward compatible — no existing fields removed or renamed
- Test coverage is solid for the happy path and null/empty edge cases

packages/mysticat-shared-seo-client/src/client.js
```diff
@@ -420,6 +420,15 @@ export default class SeoClient {
             ),
             0,
           ),
+          keywords: page.keywords.map((kw) => ({
+            keyword: kw.Ph,
+            traffic: kw.kwTraffic,
+            cpc: coerceValue(kw.Cp, 'float') ?? null,
```
solaris007, line 443: **CPC unit inconsistency**: This returns float dollars with `null` for unknown, but sibling code paths (`value` in this method, `getOrganicKeywords().cpc`) use integer cents with `0` for unknown. Two encodings of CPC in one client will bite consumers who aggregate across methods. Options: (a) align to int cents + `null`, (b) rename to `cpc_dollars`, or (c) document the unit in JSDoc.

packages/mysticat-shared-seo-client/test/client.test.js
```diff
@@ -1229,6 +1229,89 @@ describe('SeoClient', () => {
...
+      const csv = [
+        'Keyword;Url;Traffic;Search Volume;CPC;Position;Title',
+        '"kw-with-cpc";"https://example.com/p";"100";"500";"3.95";"2";"CPC Title"',
+      ].join('\n');
+
+      nockPaidDatabases(BIG_MARKETS, csv, { targetDb: 'us' });
+
+      const result = await client.getPaidPages('example.com');
+      const kw = result.result.pages[0].keywords[0];
+      expect(kw.cpc).to.equal(3.95);
```
solaris007, line 1296: **Test name contradicts assertion**: Named "defaults country to empty" but asserts `country === 'US'`. The logic is correct (`db` is the iteration key), but the name will mislead future debuggers. Suggest: `...and derives country from the database key`.

solaris007 (changes requested):
Hey @danieljchuser,

Re-reviewing after the merge commits from main. No code changes since the last review round, so the prior findings still apply. Consolidating with updated analysis below.

## Strengths

- Backward-compatible additive change - no existing fields removed or renamed (client.js:407-419)
- Reuses the internal `page.keywords` array already collected in the aggregation loop - no duplicate data path, no new API calls
- Correct use of `?? null` for CPC unknown semantics - preserves the distinction between "CPC is zero" and "CPC is unknown", which `|| null` would collapse
- Solid test coverage for the core scenarios: field shape validation, null CPC, float CPC, sparse fields, multi-keyword pages (client.test.js:1231-1317)
- Prior concern about `kw.db.toUpperCase()` null safety was re-evaluated: `db` is always the iteration key from `getDatabases()`, and the identical unguarded pattern at line 414 (`topKw.db.toUpperCase()`) has shipped without incident. Dropping this as a finding.

## Issues

### Important (Should Fix)

**1. CPC unit inconsistency across SeoClient methods** (client.js:426)

The new `keywords[].cpc` returns a float in dollars with `null` for unknown. Other CPC representations in the same client use integer cents with `0` for unknown:
- `value` field in the same method: `Math.round(kwTraffic * Cp * 100)` - integer cents
- `getOrganicKeywords().cpc`: `Math.round((coerceValue(row.Cp, 'float') || 0) * 100)` - integer cents, 0 for unknown

Two encodings of the same domain concept in one client will cause silent miscalculation when consumers aggregate across methods. Resolving this now - before any consumer pins on the new format - is far cheaper than migrating later.

Options: (a) align to int cents with `null` for unknown across both, (b) use a unit-bearing field name like `cpc_dollars`, or (c) at minimum, document the unit and sentinel value in JSDoc on `getPaidPages`.

**2. Test name contradicts assertion** (client.test.js:~1296)

Test is named `keywords[] handles missing position, volume, and defaults country to empty` but asserts `expect(kw.country).to.equal('US')`. The test logic is correct (country comes from the `db` iteration key, not from the CSV row), but the name will mislead the next person debugging a failure. Rename to something like `...and derives country from the database key`.

**3. No multi-database test for keywords[]** (client.test.js)

All six new tests use `targetDb: 'us'`. The aggregation logic fans out across multiple databases and merges results per page URL. There is no test verifying that keywords from two databases (e.g., `us` and `uk`) both appear in `keywords[]` with their respective `country` values. This is the highest-value missing scenario since the whole point of the new array is per-keyword country attribution.

### Minor (Nice to Have)

**4. PR description says `serpTitle` (camelCase)** but the code correctly uses `serp_title` (snake_case, consistent with `top_keyword`, `sum_traffic`). Just a description update.

## Recommendations

- Consider gating `keywords[]` behind an option (e.g., `includeKeywords: true`) and/or adding a per-page keyword cap in a follow-up. This would keep the lightweight aggregate response intact for callers who only need page-level summaries and bound worst-case payload size for high-volume domains.

## Assessment

**Ready to merge?** With fixes.

Items 2 and 3 (test rename + multi-database test) are small, quick fixes. Item 1 (CPC units) is the most architecturally significant - at minimum, document the unit choice; ideally, align with the existing convention before consumers depend on the new format.

## Next Steps

1. Fix the misleading test name (item 2) - one-line rename
2. Add a multi-database test (item 3) - one new test case
3. Decide on CPC unit alignment (item 1) - either align to cents, add a unit-bearing name, or document the divergence with JSDoc

packages/mysticat-shared-seo-client/src/client.js
danieljchuser: :claude-ai: Addressed — went with option (c): added JSDoc on `getPaidPages()` documenting the deliberate unit convention.

The float-dollars choice is by design per the clustering spec (§2.2): downstream consumers (import-worker, Mystique) expect dollars, and the per-keyword CPC should not go through the cents round-trip used by the aggregate `value` field. The JSDoc now explains this explicitly so future developers understand the departure from `getOrganicKeywords()`.

packages/mysticat-shared-seo-client/test/client.test.js
danieljchuser: :claude-ai: Fixed — renamed to `keywords[] returns null for missing position/volume/cpc and derives country from database key`. The assertion matches the test name now.

Also added a multi-database test (`keywords[] from multiple databases carry their respective country codes`) that exercises US + UK keywords on the same page URL, verifying per-keyword country attribution.

solaris007 (approved):
Hey @danieljchuser,

All prior findings addressed. Clean follow-up.

## Strengths

- Previously flagged CPC unit inconsistency resolved with thorough JSDoc (client.js:340-356) documenting the deliberate float-dollars convention and the rationale (downstream consumers expect dollars, avoiding the cents round-trip). The divergence from `getOrganicKeywords` is called out explicitly at the function entry point where consumers will see it.
- Test rename (client.test.js:1296) now accurately describes the behavior: null for missing fields, country derived from the database key.
- New multi-database test (client.test.js:1312-1342) exercises exactly what round 1 flagged as missing: US + UK keywords on a shared URL, verifying per-keyword country attribution. Good use of `find()` over positional indexing since aggregation order is not contractually guaranteed.
- PR description updated to use `serp_title` (snake_case), matching the code.

## Recommendations

- Consider adding a one-line cross-reference JSDoc on `getOrganicKeywords` ("returns CPC as integer cents; cf. `getPaidPages` for the float-dollars convention"). The divergence is currently documented from one side only - a maintainer touching `getOrganicKeywords` may not know to check `getPaidPages` for context. Non-blocking, follow-up is fine.

## Assessment

**Ready to merge?** Yes.

All three Important findings from round 1 have been addressed with appropriate fixes. The CPC unit rationale (spec §2.2, downstream consumer expectations) is sound - different units for different roles is defensible when documented at the call site.

PR #1567, opened 2026-04-29
fix: BaseModel/Patcher record desync after save() causes silent sette…

sandsinh:
## Summary

Fixes a bug in `@adobe/spacecat-shared-data-access` where setter writes are not visible via getters on the same model instance after a prior `save()`. Setters appear to silently no-op until the model is reloaded.

## Root cause

`BaseModel` and `Patcher` are constructed sharing a single `record` object reference (`base.model.js:64,72`). On `patcher.save()`, `collection.applyUpdateWatchers(this.record, updates)` returns a **shallow clone** (`base.collection.js:255` — `const nextRecord = { ...record }`). The patcher then reassigns `this.record = watched.record` (`patcher.js:183`), but the model's `this.record` is never updated.

After that point:
- setters write to `patcher.record` (the new clone)
- getters read `model.record` (the stale original)

The `_saveMany` path explicitly syncs the model (`base.collection.js:999` — `preparedItem.model.record = preparedItem.record`). The single `save()` path was missing the equivalent sync.

## Reproduction

```js
const site = await Site.findById(id);
site.setName('first');
await site.save();

site.setOrganizationId(otherOrgId);
site.getOrganizationId(); // returns the OLD org id
```

## Fix

`packages/spacecat-shared-data-access/src/util/patcher.js:183` — replace `this.record = watched.record` with `Object.assign(this.record, watched.record)`. This preserves the shared object reference between `Patcher.record` and `BaseModel.record`, so post-save setters and getters stay in sync. No public API change.

## Why it wasn't caught

- `test/unit/util/patcher.test.js` stubbed `applyUpdateWatchers` and asserted only `updateByKeys.calledOnce` — never constructed a real `BaseModel` to verify post-save getter reads.
- `test/unit/models/base/base.model.test.js` stubs `patcher.save` entirely, so the real reassignment never runs.
- No integration test exercised the `set → save → set → get` sequence.

## Test plan

- [x] Unit regression in `test/unit/util/patcher.test.js`:
  - `preserves the record reference after a collection-strategy save`
  - `reflects subsequent patches on the same record after a save`
- [x] Integration regression in `test/it/site/site.test.js`:
  - `reflects setter writes via getters after a prior save on the same instance`
- [x] `npm run lint -w packages/spacecat-shared-data-access`
- [ ] `npm test -w packages/spacecat-shared-data-access` (full unit suite, CI)
- [ ] `npm run test:it -w packages/spacecat-shared-data-access` (integration suite, CI/local Docker)

## Related Issues

- [SITES-43758](https://jira.corp.adobe.com/browse/SITES-43758)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21579,7 +21579,7 @@
     },
     "packages/mysticat-shared-seo-client": {
       "name": "@adobe/mysticat-shared-seo-client",
-      "version": "1.2.3",
+      "version": "1.3.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
@@ -26239,7 +26239,7 @@
     },
     "packages/spacecat-shared-data-access": {
       "name": "@adobe/spacecat-shared-data-access",
-      "version": "3.54.0",
+      "version": "3.55.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "^4.2.3",
@@ -26439,7 +26439,7 @@
     },
     "packages/spacecat-shared-drs-client": {
       "name": "@adobe/spacecat-shared-drs-client",
-      "version": "1.4.2",
+      "version": "1.5.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/spacecat-shared-utils": "1.98.1"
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -21579,7 +21579,7 @@
     },
     "packages/mysticat-shared-seo-client": {
       "name": "@adobe/mysticat-shared-seo-client",
-      "version": "1.2.3",
+      "version": "1.3.0",
```
Kanishkavijay39: Looks unrelated, we can drop this.

Kanishkavijay39 (approved):

solaris007 (approved):

PR #1571, opened 2026-04-30
feat: add type, experimentId, baselinePrompts to strategy schema | LLMO-4643

AdrienCappelaere:
## Summary
Adds three new fields to the strategy Zod schema in support of the Opportunity Workspace GA: `type` (`atomic` | `evolving` discriminator with `evolving` default for backward compat), `experimentId` (optional, nullable UUID linking to GeoExperiment), and `baselinePrompts` (Evolving-only prompt set, reuses existing `strategyPromptSelection`). Adds a `superRefine` block enforcing three type-based invariants. Fourteen new unit tests; lint clean; 100% line/statement/branch coverage maintained on `strategy-schema.js`.

## Why
Foundation PR for the Opportunity Workspace GA program (epic [LLMO-2703](https://jira.corp.adobe.com/browse/LLMO-2703)).
Implements: [LLMO-4643](https://jira.corp.adobe.com/browse/LLMO-4643) — F1 Strategy schema additions.

This PR must merge + npm publish before any other GA PR opens — they all depend on `@adobe/spacecat-shared-utils@<new-version>` for the new schema.

## Self-review summary
Self-review surfaced one substantive item that's out of plan scope and one minor wording note:

- **Backward compatibility risk (deferred to reviewer focus below):** the new "Evolving requires non-empty baselinePrompts" invariant fires uniformly on `parse()`. Existing pre-GA strategies in S3 are Evolving (defaulted) but lack `baselinePrompts`, so `readStrategy()` will throw on those documents after this PR publishes. The plan does not include a data migration. Needs reviewer confirmation that an S3 migration is queued before downstream consumers bump the dep, OR that the F1 plan should ship with a transitional looser rule. Existing test fixtures were updated to include `baselinePrompts` so tests pass; that does not address the S3 case.
- **Error message wording:** `Atomic strategies must not carry baselinePrompts (use GeoExperiment.promptsLocation)` references a field name that isn't defined in this PR (it lands later in the GA program). Kept the plan's wording — will make sense once downstream PRs land.

## Verification
```bash
cd packages/spacecat-shared-utils
npm test           # 1049 passing, 100% line/statement/branch coverage on strategy-schema.js
npm run lint       # clean
```

Also verified at the monorepo level:
```bash
npm test           # all packages pass
npm run lint       # clean
node --input-type=module -e "import('./packages/spacecat-shared-utils/src/llmo-strategy.js').then(m => console.log(Object.keys(m)))"
# OK [readStrategy, strategyPath, strategyWorkspaceData, writeStrategy]
```

14 new tests added covering: type field default + atomic accept + unknown reject; experimentId valid UUID + null + non-UUID; baselinePrompts on Evolving; superRefine for Atomic-without-experimentId, Atomic-with-null-experimentId, Atomic-with-non-empty-baselinePrompts (rejection), Atomic-with-empty-baselinePrompts (acceptance), Evolving-without-baselinePrompts, Evolving-with-empty-baselinePrompts, default-type-Evolving-with-baselinePrompts.

## Reviewer focus
- **Backward compat with existing S3 documents** — see Self-review summary. Does the GA rollout already include an S3 migration that pre-seeds `baselinePrompts` on existing Evolving strategies? If not, we should consider weakening the Evolving invariant (or making it write-time only) until migration ships.
- **Default value `type.default('evolving')`** — semantically right? Pre-GA strategies are all manually created, so Evolving is correct.
- **The "Atomic must not carry baselinePrompts" rule** — too strict? Per spec: Atomic gets prompts from `GeoExperiment.promptsLocation`, never from the strategy object itself.
- **Field placement** — `experimentId` and `baselinePrompts` are appended after `goalType` rather than adjacent to `type`; preserves existing field ordering. Move them up if you prefer cohesion over ordering stability.
- **Test coverage of the seven superRefine branches** — anything missing? Each invariant has at least one accept and one reject test; coverage is 100%.

## Cross-repo PR links
First in the program — no upstream PRs. Downstream PRs (A1, A2+A3, B1, B3-BE, B3-FE, C2, C3+C4, A4, A5) will link back here.

### DIFF
--- a/packages/spacecat-shared-utils/src/strategy-schema.js
+++ b/packages/spacecat-shared-utils/src/strategy-schema.js
@@ -166,12 +166,16 @@ export const strategyWorkspaceData = z.object({
       });
     }
 
-    // Evolving must have a non-empty baselinePrompts array
-    if (strat.type === 'evolving' && (!Array.isArray(strat.baselinePrompts) || strat.baselinePrompts.length === 0)) {
+    // Evolving: if baselinePrompts is provided, it must be non-empty.
+    // Soft rule by design — does NOT require the field to be present, so
+    // pre-GA strategies (which lack baselinePrompts) parse cleanly. The
+    // "newly-created Evolving strategies must include baselinePrompts" rule
+    // is enforced at the API layer in saveStrategy (see PR C2).
+    if (strat.type === 'evolving' && Array.isArray(strat.baselinePrompts) && strat.baselinePrompts.length === 0) {
       ctx.addIssue({
         code: 'custom',
         path: ['strategies', strategyIndex, 'baselinePrompts'],
-        message: 'Evolving strategies require a non-empty baselinePrompts array',
+        message: 'Evolving strategies must not have an empty baselinePrompts array (omit the field if not yet set)',
       });
     }
   });
--- a/packages/spacecat-shared-utils/src/strategy-schema.js
+++ b/packages/spacecat-shared-utils/src/strategy-schema.js
@@ -165,17 +165,16 @@ export const strategyWorkspaceData = z.object({
       });
     }
 
-    // Evolving: if selectedPrompts is provided, it must be non-empty.
-    // Soft rule by design — does NOT require the field to be present, so
-    // pre-GA strategies (which lack selectedPrompts) parse cleanly. The
-    // "newly-created Evolving strategies must include selectedPrompts" rule
-    // is enforced at the API layer in saveStrategy (see PR C2).
-    if (strat.type === 'evolving' && Array.isArray(strat.selectedPrompts) && strat.selectedPrompts.length === 0) {
-      ctx.addIssue({
-        code: 'custom',
-        path: ['strategies', strategyIndex, 'selectedPrompts'],
-        message: 'Evolving strategies must not have an empty selectedPrompts array (omit the field if not yet set)',
-      });
-    }
+    // Evolving: schema is intentionally permissive on `selectedPrompts`. Today's
+    // CreateStrategyDialog initializes selectedPrompts as `[]` for newly-created
+    // Evolving strategies (no prompt-selection UI yet). Enforcing "must be
+    // non-empty" at the schema would reject those in-flight reads + writes.
+    //
+    // The "newly-created Evolving strategies must include selectedPrompts" rule
+    // is deferred to the milestone where Evolving promotes out of co-innovation
+    // mode — at that point the schema tightens, the API layer (saveStrategy)
+    // adds the creation-time check, and the prompt-selection screen ships.
+    // All three land together. Until then, selectedPrompts on Evolving is
+    // structurally validated only.
   });
 });
--- a/packages/spacecat-shared-utils/src/strategy-schema.js
+++ b/packages/spacecat-shared-utils/src/strategy-schema.js
@@ -96,7 +96,6 @@ const strategy = z.object({
   completedAt: z.string().optional(), // ISO 8601 date string
   goalType: strategyGoalType.optional(),
   experimentId: z.uuid().nullable().optional(),
-  baselinePrompts: z.array(strategyPromptSelection).optional(),
 });
 
 /**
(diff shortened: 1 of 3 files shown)


AdrienCappelaere:
## Follow-up commit: softened Evolving invariant for backward-compat

The original PR body's "Reviewer focus" called out a backward-compat risk: the strict Evolving invariant (`baselinePrompts` must be non-empty) would reject pre-GA strategies on read, since they don't have the field yet. After pilot review, this was confirmed as a spec design hole.

**Pushed `1e0a1cd8` to address it.** The schema invariant is now soft: it only fires when `baselinePrompts` is *defined-but-empty*, not when absent. This means:
- Pre-GA strategies on Adobe.com / GM / Lovesac (which lack the field) parse cleanly — no migration required.
- Malformed new payloads (`baselinePrompts: []`) are still rejected.
- The "newly-created Evolving strategies must include `baselinePrompts`" business rule moves to the API layer in `saveStrategy` (planned for PR C2 in the GA program), separating schema validity from creation-time invariants.

**Tests adjusted:** the "rejects Evolving without baselinePrompts" test became "accepts Evolving without baselinePrompts (backward-compat for legacy data)"; new test added asserting the true legacy shape (no `type`, no `baselinePrompts`) parses cleanly. 1050 passing, 100% coverage maintained on `strategy-schema.js`.

**Spec updated** at `/Users/ahenna/Documents/Code/specs/opportunity-workspace-ga/2026-04-30-ga-design.md` §3 (schema rules) and §4 (lock enforcement) to reflect the corrected design.

This catches the backward-compat issue before any downstream PR (A1, B3-BE, C2) deploys the new schema to production.

packages/spacecat-shared-utils/src/strategy-schema.js
```diff
@@ -90,6 +95,8 @@ const strategy = z.object({
   createdBy: z.string().optional(), // Email of strategy creator/owner
   completedAt: z.string().optional(), // ISO 8601 date string
   goalType: strategyGoalType.optional(),
+  experimentId: z.uuid().nullable().optional(),
+  baselinePrompts: z.array(strategyPromptSelection).optional(),
```
jjenscodee: This field is unneeded. We have selectedPrompts: z.array(strategyPromptSelection).optional() already in the strategy.

packages/spacecat-shared-utils/src/strategy-schema.js
```diff
@@ -138,4 +145,38 @@ export const strategyWorkspaceData = z.object({
...
+    // "newly-created Evolving strategies must include baselinePrompts" rule
+    // is enforced at the API layer in saveStrategy (see PR C2).
+    if (strat.type === 'evolving' && Array.isArray(strat.baselinePrompts) && strat.baselinePrompts.length === 0) {
+      ctx.addIssue({
+        code: 'custom',
+        path: ['strategies', strategyIndex, 'baselinePrompts'],
+        message: 'Evolving strategies must not have an empty baselinePrompts array (omit the field if not yet set)',
+      });
+    }
+  });
```
jjenscodee: Could you remind me if we’ve already decided that all strategies must have prompts tied to them?

In other words, does this mean users can no longer select only topics or URLs, and must always select specific prompts?

packages/spacecat-shared-utils/src/strategy-schema.js
AdrienCappelaere: Good catch — you're right, this was redundant. I missed `selectedPrompts` during brainstorming and treated `baselinePrompts` as net-new without checking what already existed. Same type, same semantic role, no use case where they'd differ.

Pushed `9363c0e7` removing `baselinePrompts` and switching the superRefine invariants to reference `selectedPrompts`:
- Atomic must not carry `selectedPrompts` (prompts come from `GeoExperiment.promptsLocation`)
- Evolving: if `selectedPrompts` is defined, must be non-empty (soft rule, preserves legacy reads)

Schema delta is now **2 new fields** (`type`, `experimentId`) — down from 3. Tests adjusted; 1050 passing, 100% coverage maintained.

packages/spacecat-shared-utils/src/strategy-schema.js
AdrienCappelaere: Yes, decided — Evolving strategies require prompts at GA. Documented in the [wiki](https://wiki.corp.adobe.com/spaces/AEMSites/pages/3847829130) §2.1: "Mandatory prompt-selection wizard step. User picks ≥1 prompt from existing Brand Presence prompts. Cannot skip, cannot save without selection."

Going further than the original spec: Adrien confirmed today that this becomes a **dedicated screen inside the create-strategy wizard** (not just an inline form field). The dedicated screen will be the canonical place to browse, search, and select prompts during Evolving strategy creation.

**Two important boundaries on this F1 PR specifically:**

1. The schema rule in this PR is intentionally **soft** — it does NOT require `selectedPrompts` to be present, only that if provided, the array is non-empty. This is so pre-GA Evolving strategies on Adobe.com / GM / Lovesac (which lack `selectedPrompts`) parse cleanly when read. No migration needed.

2. The **hard "must include selectedPrompts on creation"** rule lands in PR **C2** (`saveStrategy` server-side enforcement), not here. That's the right place for creation-time business invariants — the schema is for structural validity.

**Open question for C2 you've helpfully surfaced:** before C2 ships, we need a quick audit — do current Adobe.com / GM / Lovesac strategies actually have `selectedPrompts` populated? If not, the C2 rule will block edits to existing strategies, and we'll need to either (a) backfill, (b) carve out a legacy exception in the validator, or (c) accept the breakage and tell customers to add prompts on next edit. Tagging this for the C2 design conversation.

jjenscodee: I checked the UI code, and selectedPrompts is currently initialized as an empty array. We will need a backfill then before we merge this PR

AdrienCappelaere: Good catch — you're right that the empty-array rule conflicts with how `CreateStrategyDialog` initializes today.

Two ways to resolve it:

1. Build out the prompt-selection UI for Evolving strategies before merging — your suggestion
2. Soften the schema rule to drop the empty-array constraint. The schema ships as a structural-only foundation; the "must have prompts" enforcement defers to the API layer.

Given we've aligned on prioritizing Atomic for GA and keeping Evolving in co-innovation mode (flag-gated for existing customers, promotes out of flag post-GA), the prompt-selection UI work for Evolving is part of the co-innovation polish that lands when Evolving promotes — not GA-blocking.

So path 2 is the right fit: drop the empty-array invariant from this PR; the schema ships as foundation, and the prompt-mandatory enforcement (plus the prompt-selection screen) land together post-GA when Evolving promotes. The two Atomic invariants stay.

Pushed `a42d7e35` dropping that rule. Test "rejects Evolving with empty selectedPrompts" became "accepts Evolving with empty selectedPrompts (FE init shape)". 1050 passing, 100% coverage maintained.

Does that work?

jjenscodee (approved):

PR #1589, opened 2026-05-05
fix(drs-client): reject non-XLSX uploads before writing to S3

irenelagno:
## Summary

- Adds `XLSX_MAGIC` constant (`PK\x03\x04`) to `drs-client/src/index.js`
- Guards `uploadExcelToDrs()` with a magic-bytes check before `PutObjectCommand` is called — any buffer that doesn't start with the ZIP/XLSX signature is rejected with a descriptive error
- Prevents SharePoint HTML error pages (and other non-XLSX content) from being written to `s3://drs-v2-prod-bp/external/spacecat/{site_id}/{job_id}/source.xlsx`, which was causing DRS Fargate tasks to fail with `openpyxl: File is not a zip file`

## Root cause

SpaceCat Lambda followed an expired SharePoint URL, received an HTML error page, and uploaded it to S3 without content validation. Four brand presence analyses failed for Astral Pool and Zodiac Pool on 2026-05-05 across gemini, perplexity, and chatgpt_paid.

## Test plan

- [x] New test: `throws when excelBuffer does not start with XLSX magic bytes` — passes an HTML buffer, asserts rejection with size in message
- [x] Updated existing tests that used bare `Buffer.from('x')` / `'fake-excel-bytes'` to supply a valid `PK\x03\x04` prefix
- [x] All 76 tests pass, 100% line/branch/statement/function coverage

### DIFF
--- a/packages/spacecat-shared-drs-client/src/index.js
+++ b/packages/spacecat-shared-drs-client/src/index.js
@@ -17,6 +17,8 @@ import { randomUUID } from 'crypto';
 
 const EXTERNAL_SPACECAT_PROVIDER_ID = 'external_spacecat';
 const DRS_S3_KEY_PREFIX = 'external/spacecat';
+// XLSX files are ZIP archives; all valid .xlsx start with PK\x03\x04
+const XLSX_MAGIC = Buffer.from([0x50, 0x4B, 0x03, 0x04]);
 
 export const EXPERIMENT_PHASES = Object.freeze({
   PRE: 'pre',
@@ -468,6 +470,9 @@ export default class DrsClient {
     }
     if (!excelBuffer || excelBuffer.length === 0) {
       throw new Error('excelBuffer is required and must be non-empty');
+    }
+    if (!Buffer.from(excelBuffer).slice(0, 4).equals(XLSX_MAGIC)) {
+      throw new Error(`Refusing to upload non-XLSX content to S3 (size=${excelBuffer.length})`);
     }
 
     const key = `${DRS_S3_KEY_PREFIX}/${siteId}/${jobId}/source.xlsx`;
--- a/packages/spacecat-shared-drs-client/src/index.js
+++ b/packages/spacecat-shared-drs-client/src/index.js
@@ -471,7 +471,7 @@ export default class DrsClient {
     if (!excelBuffer || excelBuffer.length === 0) {
       throw new Error('excelBuffer is required and must be non-empty');
     }
-    if (!Buffer.from(excelBuffer).slice(0, 4).equals(XLSX_MAGIC)) {
+    if (!Buffer.from(excelBuffer.subarray(0, 4)).equals(XLSX_MAGIC)) {
       throw new Error(`Refusing to upload non-XLSX content to S3 (size=${excelBuffer.length})`);
     }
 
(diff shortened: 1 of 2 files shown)


irenelagno:
## Related PRs — LLMO-4739 fix set

This is part of a 3-PR fix for the 2026-05-05 incident where SharePoint HTML error pages were written to S3 as `source.xlsx`, causing DRS Fargate tasks to fail.

| Repo | PR | Responsibility |
|------|----|---------------|
| `spacecat-shared` | **this PR** | Last-resort guard in `uploadExcelToDrs()` — reject non-XLSX before `PutObjectCommand` |
| `spacecat-audit-worker` | [#2477](https://github.com/adobe/spacecat-audit-worker/pull/2477) | Retry `readFromSharePoint()` up to 3× with 60s backoff before aborting upload |
| `llmo-data-retrieval-service` | [#1575](https://github.com/adobe-rnd/llmo-data-retrieval-service/pull/1575) | DRS-side safety net: pre-flight magic-byte check before Fargate ECS launch + fail-fast in `parse_excel_to_rows()` |

All three should merge together (or `spacecat-audit-worker` first, since it prevents the bad write at the source).

dzehnder (commented):
Hey @irenelagno,

## Strengths
- `uploadExcelToDrs()` is the sole `PutObjectCommand` path in `DrsClient`, so the guard at line 474 protects every caller without any consumer changes.
- Defense-in-depth framing is correct: audit-worker handles the retry upstream, drs-client is the last-resort backstop at the S3 boundary, DRS Fargate validates on ingest. Three independent layers for a recurring incident class is appropriate.
- Test is accurate to the incident -- HTML buffer rejected, existing happy-path tests updated so the guard is exercised rather than bypassed.
- No new dependencies.

## Issues

### Minor (Nice to Have)

**1. Redundant `Buffer.from` wrap at `packages/spacecat-shared-drs-client/src/index.js:474`.** `Buffer.from(excelBuffer.subarray(0, 4))` copies 4 bytes needlessly. `excelBuffer.subarray(0, 4)` already returns a `Buffer` when `excelBuffer` is a `Buffer` (per the `excelBuffer.length` guards above), and `Buffer.prototype.equals` works on it directly. The second commit message claims "zero-copy subarray" but the `Buffer.from()` wrap negates that. Simplify to:
```js
if (!excelBuffer.subarray(0, 4).equals(XLSX_MAGIC)) {
```

**2. No `log.warn` before the throw.** When this guard fires in prod, the only signal comes from whatever consumer catches and re-logs the error. The companion `spacecat-audit-worker#2477` logs a structured error before throwing. One line before the throw would make this guard self-observable in Coralogix:
```js
this.log.warn('Refusing non-XLSX upload to S3', { siteId, jobId, size: excelBuffer.length });
```

## Recommendations
- `XLSX_MAGIC` duplication with audit-worker is acceptable for a 4-byte file-format sentinel. Exporting an `isXlsx()` predicate would create more coupling than it saves in DRY.

## Assessment
Ready to merge? Yes.
Reasoning: Correct placement, clean implementation, accurate tests. The two minor items (drop the redundant `Buffer.from` wrap, add a `log.warn` before throw) are polish that can ship in a follow-up commit or be skipped.

dzehnder (approved):

PR #1591, opened 2026-05-06
fix(utils): update calculateCPCValue S3 path from ahrefs to seo

danieljchuser:
## Summary
- Fixes the `calculateCPCValue` function in `spacecat-shared-utils/metrics-store.js` which still read metrics from the old `metrics/{siteId}/ahrefs/organic-traffic.json` S3 path
- The import-worker now writes to `metrics/{siteId}/seo/` paths after the ahrefs→seo rebrand (spacecat-shared#1499), causing this function to silently fall back to default CPC ($0.80)
- Updates test fixture data and assertion to match the new path

Refs: [SITES-43199](https://jira.corp.adobe.com/browse/SITES-43199)

## Test plan
- [x] `metrics-store.test.js` — all 16 tests pass
- [x] Full `spacecat-shared-utils` test suite — 100% coverage on `metrics-store.js`
- [x] Lint clean

### DIFF
--- a/packages/spacecat-shared-utils/src/metrics-store.js
+++ b/packages/spacecat-shared-utils/src/metrics-store.js
@@ -106,7 +106,7 @@ export async function calculateCPCValue(context, siteId) {
   }
   const { s3Client, log } = context;
   const bucketName = context.env.S3_IMPORTER_BUCKET_NAME;
-  const key = `metrics/${siteId}/seo/organic-traffic.json`;
+  const key = createFilePath({ siteId, source: 'seo', metric: 'organic-traffic' });
   try {
     const organicTrafficData = await getObjectFromKey(s3Client, bucketName, key, log);
     if (!Array.isArray(organicTrafficData) || organicTrafficData.length === 0) {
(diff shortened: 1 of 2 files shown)


danieljchuser (commented):
Hey @danieljchuser,

## Strengths
- Surgical fix that correctly resolves a silent regression: `calculateCPCValue` returned the default CPC ($0.80) for all sites after the ahrefs-to-seo rebrand (spacecat-shared#1499) because it read from a path the import-worker no longer populates.
- The follow-up commit uses the existing `createFilePath` helper instead of re-hardcoding the path (`metrics-store.js:109`), eliminating the structural cause of this bug - two hardcoded paths that drift independently.
- Test assertion at `metrics-store.test.js:250` locks in the new path.
- Existing test suite exercises all three fallback branches, so coverage is maintained.

## Recommendations
- The "Organic traffic data not available" warning on fallback is indistinguishable between a legitimately new site and a fully broken path (the state during this bug). A follow-up to differentiate these or track fallback rate would surface this class of bug earlier.
- Worth sweeping the wider ecosystem for any remaining `metrics/{siteId}/ahrefs/...` references with the same silent-fallback bug.

## Assessment
Ready to merge? Yes

Correct fix for a real silent-failure bug, now using the shared path helper to prevent recurrence. Clean scope, tests updated.

LeChef318 (approved):
LGTM

PR #1593, opened 2026-05-07
feat(data-access): extend DOMAIN_PATTERN to support subpath domains (LLMO-4187)

Kanishkavijay39:
### Problem

Domain validation in `PlgOnboarding` used a regex that only accepted plain
hostnames (e.g. `nba.com`). Subpath domains like `nba.com/kings` were
rejected, blocking PLG and LLMO customers from self-onboarding subpath sites.

### Changes

**`plg-onboarding.model.js`**
- Updated `DOMAIN_PATTERN` to allow optional path segments after the hostname
  (e.g. `nba.com/kings`, `nba.com/us/kings`)
- Added IPv4 negative lookahead so addresses like `127.0.0.1` are rejected
- Simplified label sub-pattern to stay within SonarQube regex complexity limit (20)
- Removed duplicate `_` from `DOMAIN_PATTERN` character class (cosmetic)

**`plg-onboarding.schema.js`**
- Fixed the `domain` attribute length check to measure the hostname only
  (`value.split('/')[0].length <= 253`) so the path segment doesn't count
  against the 253-char DNS limit
- Reject path segments ending with a dot (`foo.`, `foo..`, `foo../bar`) — these
  can be misinterpreted as traversal forms on some canonicalizers
- Reject non-printable/control characters (null bytes etc.) at the validator
  boundary before any downstream processing

### Valid after this change

| Value | Before | After |
|-------|--------|-------|
| `nba.com` | ✅ | ✅ |
| `www.nba.com` | ✅ | ✅ |
| `nba.com/kings` | ❌ | ✅ |
| `nba.com/us/kings` | ❌ | ✅ |
| `example.com/foo_bar` | ❌ | ✅ |
| `https://nba.com` | ❌ | ❌ |
| `127.0.0.1` | ✅ (bug) | ❌ |
| `nba.com?q=1` | ❌ | ❌ |
| `nba.com#top` | ❌ | ❌ |
| `nba.com/foo.` | ✅ (bug) | ❌ |
| `nba.com/foo..` | ✅ (bug) | ❌ |
| `foo_bar.com` | ✅ (bug) | ❌ |
| `nba.com\x00/evil` | ✅ (bug) | ❌ |

## Local testing
image
image

## Related Issues

- [LLMO-4187](https://jira.corp.adobe.com/browse/LLMO-4187)

---

Thanks for contributing!


Kanishkavijay39 (commented):
Testing done

solaris007 (changes requested):
Hey @Kanishkavijay39,

Reviewed this as the foundation the api-service and ESS UI changes build on. The validator itself is correct and the test catalog is genuinely strong - my one substantive ask is about what the package exports, because that determines whether the companion PR can consolidate onto it without regressing.

## Strengths
- The tightening is right and enforced at the persistence layer, which is the correct last line of defense regardless of which HTTP path wrote the row: the old pattern accepted single-label hostnames, bare IPs, and underscored hostnames; the new one does not.
- `validate()` (plg-onboarding.schema.js:31) is a clean, fail-fast layered check - typeof guard, lowercase rejection (not silent coercion), control-character rejection, all-numeric-hostname rejection, trailing-dot path rejection, then pattern and length caps. It returns false (no throw) for null/undefined/number/object.
- Exemplary boundary tests: 253/254 hostname, 2048/2049 total, short-form IP, decimal IP, null byte, control char, and the subtle "path length does not inflate the hostname cap" case - each asserting the fixture length before validating. The catalogs are DAMP (labeled, distinct cases), not per-input mirroring; the ~4:1 test-to-code ratio is appropriate for a validator whose value is the breadth of inputs it accepts and rejects.
- ReDoS-safe: both regexes are linear (delimiter-anchored groups, bounded quantifiers) and input is capped at 2048.
- Regression-safe in this repo: the only PlgOnboarding domain fixtures (`example.com`, `staging.example.com`, `stage1.lovesac.com`, `staging.lovesac.com`) are all multi-label, lowercase, and non-IP, so they remain valid; and `domain` is `readOnly` (validated on write only), so existing rows are not retroactively invalidated.
- The comments are honest about the design - they say outright that `DOMAIN_PATTERN` is not sufficient alone and point to the schema validator.

## Issues

### Important
**The package exports the partial validator, not the complete one (plg-onboarding.model.js:48, plg-onboarding.schema.js:31).** The safety-critical checks - all-numeric/short-form-IP rejection, control-character rejection, length caps, trailing-dot rejection - live inside the inline `validate()` closure, which is not exported. What a consumer can import is the bare `DOMAIN_PATTERN` (plus `normalizeDomain` and the length constants), and the bare pattern accepts `127.1`, `0x7f.0.0.1`, `nba.com/foo.`, `nba.com/foo..`, with no length cap. The companion adobe/spacecat-api-service#2363 explicitly plans to import `DOMAIN_PATTERN` to "stay in sync," which would silently re-open exactly those holes in the consumer. This is not exploitable inside spacecat-shared (the schema always runs the full check and the data-access layer never makes outbound fetches), so it is Important rather than Critical, but it is an attractive-nuisance API surface. Lift the `validate()` body into a reusable `static isValidDomain(value)` on `PlgOnboarding` and have the schema delegate to it, so the schema and external consumers share one implementation - and pin it with a test that the bare `DOMAIN_PATTERN` accepts `127.1` while `isValidDomain` rejects it. That single change is what lets #2363's consolidation be correct instead of a regression.

### Minor
- **The lowercase-only rule also forces the path lowercase (plg-onboarding.schema.js:31, model.js:48).** Hostnames are case-insensitive, but URL paths are case-sensitive (RFC 3986), so `nba.com/Kings` is rejected and a customer whose site lives at a mixed-case path cannot onboard. If that is intentional canonicalization, document it; if not, the lowercase rule should apply to the host segment only. This is also the axis on which the companion api-service validator (case-insensitive) disagrees with this one.
- **The IPv4 negative lookahead `(?!\d+(\.\d+){3})` over-rejects (plg-onboarding.model.js:48).** It rejects any string beginning with a dotted-quad, including legitimate hostnames like `1.2.3.4.example.com` or `192.168.1.1.nip.io` (nip.io-style wildcard DNS). It errs safe and is partly redundant with the schema's `/^[\d.]+$/` all-numeric guard, but a customer on such a host would be blocked with no obvious reason. Worth a conscious decision; consider anchoring the IPv4 check to a whole-hostname match.
- **The `typeof value !== 'string'` guard (plg-onboarding.schema.js:32) is untested.** The invalid catalog is all strings, so the guard's true branch never runs. Add `validate(null)`, `validate(123)`, `validate(undefined)` -> false. Without the guard a non-string would throw on `.toLowerCase()`.

## Recommendations
- Document explicitly that `validate()` is a syntax and data-integrity validator, not an SSRF gate (the data-access layer performs no outbound fetches; SSRF defense belongs at the fetch layer). And if the exported validator is ever intended to back a fetch decision, the all-numeric check needs to also reject hex/octal/mixed IP forms - `0x7f.0.0.1` currently passes the full `validate()`.
- `normalizeDomain` is exported and tested but has no caller in this repo, and the documented "call before `findByImsOrgIdAndDomain`" contract is unenforced (the finder is index-generated and does not normalize; nothing internal normalizes). Without enforced normalization, `NBA.COM` and `nba.com` are distinct sort-key values and can create duplicate onboarding rows - the exact failure the comment warns about. Either wire it into a path that matters or relabel it as consumer-only forward-looking API.

- The durable fix for the same validation rule living in three repos (this model, the api-service controller, and the ESS UI `getDomainFromUrl.ts`) is a single zero-dependency validation module that all three import, rather than three regexes kept in sync by comment. Likely a thin policy wrapper over the WHATWG URL parser. Its own ticket.

## Assessment
With fixes. The schema validation is correct and well-tested on its own merits. Because this is about to be the published API the companion imports, ship the complete validator (not just `DOMAIN_PATTERN`) first - it is a small, in-scope change here and it is the linchpin that makes the api-service consolidation safe.

## Next Steps
1. Export a complete `isValidDomain(value)` and have the schema delegate to it (Important).
2. Decide the path case-sensitivity question and add the non-string guard test (Minor).
3. Wire up or relabel `normalizeDomain`, and document `validate()` as non-SSRF (Recommendations).

solaris007 (changes requested):
Hey @Kanishkavijay39,

Re-reviewed the two new commits. The validator consolidation is exactly what was asked for: `validate()` is lifted into the complete `PlgOnboarding.isValidDomain`, the schema delegates to it, the IPv4 lookahead is anchored, the non-string guard is tested, and the pinning-regression suite is a model example. Every prior finding is resolved.

One issue remains, and it is the root cause of a Critical SSRF on the companion api-service PR: the IP-literal rejection has a hex-label gap.

## Strengths
- Complete validator exported: the schema's inline `validate()` is lifted into a single `PlgOnboarding.isValidDomain` static and the schema delegates (`plg-onboarding.schema.js:93`). Consumers get the full check, not the bare pattern, and the JSDoc says so.
- IPv4 lookahead anchored correctly: `(?!\d+(\.\d+){3}(?:\/|$))` rejects `127.0.0.1` and `127.0.0.1/path` while accepting `1.2.3.4.example.com` and `192.168.1.1.nip.io`. Verified.
- Consecutive-dot tightening (`seg.includes('..')`) closes `foo..bar` / `v1..0` with matching tests.
- The pinning-regression suite (bare `DOMAIN_PATTERN` accepts `127.1` / `nba.com/foo.` / unbounded length while `isValidDomain` rejects) is the right guardrail against a future caller importing the bare pattern.
- Schema delegation is a pure refactor: `domain` is readOnly (write-time validation only), existing rows are lowercase multi-label non-IP, so nothing is retroactively invalidated.

## Issues

### Important (Should Fix)

**1. isValidDomain's IP-literal rejection is incomplete - hex-label hostnames pass. `plg-onboarding.model.js` (`DOMAIN_PATTERN` and `isValidDomain`).**

The JSDoc and the two IP guards intend to reject IP literals, but hex forms slip through: `0x7f.0.0.1`, `0xa.0.0.1`, and `0xa9.254.169.254` all return `true` from `isValidDomain`. The `/^[\d.]+$/` all-numeric guard misses them because hex contains `x`/`a-f`, and the anchored dotted-quad lookahead does not fire on a `0x...` prefix. In this package alone that is a data-integrity defect (it persists `0xa9.254.169.254` as a "valid domain", contradicting the validator's own stated intent). It becomes Critical in the companion adobe/spacecat-api-service#2363, where that service delegates to this validator and then fetches the domain server-side: `new URL` canonicalizes `0xa9.254.169.254` to `169.254.169.254` (the AWS metadata IP) and the service's string-prefix SSRF denylist does not catch the hex form. This is the cleanest single place to fix the root cause.

Fix: require an alphabetic or punycode final label (`[a-z]{2,}` or `xn--...`) in `DOMAIN_PATTERN`, matching the constraint the api-service's now-deleted local regex enforced. Verified this keeps `nba.com`, `www.nba.com`, `1.2.3.4.example.com`, `192.168.1.1.nip.io`, `xn--nba-6na.com`, and `nba.com/us/kings`, and rejects `0x7f.0.0.1`, `0xa9.254.169.254`, and `foo.1`. Add those to the `isValidDomain` invalid-values list and the "DOMAIN_PATTERN alone is insufficient" suite.

### Minor (Nice to Have)

**2. Length caps are checked after the regex runs. `plg-onboarding.model.js` (`isValidDomain`).**

`DOMAIN_PATTERN.test(value)` runs on the full untrusted input before the length check. The pattern is linear (anchored, bounded quantifiers), so this is not a ReDoS vector today - cheap defense in depth only. Checking `value.length`/`hostname.length` first would bound work on pathological input. Low priority, and partial (the control-char regex already scans the full input first).

## Recommendations
- After the IP-literal fix, tighten the `isValidDomain` JSDoc to state that IP-literal hostnames are rejected via the alphabetic-TLD requirement, so downstream callers do not over-trust the all-numeric check.
- The dual export (both the incomplete `DOMAIN_PATTERN` and the complete `isValidDomain` reachable on the public model) is an acceptable foot-gun for now given the JSDoc and pinning tests, but when the validator moves to its own module, make the raw pattern module-private and export only `isValidDomain`.
- Add a changelog note that the validator now accepts more forms (`1.2.3.4.example.com`, nip.io-style hosts) so any consumer doing domain-keyed dedup is aware.

- The durable fix for the same rule living in three repos (this model, the api-service controller, the ESS UI `getDomainFromUrl.ts`) is one zero-dependency validation module all three import, with the model importing the leaf rather than owning it.

## Assessment
Ready to merge? With fixes. The extraction, anchor fix, consecutive-dot tightening, and pinning tests are well built and all prior findings are resolved. The one should-fix is the hex-label IP-literal gap - closing it here is the cleanest root-cause fix for the companion SSRF, and it is a small, in-scope change. Land this first (with the fix) so api-service can repoint off its gist tarball.

## Next Steps
1. Add the alphabetic/punycode TLD requirement to close the hex-label gap (Important), with tests.
2. Length-cap ordering and JSDoc/changelog notes are optional.

solaris007 (approved):
Hey @Kanishkavijay39,

Picking the re-review back up after the branch advanced (commit `69408dac` plus a merge from main). The core `DOMAIN_PATTERN` is unchanged from the last reviewed commit, so the IP-literal and SSRF verification still holds, and the new commit cleanly addresses both points from the prior round. Two small test and consistency items remain, neither blocking.

## Strengths

- **Previously flagged hex-label IP-literal gap stays resolved.** The structural requirement that the final label be alphabetic (`[a-z]{2,}`) or punycode (`xn--[a-z0-9-]+`) rejects every IP-literal encoding in one rule (dotted-quad, short-form, decimal, hex including `0xa9.254.169.254`, all-hex, octal). The regex is unchanged in the new commits, so this remains verified. (plg-onboarding.model.js)
- **Length-cap ordering (previously pushed back) is now implemented, not just argued.** `isValidDomain` checks `value.length` and `hostname.length` before running the regex, with a comment explaining the intent (plg-onboarding.model.js:73,80). It is outcome-equivalent to the prior end-of-chain checks - the 253/254 and 2048/2049 boundary tests confirm - so it is a safe reorder.
- **The two stale "IPv4 lookahead" comments are fixed.** Both valid-value blocks in plg-onboarding.model.test.js now describe the alphabetic/punycode TLD requirement instead of the deleted lookahead, matching the model.js JSDoc.
- **The bare-pattern foot-gun guard got stronger.** A prominent WARNING block on `DOMAIN_PATTERN` now spells out exactly what the raw regex omits (length cap, control-character rejection, all-numeric check, dot-path rejection, typeof guard) and directs consumers to `isValidDomain`, on top of the existing pinning suite. (plg-onboarding.model.js:16-21)
- **The fix is complete by construction, not just by test catalog.** WHATWG URL only enters IPv4 canonicalization when the final label is all-digits or parses as an IPv4 number; an alphabetic or punycode final label can be neither, so that path is unreachable for any accepted value, and IPv6 needs characters outside the charset. Verified through hand-crafted attacks, fuzzing, and IDN-folding vectors - no accepted input canonicalizes to an IP via `new URL()` / `net.isIP`.
- **The `[^\x21-\x7e]` control-character guard closes the IDN-folding class.** Inputs like `１２７.0.0.1` (full-width digits) or `127。0。0。1` (ideographic dots) would otherwise fold to a dotted-quad IP via ToASCII; they are rejected before the pattern runs.
- **Tests are DAMP and layered.** Rejection is asserted at all three gates (DOMAIN_PATTERN, isValidDomain, schema `domain` attribute), the SSRF suite documents each input's canonicalization target, and the obsolete `127.1` pin was correctly migrated. Net coverage went up.
- **Safe tightening for the running system.** `domain` is `readOnly` and validated only on write, so stored rows are never re-validated; no real DNS uses numeric or single-character TLDs (IDN TLDs are `xn--`, accepted), so no legitimate onboarding is newly blocked; and the pattern is linear, so there is no ReDoS surface.

## Issues

### Minor (Nice to Have)

- **The `xn--[a-z0-9-]+` punycode-TLD branch still has no positive test** (plg-onboarding.model.js DOMAIN_PATTERN). This branch exists to keep accepting IDN ccTLDs (e.g. `.рф` = `xn--p1ai`), but no valid-catalog entry exercises it - removing the branch leaves the entire valid catalog green. The existing `xn--nba-6na.com` cases only exercise a punycode host label with a `.com` TLD, so they pass via the `[a-z]{2,}` branch. A regression that dropped IDN-TLD support would go uncaught. Add a punycode-TLD entry such as `example.xn--p1ai` to the valid-values catalogs (it returns true today).
- **The length cap does not fully deliver the O(1) rejection its comment promises** (plg-onboarding.model.js, isValidDomain). The new comment says oversized input is "rejected in O(1) rather than driving a multi-MB regex scan," but the `value !== value.toLowerCase()` check on the line above already allocates and scans the full input, so a multi-MB value does O(n) work before the length cap fires. This is not a real DoS vector (the write path is bounded upstream), purely a consistency point: if you want the guarantee the comment states, split the typeof and lowercase checks and put the `value.length > MAX_DOMAIN_LENGTH` check between them (typeof, then length, then lowercase, then the rest).

## Recommendations

- Add IDN-folding regression cases to the SSRF suite so the exact property this commit protects is pinned: assert that `１２７.0.0.1` (full-width digits) and `127。0。0。1` (ideographic dots) are rejected. The `[^\x21-\x7e]` guard is the only thing between those inputs and a folded IMDS IP, but no SSRF-labeled test asserts it today.
- Optionally make the trailing path group non-capturing (`(?:\/(?!\.)[a-z0-9._~-]+)*`) to match the now non-capturing label groups; the pattern is only used via `.test()`, so this is a pure consistency nit that would leave it with zero capture groups.
- Optionally add the all-hex form `0xa9.0xfe.0xa9.0xfe` to the DOMAIN_PATTERN-direct invalid block for explicitness (it is already exercised through isValidDomain), and add a one-line note that the new rule makes single-char alpha TLDs like `foo.a` invalid - harmless since no such public TLD exists, but it is an untested behavior change from the prior regex.

- The same domain rule lives in three repos (this model, the api-service controller, and the ESS UI `getDomainFromUrl.ts`). The durable fix is a single zero-dependency validation module all three import - already agreed by you and the reviewer as its own ticket. This commit's simpler allowlist formulation makes that consolidation easier, not harder.

## Assessment

Ready to merge? Yes. Both points from the prior round are now addressed in code; the remaining items are a missing positive test for the IDN-TLD branch and an optional check-ordering refinement, both non-blocking.

Reasoning: The validator closes the previously flagged IP-literal gap at the root, the fix is verified complete from code-level, architectural, security, and test angles, and it is a safe write-time tightening with no retroactive impact and no ReDoS surface. CI is green on the current head.

## Next Steps

1. Add a punycode-TLD positive case (`example.xn--p1ai`) to the valid catalogs.
2. Optionally move the length cap above the `toLowerCase()` comparison to match the comment's O(1) claim.
3. The IDN-folding rejection tests, the capturing-group nit, and the explicitness items are optional.

solaris007 (approved):
Hey @Kanishkavijay39,

Re-reviewed through the latest head (the IP-literal commit plus the follow-up chore 69408dac8 and the main merge). The blocking hex-label IP-literal gap is genuinely closed, and the chore went further than asked: it implemented the length-cap reorder and corrected the stale test comments. The security-critical regex is byte-identical to what I verified empirically (node plus a ~29k-host fuzz), so those results still hold. CI is green. Approving. One trivial Minor and a few optional polishes remain, none blocking.

## Strengths
- **Previously flagged hex-label IP-literal gap is resolved (`DOMAIN_PATTERN` in plg-onboarding.model.js).** The alphabetic/punycode TLD requirement (`[a-z]{2,}|xn--[a-z0-9-]+`) rejects every IP-literal canonicalization form: dotted-quad, short-form, decimal-dword, hex (with and without dots), octal, mixed, zero-padded, and trailing-dot. WHATWG URL only parses a host as IPv4 when the final label is numeric or hex, so an alphabetic TLD is structurally disjoint from the IP-canonicalizable set. Every form that `new URL()` resolves to an IP (including `0xa9.254.169.254` to the 169.254.169.254 metadata IP) is rejected, and a fuzz of ~29k pattern-accepted 4-label hosts produced 0 IP canonicalizations.
- **Inverting the denylist into an allowlist is the durable fix.** The old enumerate-the-IP-forms approach was always one encoding behind, which is how the hex gap existed. Keying off the one property that triggers IPv4 canonicalization closes the whole class in a single rule.
- **Length caps now short-circuit before the regex (`isValidDomain`, commit 69408dac8).** You implemented the reorder you had offered: the total-length and hostname-length checks now return early, so a multi-MB pathological input is rejected in O(1) instead of driving a full regex scan. I had been ready to drop this as non-blocking (the regex is provably linear), but doing it is a clean defense-in-depth win for external callers that may not bound input upstream, and I confirmed the reorder is behavior-preserving on the 253/254 and 2048/2049 boundaries.
- **The export surface is now hard to misuse.** `DOMAIN_PATTERN` carries an explicit "do NOT use directly" warning enumerating what it omits (length cap, control-char, all-numeric, dot-path, typeof guard), and the schema delegates to `isValidDomain`. The two test comments now correctly attribute IP rejection to the TLD requirement rather than the removed lookahead.
- **No legitimate-domain regression, no migration risk.** nip.io wildcard DNS (`192.168.1.1.nip.io`), punycode (`xn--nba-6na.com`), numeric-prefixed hosts (`1.2.3.4.example.com`), and subpath domains (`nba.com/us/kings`) all still validate. `domain` is `readOnly` (validated on write only), so existing rows are not retroactively invalidated.
- **ReDoS-safe.** The pattern is delimiter-anchored with bounded quantifiers; timed against catastrophic-backtracking shapes up to 500K chars it stays linear and sub-millisecond.
- **Regression coverage at all three layers.** IP-literal cases are pinned at the pattern, validator, and schema levels. The `127.1` pinning test was correctly removed (its "pattern accepts, validator rejects" premise is now false) and replaced with a still-valid `nba.com/v1..0` consecutive-dot pin.
- The non-ASCII guard (`[^\x21-\x7e]`) also rejects homoglyph-dot forms (such as `0x7f。0。0。1` with U+3002) before they could canonicalize, a vector many TLD-rule fixes miss.

## Issues

### Minor (Nice to Have)
- **Punycode TLD branch permits a trailing hyphen (`DOMAIN_PATTERN` in plg-onboarding.model.js).** `xn--[a-z0-9-]+` accepts `example.xn--abc-`. The normal-label branch correctly forbids leading and trailing hyphens, and the adjacent comment states "Labels must not start or end with a hyphen (RFC 1035)", so the punycode branch contradicts both its sibling branch and its own doc. A trailing-hyphen ACE label is not valid IDNA. Impact is effectively nil (it only over-accepts a non-resolvable string into a write-time field), so this is optional. For consistency: `xn--[a-z0-9-]*[a-z0-9]`.

## Recommendations
- **Make the SSRF-defense test column load-bearing, or drop it (the "SSRF defense" block in plg-onboarding.model.test.js).** The rows are `[label, input, canonicalizedTargetIP]` but the iterator destructures only `[label, input]`, so the documented target IP (for example `169.254.169.254`) is never asserted and a wrong value there would fail nothing. Either assert it (`new URL('https://' + input).hostname === expectedIP`), which proves the canonicalization premise the block documents, or drop the column and fold its one unique input (`0xa.0.0.1`, the RFC1918 case) into the existing invalid catalog (the other four duplicate it).
- The all-numeric `/^[\d.]+$/` guard in `isValidDomain` is now strictly subsumed by the TLD rule. It is correctly labeled as intentional defense-in-depth, so it is not accidental complexity today; the moment to reconsider it is when this validator graduates to its own module.
- Optional consistency: decimal `2130706433` is pinned at the `isValidDomain` level but not at the bare `DOMAIN_PATTERN` level (it is already covered by the pattern's single-label rejection). A one-line pattern-level case would treat it like the other IP forms.

## Prior findings
- **[Important] Hex-label IP-literal gap - resolved** (see Strengths), verified empirically rather than from the JSDoc alone.
- **[Minor] Length caps checked after the regex - resolved by implementation in 69408dac8.** You both made the linear-regex argument and then reordered the checks anyway; either way the thread is closed, and the reorder is the stronger outcome for external consumers.

- The same validation rule lives in three repos (this model, the api-service controller, and the ESS UI `getDomainFromUrl.ts`); the durable fix is one zero-dependency module all three import. You have already flagged this as its own ticket, noting once, not as a finding.
- Heads-up for the companion api-service change (adobe/spacecat-api-service#2363, a different repo, not a change here): a `net.isIP`-only canonicalization check will pass DNS-rebind hosts like `127.0.0.1.nip.io` (`net.isIP` returns 0 for them); only DNS-resolution-time IP checks close that class. This validator's comment already correctly assigns DNS-resolution checks to the caller.
- FYI, not a finding: the main merge pulled in an unrelated `PENDING` value on `REVIEW_DECISIONS` from main, outside this PR's scope.

## Assessment
Ready to merge? **Yes.** The blocking IP-literal gap from the two prior reviews is genuinely fixed with the right structural approach, verified empirically with no residual bypass and no legitimate-domain regression, and the follow-up chore tightened the length-cap ordering and export surface. CI is green. The single Minor (punycode trailing hyphen) and the test-polish recommendations are all non-blocking.

## Next Steps
1. Optional: tighten the punycode branch to `xn--[a-z0-9-]*[a-z0-9]` for label-rule consistency.
2. Optional: make the SSRF-defense test column assert the canonical IP, or fold its unique case into the catalog.
3. Otherwise this is good to land so the companion api-service change can repoint onto the corrected shared validator.

PR #1595, opened 2026-05-08
feat(http-utils): allow RO admin write ops on resources they own

ravverma:
## Problem

The `readOnlyAdminWrapper` previously blocked **all** write operations for read-only admin users. This was overly restrictive: a read-only admin whose IMS org owns a site or organization should be able to mutate that resource — they just should not be able to touch resources belonging to other orgs.

### `route-utils.js` — new `extractRouteParams`

Added a focused, standalone `extractRouteParams(context, routeMap)` function that returns the named path parameters extracted from the matched route pattern (e.g. `{ siteId: 'abc-123' }` from `PATCH /sites/:siteId`). The existing `resolveRouteCapability` is left completely unchanged so the s2s wrapper is unaffected.

### `read-only-admin-wrapper.js` — ownership check before blocking writes

For write actions, before returning 403 the wrapper now calls `isOwnerOfResource(context, authInfo, params)`:

- If `siteId` is in the path params: fetches `context.dataAccess.Site.findById(siteId)`, resolves the org, and calls `authInfo.hasOrganization(org.getImsOrgId())`
- If `organizationId` is in the path params: fetches `context.dataAccess.Organization.findById(organizationId)` and checks the same
- Fail-closed in all error cases: `dataAccess` absent, entity not found, no recognisable ID in the path, or any DB exception — the exception is logged via `log.error` and the request is blocked

Read paths are completely untouched.

### `context.data` fallback for ID-less routes

Some write routes carry no resource ID in the path but pass it in the request body instead (e.g. `POST /preflight/jobs` sends `siteId` in the JSON payload, resolved via `context.data.siteId`). For these cases `isOwnerOfResource` falls back to `context.data` after finding no match in the route params:

```js
const siteId = params.siteId ?? context.data?.siteId;
const organizationId = params.organizationId ?? context.data?.organizationId;
```

Path params always take precedence over body data when both are present. Routes that carry neither (e.g. `POST /sites` creating a new resource) still return false and are blocked.

## Test coverage

**`route-utils.test.js`** — 8 new cases for `extractRouteParams`: exact match (no params), single param, multiple nested params, org param, no match, missing method/suffix/pathInfo.

**`read-only-admin-wrapper.test.js`** — 15 new cases under "write on owned resource":
- Owned site write (path param) → allowed + audit log emitted
- Unowned site write → 403
- Site not found → 403
- Site with no org → 403
- Owned org write (path param) → allowed
- Organization not found → 403
- Unowned org write → 403
- No ID in path and no `context.data` (e.g. `POST /sites`) → 403
- No `dataAccess` on context → 403 (fail-closed)
- DB lookup throws → 403 + `log.error` called with the error (fail-closed)
- Owned site write via `context.data.siteId` (no path param) → allowed
- Unowned site via `context.data.siteId` → 403
- Owned org write via `context.data.organizationId` (no path param) → allowed
- Path param takes precedence over `context.data` when both present
- Audit log emitted on allowed owned write

All 309 tests pass. Coverage: 100% lines/statements, 98.18% branches (threshold: 97%).

## Checklist

- [x] `resolveRouteCapability` untouched — s2s wrapper unaffected
- [x] Fail-closed for every error path in the ownership check
- [x] Exceptions logged before returning false (no silent failures)
- [x] Ownership check is lazy — only runs for write routes, never for reads
- [x] `context.data` fallback handles routes like `POST /preflight/jobs` with no ID in path
- [x] Path params take precedence over body data when both are present
- [x] No new package dependencies

### DIFF
--- a/packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
+++ b/packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
@@ -34,7 +34,9 @@ function forbidden(message) {
  * IMS org matches the organization that owns the referenced site or organization entity.
  *
  * ID resolution order:
- * 1. Named path params (e.g. :siteId, :organizationId) — always preferred.
+ * 1. Named path params (e.g. :siteId, :organizationId, :spaceCatId) — always preferred.
+ *    spaceCatId is an alias for organizationId used by /v2/orgs/:spaceCatId/* routes
+ *    in spacecat-api-service.
  * 2. context.data (parsed request body) — used only when the matched route has no path
  *    params at all (e.g. POST /preflight/jobs passes siteId in the body). Requires that
  *    the body-parser wrapper runs before readOnlyAdminWrapper in the .with() chain.
@@ -57,13 +59,11 @@ async function isOwnerOfResource(context, authInfo, params) {
   // Body fallback only when the route carries no path params at all (e.g. POST /preflight/jobs).
   // When params has keys, the route does have path identifiers; relying on body data in that
   // case could allow a caller to spoof ownership by naming params differently than :siteId.
-  // spaceCatId is an alias for organizationId.
   const hasPathParams = Object.keys(params).length > 0;
-  const siteId = hasPathParams ? params.siteId : (params.siteId ?? context.data?.siteId);
+  const siteId = hasPathParams ? params.siteId : context.data?.siteId;
   const organizationId = hasPathParams
     ? (params.organizationId ?? params.spaceCatId)
-    : (params.organizationId ?? params.spaceCatId
-        ?? context.data?.organizationId ?? context.data?.spaceCatId);
+    : (context.data?.organizationId ?? context.data?.spaceCatId);
 
   try {
     if (siteId) {
@@ -75,7 +75,12 @@ async function isOwnerOfResource(context, authInfo, params) {
       if (!org) {
         return false;
       }
-      return authInfo.hasOrganization(org.getImsOrgId());
+      const imsOrgId = org.getImsOrgId();
+      if (!imsOrgId) {
+        log.warn({ tag: 'ro-admin', siteId }, 'Owning organization has no imsOrgId; denying RO admin access');
+        return false;
+      }
+      return authInfo.hasOrganization(imsOrgId);
     }
 
     if (organizationId) {
@@ -83,7 +88,12 @@ async function isOwnerOfResource(context, authInfo, params) {
       if (!org) {
         return false;
       }
-      return authInfo.hasOrganization(org.getImsOrgId());
+      const imsOrgId = org.getImsOrgId();
+      if (!imsOrgId) {
+        log.warn({ tag: 'ro-admin', organizationId }, 'Organization has no imsOrgId; denying RO admin access');
+        return false;
+      }
+      return authInfo.hasOrganization(imsOrgId);
     }
   } catch (err) {
     log.error({ tag: 'ro-admin', err }, 'Error checking resource ownership for RO admin');
@@ -138,12 +148,28 @@ async function evaluateFeatureFlag(context, authInfo) {
  * If so it:
  *
  * 1. Evaluates the `FT_READ_ONLY_ORG` LaunchDarkly feature flag (fail-closed).
- * 2. Resolves the route's action from the routeCapabilities map and blocks
- *    write operations (or unmapped routes) for RO admins, unless they own the resource.
- * 3. Emits a structured audit log entry for all allowed RO admin requests, including
- *    the resolved resource ID and whether it came from the path or request body.
+ * 2. For read routes (capability 'read' or 'readAll'): allows immediately without an
+ *    ownership DB lookup — capability alone permits.
+ * 3. For write routes and unmapped routes (all non-read capabilities, including null):
+ *    performs an ownership check on the referenced resource. Ownership is resolved from
+ *    path params first (:siteId, :organizationId, :spaceCatId) and falls back to the
+ *    request body (context.data) only when the matched route has no path params at all.
+ *    This allows RO admins to operate on their own resources even when the specific
+ *    route has no entry in routeCapabilities — the key invariant is ownership, not
+ *    capability mapping. Routes with no resolvable resource ID are denied.
+ * 4. Emits a structured access log (tag: ro-admin-access) when an allowed write is
+ *    authorized by ownership. Emits an audit log (tag: ro-admin-audit) for all other
+ *    allowed RO admin requests where an access log was not already emitted.
  *
  * Non-RO-admin requests pass through untouched.
+ *
+ * Wiring requirements (ordering within the .with() chain):
+ * - `dataAccessWrapper` must run before this wrapper so context.dataAccess is available.
+ * - The body-parser wrapper must run before this wrapper for routes that authorize from
+ *   context.data (collection writes where siteId/organizationId is in the body).
+ * - Handlers behind body-fallback authorization MUST only mutate the declared
+ *   siteId/organizationId resource; the wrapper cannot verify what the handler does
+ *   with other fields in the request body.
  *
  * @param {Function} fn - The handler to wrap.
  * @param {{ routeCapabilities: Object<string, string> }} opts - Required map of route
@@ -172,81 +198,69 @@ export function readOnlyAdminWrapper(fn, { routeCapabilities } = {}) {
       }
 
       if (isObject(routeCapabilities)) {
-        const params = extractRouteParams(context, routeCapabilities);
-        const hasPathParams = Object.keys(params).length > 0;
-
-        if (hasPathParams) {
-          // Parameterized route: ownership is checked first. Owners may proceed even
-          // when the route is absent from routeCapabilities (unmapped), avoiding an
-          // incorrect fail-closed denial for resources the RO admin owns.
-          const isOwner = await isOwnerOfResource(context, authInfo, params);
-          if (isOwner) {
-            log.info({
-              tag: 'ro-admin-access',
-              email: authInfo.getProfile?.()?.email,
-              method: context.pathInfo?.method,
-              suffix: context.pathInfo?.suffix,
-              org: authInfo.getTenantIds?.()[0],
-              resolvedSiteId: params.siteId ?? null,
-              resolvedOrgId: params.organizationId ?? params.spaceCatId ?? null,
-              idSource: 'path',
-            }, 'RO admin access allowed on owned resource');
-          } else {
-            // Not the owner: fall back to capability — only reads are permitted.
-            // capability format: 'resource:action' (e.g. 'site:read', 'site:write').
-            // A missing/unmapped route yields undefined action, which is not a read
-            // and correctly denies the request.
-            const capability = resolveRouteCapability(context, routeCapabilities);
-            const action = capability?.split(':').pop();
-            if (action !== 'read' && action !== 'readAll') {
-              log.warn({
-                tag: 'ro-admin',
-                email: authInfo.getProfile?.()?.email,
-                method: context.pathInfo?.method,
-                suffix: context.pathInfo?.suffix,
-                org: authInfo.getTenantIds?.()[0],
-              }, 'Read-only admin blocked from route');
-              return forbidden('Forbidden');
-            }
-          }
-        } else {
-          // Collection/no-params route: capability check first (no path-based ownership
-          // to verify). For write actions, fall back to body-supplied resource id.
+        let accessLogged = false;
+        try {
+          const params = extractRouteParams(context, routeCapabilities);
+          const hasPathParams = Object.keys(params).length > 0;
           const capability = resolveRouteCapability(context, routeCapabilities);
+          // capability format: 'resource:action' (e.g. 'site:read', 'site:write').
+          // split(':').pop() extracts the action; a missing or malformed value yields
+          // undefined, which is not a read action and correctly triggers the ownership check.
           const action = capability?.split(':').pop();
+
           if (action !== 'read' && action !== 'readAll') {
+            // Write or unmapped route: verify ownership of the referenced resource.
+            // Ownership is the governing invariant — a null/absent capability is not an
+            // automatic deny; it means the route is not mapped, but RO admins who own
+            // the resource may still perform the operation. Routes with no resolvable
+            // resource ID (no path param and no body siteId/organizationId) fail-closed.
             const isOwner = await isOwnerOfResource(context, authInfo, params);
-            if (!isOwner) {
+            if (isOwner) {
+              log.info({
+                tag: 'ro-admin-access',
+                email: authInfo.getProfile?.()?.email,
+                method: context.pathInfo?.method,
+                suffix: context.pathInfo?.suffix,
+                org: authInfo.getTenantIds?.()[0],
+                resolvedSiteId: hasPathParams
+                  ? (params.siteId ?? null)
+                  : (context.data?.siteId ?? null),
+                resolvedOrgId: hasPathParams
+                  ? (params.organizationId ?? params.spaceCatId ?? null)
+                  : (context.data?.organizationId ?? context.data?.spaceCatId ?? null),
+                idSource: hasPathParams ? 'path' : 'body',
+              }, 'RO admin access allowed on owned resource');
+              accessLogged = true;
+            } else {
               log.warn({
                 tag: 'ro-admin',
                 email: authInfo.getProfile?.()?.email,
                 method: context.pathInfo?.method,
                 suffix: context.pathInfo?.suffix,
                 org: authInfo.getTenantIds?.()[0],
+                reason: 'not-owner',
               }, 'Read-only admin blocked from route');
               return forbidden('Forbidden');
             }
-            log.info({
-              tag: 'ro-admin-access',
-              email: authInfo.getProfile?.()?.email,
-              method: context.pathInfo?.method,
-              suffix: context.pathInfo?.suffix,
-              org: authInfo.getTenantIds?.()[0],
-              resolvedSiteId: context.data?.siteId ?? null,
-              resolvedOrgId: context.data?.organizationId ?? context.data?.spaceCatId ?? null,
-              idSource: 'body',
-            }, 'RO admin access allowed on owned resource');
           }
+          // Read action: capability alone permits — no ownership DB lookup needed.
+        } catch (err) {
+          log.error({ tag: 'ro-admin', err }, 'Unexpected error in RO admin authorization; denying access');
+          return forbidden('Forbidden');
+        }
+
+        // Emit audit log only when the richer access log was not already emitted
+        // (i.e. for reads and any path where ownership-based access was not recorded).
+        if (!accessLogged) {
+          log.info({
+            tag: 'ro-admin-audit',
+            email: authInfo.getProfile?.()?.email,
+            method: context.pathInfo?.method,
+            suffix: context.pathInfo?.suffix,
+            org: authInfo.getTenantIds?.()[0],
+          }, 'RO admin accessed route');
         }
       }
-
-      log.info({
-        tag: 'ro-admin-audit',
-        email: authInfo.getProfile?.()?.email,
-        method: context.pathInfo?.method,
-        suffix: context.pathInfo?.suffix,
-        org: authInfo.getTenantIds?.()[0],
-      }, 'RO admin accessed route');
     }
 
     return fn(request, context);
--- a/packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
+++ b/packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
@@ -216,6 +216,15 @@ export function readOnlyAdminWrapper(fn, { routeCapabilities } = {}) {
             // resource ID (no path param and no body siteId/organizationId) fail-closed.
             const isOwner = await isOwnerOfResource(context, authInfo, params);
             if (isOwner) {
+              if (capability === null) {
+                log.warn({
+                  tag: 'ro-admin',
+                  reason: 'unmapped-route-allowed',
+                  method: context.pathInfo?.method,
+                  suffix: context.pathInfo?.suffix,
+                  org: authInfo.getTenantIds?.()[0],
+                }, 'RO admin allowed on unmapped route via ownership — add this route to routeCapabilities');
+              }
               log.info({
                 tag: 'ro-admin-access',
                 email: authInfo.getProfile?.()?.email,
--- a/packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
+++ b/packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
@@ -172,7 +172,11 @@ export function readOnlyAdminWrapper(fn, { routeCapabilities } = {}) {
       }
 
       if (isObject(routeCapabilities)) {
-        const params = extractRouteParams(context);
+        const params = extractRouteParams(context, routeCapabilities);
+        // print params to the log
+        log.info({ tag: 'ro-admin', params }, 'params');
+        // print context.data to the log
+        log.info({ tag: 'ro-admin', data: context.data }, 'data');
         const hasPathParams = Object.keys(params).length > 0;
 
         if (hasPathParams) {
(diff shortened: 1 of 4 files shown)


packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
```diff
@@ -24,6 +28,86 @@ function forbidden(message) {
...
+  if (!dataAccess) {
+    log.error({ tag: 'ro-admin' }, 'isOwnerOfResource: dataAccess not on context — ensure dataAccessWrapper runs before readOnlyAdminWrapper');
+    return false;
+  }
+
+  // Body fallback only when the route carries no path params at all (e.g. POST /preflight/jobs).
+  // When params has keys, the route does have path identifiers; relying on body data in that
+  // case could allow a caller to spoof ownership by naming params differently than :siteId.
+  const hasPathParams = Object.keys(params).length > 0;
+  const siteId = hasPathParams ? params.siteId : context.data?.siteId;
```
solaris007, line 63: **Critical: confused-deputy / IDOR bypass for unmapped path-resource routes.**

The `hasPathParams === false` gate relies on the wrapper's own route lists (`routeCapabilities` + `internalRoutes`) being exhaustive. When a write route lives in NEITHER list but its URL contains a resource segment, `extractRouteParams` returns `{}` (no match anywhere), `hasPathParams` is false, the body fallback consults `context.data`, and the wrapper authorizes against the body-claimed resource while the handler may operate on the URL-claimed resource.

Attack: developer adds `PUT /sites/:siteId/transfer-ownership` and forgets both lists. Mallory (RO admin) owns `mallory-site` and sends `PUT /sites/victim-site/transfer-ownership` with body `{"siteId":"mallory-site"}`. Wrapper authorizes against `mallory-site`. Handler reads its own `context.params.siteId = 'victim-site'` from the framework's URL pattern and mutates the victim site in another org. The drift `log.warn` fires but does NOT block.

Fix (recommended): only allow body fallback when `capability !== null`. Preserves legitimate `POST /preflight/jobs` (mapped, no path params, body siteId IS the resource) while denying ANY unmapped route. Add a test asserting `PUT /sites/<victim>/<unmapped>` with body `{siteId: '<owned>'}` returns 403 and that `Site.findById` is NOT called with the owned body siteId.

Alternative: have `extractRouteParams` return `null` for no-match vs `{}` for matched-with-no-params, and fail-closed in the wrapper on `params === null`.

packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
```diff
@@ -95,31 +202,78 @@ export function readOnlyAdminWrapper(fn, { routeCapabilities } = {}) {
...
+
+          if (action !== 'read' && action !== 'readAll') {
+            // Write or unmapped route: verify ownership of the referenced resource.
+            // Ownership is the governing invariant — a null/absent capability is not an
+            // automatic deny; it means the route is not mapped, but RO admins who own
+            // the resource may still perform the operation. Routes with no resolvable
+            // resource ID (no path param and no body siteId/organizationId) fail-closed.
+            const isOwner = await isOwnerOfResource(context, authInfo, params);
+            if (isOwner) {
+              if (capability === null) {
```
solaris007, line 223: **Important: permissive default for unmapped writes erodes the explicit-allow-list model.**

The "ownership-first" choice (capability=null + isOwner=true ALLOWED with a drift warn) inverts the safe default for a security wrapper. Owning a site is not a generalized authorization claim for every operation defined on that site. RO admin is bounded by `FT_READ_ONLY_ORG` and capability mapping precisely so that adding sensitive operations requires deliberate policy.

A new destructive endpoint added without a `routeCapabilities` entry is callable by any RO admin who owns the target, with only `log.warn` to flag it. The drift signal is reactive; the gate is missing.

Recommended fix (addresses the Critical above + the `internalRoutes` drift item together): when `capability === null`, deny unmapped writes regardless of ownership. The "RO admin can write resources their org owns" property is then expressed by explicitly mapping the route in `routeCapabilities`. `internalRoutes` becomes unnecessary - consumers add new routes to one map only.

If you keep the permissive default, please elevate the drift signal to an alert or fail-closed in non-prod environments.

packages/spacecat-shared-http-utils/src/auth/route-utils.js
```diff
@@ -64,6 +64,45 @@ export function resolveRouteCapability(context, routeMap) {
...
+ * (the full internal route list) so ownership checks can still resolve path params for
+ * routes that are known to the service but not mapped in routeCapabilities.
+ *
+ * @param {Object} context - Universal context with pathInfo
+ * @param {Object<string, string>} routeMap - Route pattern to value map
+ * @param {string[]} [fallbackRoutes=[]] - Additional route patterns to try when routeMap has
+ *   no match (e.g. 'DELETE /sites/:siteId')
+ * @returns {Object<string, string>}
+ */
+export function extractRouteParams(context, routeMap, fallbackRoutes = []) {
```
solaris007, line 82: **Important: `routeCapabilities` non-object input silently fails open.**

The construction-time guards are `if (!routeCapabilities) throw` plus `guardNonEmptyRouteCapabilities`, which only throws when `isObject(...) && length === 0`. A caller passing a truthy non-plain-object - most likely real misconfiguration: `routeCapabilities: ['GET /sites', ...]` - passes both guards. Then in the wrapper body `if (isObject(routeCapabilities))` evaluates false and the entire auth block is skipped, so RO admins bypass all authorization.

Fix: tighten this guard to require a non-empty plain object. Throw on arrays, strings, numbers - anything that is not a plain object with at least one key.

```js
export function guardNonEmptyRouteCapabilities(wrapperName, routeCapabilities) {
  if (!isObject(routeCapabilities) || Object.keys(routeCapabilities).length === 0) {
    throw new Error(`${wrapperName}: routeCapabilities must be a non-empty object`);
  }
}
```

The same guard is used by `s2sAuthWrapper`, so tightening it is defense-in-depth for both wrappers. Add a unit test for `routeCapabilities: []`.

packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
```diff
@@ -24,6 +28,86 @@ function forbidden(message) {
...
+    return false;
+  }
+
+  // Body fallback only when the route carries no path params at all (e.g. POST /preflight/jobs).
+  // When params has keys, the route does have path identifiers; relying on body data in that
+  // case could allow a caller to spoof ownership by naming params differently than :siteId.
+  const hasPathParams = Object.keys(params).length > 0;
+  const siteId = hasPathParams ? params.siteId : context.data?.siteId;
+  const organizationId = hasPathParams
+    ? (params.organizationId ?? params.spaceCatId)
```
solaris007, line 65: **Important: `spaceCatId` alias couples shared utils to spacecat-api-service routing conventions.**

`spaceCatId` is a route-naming convention owned by `spacecat-api-service` (used by `/v2/orgs/:spaceCatId/*`). The shared HTTP-utils package now hardcodes that knowledge in both the path-param branch and the body branch.

A future consumer of this wrapper that uses `:spaceCatId` for a different resource type (or a future API-service refactor that retires the alias) gets silently wrong authorization decisions because this wrapper unconditionally treats `:spaceCatId` as an organization ID.

Fix: make the alias configurable per wrapper instance. Default `{}`; spacecat-api-service supplies the alias at its call site.

```js
// In opts:
paramAliases: { spaceCatId: 'organizationId' }

// In isOwnerOfResource:
const organizationId = hasPathParams
  ? (params.organizationId ?? resolveAlias(params, paramAliases, 'organizationId'))
  : (context.data?.organizationId ?? resolveAlias(context.data, paramAliases, 'organizationId'));
```

Pushes the convention back to the consumer; keeps the wrapper neutral.

packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
```diff
@@ -95,31 +202,78 @@ export function readOnlyAdminWrapper(fn, { routeCapabilities } = {}) {
...
-        // split(':').pop() extracts the action; a missing or malformed value yields
-        // undefined, which is not a read action and correctly blocks the request.
-        const action = capability?.split(':').pop();
-
-        if (action !== 'read' && action !== 'readAll') {
-          log.warn({
-            tag: 'ro-admin',
+        let accessLogged = false;
+        try {
+          const params = extractRouteParams(context, routeCapabilities, internalRoutes);
```
solaris007, line 207: **Important: `internalRoutes` introduces a parallel source of truth with drift risk.**

There are now two collections of route patterns consumers must keep in sync: `routeCapabilities` (capability mapping) and `internalRoutes` (param extraction only). Adding a new `PUT /sites/:siteId` requires the author to remember both. Forgetting `internalRoutes` is a silent denial; forgetting both is the Critical bypass.

Cleanest fix (combined with the Critical fix above): collapse to a single map. Allow `routeCapabilities: { 'PUT /sites/:siteId': null }` to mean "known route, no capability required, deny" or use a sentinel like `'site:owner-only'`. Update `resolveRouteCapability` to distinguish explicit null from "not in map" via `hasOwnProperty` rather than truthy check (route-utils.js:60). Then `extractRouteParams` does not need a `fallbackRoutes` parameter at all - it walks the same map.

Less surface, no drift, and the Critical bypass disappears because every guarded route is enumerated.

packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
```diff
@@ -24,6 +28,86 @@ function forbidden(message) {
...
+ *
+ * Fail-closed: returns false if dataAccess is unavailable, the entity is not found,
+ * no recognizable ID can be resolved, or any lookup throws.
+ *
+ * @param {Object} context - Universal context (must have context.dataAccess and context.log)
+ * @param {Object} authInfo - The AuthInfo instance for the current user
+ * @param {Object<string, string>} params - Named path params extracted from the route pattern
+ * @returns {Promise<boolean>}
+ */
+async function isOwnerOfResource(context, authInfo, params) {
```
solaris007, line 52: **Important: parent-child resource ambiguity for nested path params.**

For a route like `PATCH /sites/:siteId/audits/:auditId`, the wrapper authorizes on `siteId` alone. If `audit-x` actually belongs to `site-b`, but Mallory sends `PATCH /sites/mallory-site/audits/audit-x`, the wrapper:
- extracts `{ siteId: 'mallory-site', auditId: 'audit-x' }`
- looks up `mallory-site` -> Mallory's org -> `hasOrganization` returns true
- allows the request

The handler may or may not verify that `audit-x.siteId === 'mallory-site'`. If it does not, Mallory mutates an audit in another org via her own site's URL prefix.

This is materially different from the body-fallback contract that JSDoc documents ("Handlers behind body-fallback authorization MUST only mutate the declared siteId/organizationId resource"). The same risk applies to path params when multiple resource IDs are present, but it is not documented and not enforced.

Fix: pick one of
- Document explicitly in the wrapper JSDoc: "Handlers behind path-param authorization MUST verify all nested resource IDs belong to the authorized parent." Make this part of the public contract.
- Stop authorizing on the parent alone when multiple resource IDs are present - require an explicit child-level check (heavier but removes the implicit-handler-contract footgun).

packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
```diff
@@ -35,6 +119,7 @@ function forbidden(message) {
  * @returns {Promise<boolean>}
  */
 async function evaluateFeatureFlag(context, authInfo) {
+  const { log } = context;
```
solaris007, line 122: **Important: feature flag scoped to `getTenantIds()[0]` only.**

For multi-org users, only the first tenant ID is checked. Behavior is order-dependent: a user in orgs `[A, B]` is treated differently than the same user in `[B, A]`. If `FT_READ_ONLY_ORG` is enabled for A but not B, the user in `[A, B]` proceeds past the feature flag, and can then access B-owned resources via `hasOrganization(B)` returning true.

Effectively: an RO admin in an enrolled org can act on resources in a non-enrolled org they also belong to.

Decide and document the semantics:
- (a) "user-org gate": check the user's primary tenant (current behavior - defensible but should be explicit).
- (b) "resource-org gate": check the flag against the resource's owning IMS org AFTER the ownership lookup (closer to per-tenant rollout semantics).

Also: `tenantIds[0]` reflects whatever order the JWT issuer puts tenants in. Document the dependency, or evaluate the flag for all tenants the user has (any-true / all-true depending on intent).

Add a test pinning the multi-tenant ordering behavior so the choice does not silently flip later.

solaris007 (changes requested):
Hey @ravverma,

Thanks for the diligent self-review iterations and the depth of test coverage. The fail-closed posture, the explicit spoofing test for `:id`-named routes, the drift-detection log, and the read fast-path all show careful security thinking. That said, this first formal review surfaces one Critical bypass path and a cluster of Important issues around the design surface. The Critical issue is in scope to fix on this branch with a small change.

## Strengths

- Fail-closed exhaustiveness is consistent across every error path I traced: missing `dataAccess`, missing `Site`/`Organization` accessor, null entity, null `imsOrgId`, lookup throws, FF eval throws, and unexpected exceptions in the wrapper body (read-only-admin-wrapper.js:53-100, 137-141, 256-267).
- The `hasPathParams === false` gate closes the param-name spoofing attack for routes that ARE in `routeCapabilities` with non-`:siteId` param names. The test at read-only-admin-wrapper.test.js:786 is the right test to have.
- The `imsOrgId == null` guard before `hasOrganization(...)` prevents the TypeError that `auth-info.js` would throw on `undefined.split('@')`.
- Read fast-path skips the DB lookup for `read`/`readAll` capabilities.
- Drift-detection `unmapped-route-allowed` warn gives operators an actionable signal (though see Important item 1 for why detection alone is not enough here).
- The `accessLogged` flag cleanly prevents double audit-log emission.
- Wrapper-ordering requirements (`dataAccessWrapper` and body-parser before this wrapper) are explicit in JSDoc rather than left as folklore.
- 538 lines of tests with strong negative-case coverage, including DB throws on both site and org paths and the `getOrganization()` throw path.

## Issues

### Critical (Must Fix)

**Body-fallback bypass for completely unmapped path-resource routes** (read-only-admin-wrapper.js:63-66)

The `hasPathParams === false` gate depends on the wrapper's own route lists (`routeCapabilities` + `internalRoutes`) being exhaustive. When a write route lives in NEITHER list but its URL contains a resource segment, `extractRouteParams` returns `{}` (no match anywhere), `hasPathParams` is false, the body fallback consults `context.data`, and the wrapper authorizes against the body-claimed resource while the handler may operate on the URL-claimed resource. This is a confused-deputy / IDOR pattern.

Attack scenario:
1. A developer adds `PUT /sites/:siteId/transfer-ownership` to spacecat-api-service. They forget to add it to either `routeCapabilities` or `internalRoutes`.
2. Mallory (RO admin) owns site `mallory-site`. She wants to attack `victim-site` in another IMS org.
3. Mallory sends `PUT /sites/victim-site/transfer-ownership` with body `{"siteId":"mallory-site"}`.
4. `extractRouteParams` returns `{}` (route in neither list); `hasPathParams = false`.
5. Body fallback: `siteId = 'mallory-site'`. `Site.findById('mallory-site')` resolves to Mallory's org. `hasOrganization(...)` returns true.
6. `capability === null` -> drift `log.warn` fires; `ro-admin-access` log records `idSource: 'body'`, `resolvedSiteId: 'mallory-site'`.
7. Wrapper calls `fn(request, context)`. The handler reads its own `context.params.siteId` from the URL = `'victim-site'` and mutates the victim site.

The wrapper authorized against the body claim; the handler operated on the URL claim. The drift `log.warn` fires but does NOT block - logs are detection, not control.

The existing test at read-only-admin-wrapper.test.js:729 ("allows access on unmapped route when siteId is supplied via body") uses suffix `/jobs/preflight` - no resource segment, no victim. The `PUT /sites/:siteId/<unmapped>` + body-siteId combination is not tested.

Fix (recommended): only allow body fallback when `capability !== null`. This preserves the legitimate `POST /preflight/jobs` case (mapped capability, no path params, body siteId is the resource) while denying ANY unmapped route, with or without resource segments. Add a test asserting `PUT /sites/<victim-id>/<unmapped-action>` with body `{siteId: '<owned-id>'}` returns 403 and that `Site.findById` is NOT called with the body siteId.

Alternative fixes: distinguish "no match" from "matched-with-no-params" in `extractRouteParams` (e.g. return `null` vs `{}`), or introduce an explicit `bodyFallbackRoutes` allowlist.

### Important (Should Fix)

**1. Permissive default for unmapped writes erodes the explicit-allow-list model** (read-only-admin-wrapper.js:223)

The "ownership-first" choice (capability=null + isOwner=true ALLOWED, with a drift warn) inverts the safe default for a security wrapper. Owning a site is not a generalized authorization claim for every operation defined on that site. RO admin is bounded by `FT_READ_ONLY_ORG` and capability mapping precisely so that adding sensitive operations requires deliberate policy. A new destructive endpoint added without a `routeCapabilities` entry is callable by any RO admin who owns the target, with only a `log.warn` to flag it.

Recommended fix (also addresses the Critical and Important 4): when `capability === null`, deny unmapped writes regardless of ownership. The "RO admin can write resources their org owns" property is then expressed by explicitly mapping the route in `routeCapabilities`. `internalRoutes` becomes unnecessary - consumers add new routes to one map only.

If you keep the permissive default, please elevate the drift signal from `log.warn` to an alert or fail-closed in non-prod.

**2. `routeCapabilities` non-object input silently fails open** (route-utils.js:115, read-only-admin-wrapper.js:204)

`guardNonEmptyRouteCapabilities` only throws when `isObject(...) && length === 0`. A caller passing a truthy non-plain-object - most likely real misconfiguration: `routeCapabilities: ['GET /sites', ...]` - passes both this guard and the `if (!routeCapabilities) throw`. In the wrapper body, `if (isObject(routeCapabilities))` evaluates false and the entire auth block is skipped. RO admins bypass all authorization.

Fix: tighten the guard to require a non-empty plain object. Throw on arrays, strings, numbers. The same shape exists in `s2sAuthWrapper`, so this improves both wrappers. Add a unit test for `routeCapabilities: []`.

**3. `spaceCatId` alias couples shared utils to spacecat-api-service routing conventions** (read-only-admin-wrapper.js:65-66)

`spaceCatId` is a route-naming convention owned by `spacecat-api-service` (`/v2/orgs/:spaceCatId/*`). The shared HTTP-utils package now hardcodes that knowledge. A future consumer that uses `:spaceCatId` for a different resource - or a future API-service refactor that retires the alias - gets silently wrong authorization decisions because this wrapper unconditionally treats `:spaceCatId` as an organization ID.

Fix: make the alias configurable per wrapper instance, e.g. `opts.paramAliases: { spaceCatId: 'organizationId' }`. Default `{}`; spacecat-api-service supplies the alias at its call site.

**4. `internalRoutes` introduces a parallel source of truth with drift risk** (read-only-admin-wrapper.js:207)

Consumers must keep `routeCapabilities` (capability mapping) and `internalRoutes` (param extraction only) in sync. Forgetting `internalRoutes` is a silent denial; forgetting both is the Critical bypass above.

Cleanest fix (combined with the Critical fix): collapse to a single map. Allow `routeCapabilities: { 'PUT /sites/:siteId': null }` to mean "known route, no capability, deny" or `'site:owner-only'` to mean "ownership gates". Update `resolveRouteCapability` to distinguish explicit null from "not in map" via `hasOwnProperty`. Then `extractRouteParams` doesn't need a `fallbackRoutes` parameter at all.

**5. Parent-child resource ambiguity for nested path params** (read-only-admin-wrapper.js:52-100)

For `PATCH /sites/:siteId/audits/:auditId`, the wrapper authorizes on `siteId` alone. If `audit-x` belongs to `site-b` but Mallory sends `PATCH /sites/mallory-site/audits/audit-x`, the wrapper allows because Mallory owns `mallory-site`. If the handler doesn't verify `audit-x.siteId === 'mallory-site'`, Mallory mutates an audit in another org via her own site's URL prefix.

Same risk class as the body-fallback contract documented in JSDoc, but not documented for nested path params.

Fix: document the parent-child path-param contract ("Handlers behind path-param authorization MUST verify all nested resource IDs belong to the authorized parent"), or stop authorizing on the parent alone when multiple resource IDs are present.

**6. Feature flag scoped to `getTenantIds()[0]` only** (read-only-admin-wrapper.js:129-130)

For multi-org users, only the first tenant ID is checked. A user in orgs `[A, B]` where `FT_READ_ONLY_ORG` is enabled for A but not B proceeds past the feature flag, then can access B-owned resources via `hasOrganization(B)` returning true. Effectively: an RO admin in an enrolled org can act on resources in a non-enrolled org they also belong to.

Decide and document the semantics: is the flag (a) "user-org gate" (check primary tenant, current behavior) or (b) "resource-org gate" (check after ownership lookup against the target org)? Add a test pinning the multi-tenant ordering behavior.

### Minor (Nice to Have)

**1. Silent failure when `dataAccess.Site` / `dataAccess.Organization` accessor is missing** (read-only-admin-wrapper.js:73, 87)

Optional chaining returns `undefined` when `dataAccess.Site` is absent; `!site` then returns false with no log. The behavior is fail-closed (good), but operationally indistinguishable from "site not found". Add `log.error({ tag: 'ro-admin' }, 'dataAccess.Site accessor missing')` before returning false. The test at line 791 asserts 403 but no log.

**2. JSDoc does not document siteId-over-organizationId precedence** (read-only-admin-wrapper.js:22-50)

`isOwnerOfResource` checks `siteId` first; if a route surfaces both, organizationId is dead code. Add to JSDoc: "When both `siteId` and `organizationId` are resolvable, `siteId` is preferred."

**3. `err` logged with full stack in production paths** (read-only-admin-wrapper.js:96, 137, 256)

`log.error({ err })` typically serializes `err.message` + `err.stack`. DB driver stack frames can include table names, region, and connection identifiers. Consider passing `{ message: err.message, name: err.name }` at error level and gating full error to debug, or confirm the central log pipeline strips stacks.

**4. Two coverage refinements for the new logging contract** (read-only-admin-wrapper.test.js)

The body-fallback access-log test asserts the access log is emitted but doesn't assert the audit log is NOT emitted (the path-based equivalent does both). Add the suppression assertion. Separately, `resolvedOrgId` is logged but never asserted in any access-log test.

**5. `organizationId` precedence over `spaceCatId` not pinned by test**

`params.organizationId ?? params.spaceCatId` and the body equivalent: `??` means `organizationId` wins when both are set. Add `context.data = { organizationId: 'org-A', spaceCatId: 'org-B' }` -> assert `findById('org-A')` is called.

## Recommendations

- File a follow-up to extract `isOwner` as a pluggable strategy. The wrapper hardcodes two ownership chains; extending to a third resource type (project, brand, opportunity, scrape job) means editing this shared library every time. An `opts.isOwner: async (context, authInfo, params) => boolean` shape lets consumers contribute resolvers without touching the security choke point.
- Cross-domain consistency with `s2sAuthWrapper`: both consume `routeCapabilities`. After this PR, RO admin allows unmapped routes by ownership while s2s denies. Intentional given different threat models, but add a one-liner in each JSDoc pointing at the sibling wrapper so future readers see them as a pair with deliberately different policies.
- Performance note in JSDoc: every RO admin write adds a DB round-trip (`Site.findById` + `getOrganization()` is two; `Organization.findById` is one). If `Site` carries `imsOrgId` directly, inline it: `site.getImsOrgId()` skips the org round-trip.
- Consider a startup-time invariant check: cross-reference routes mounted under this wrapper against `routeCapabilities`. Turns "consumer forgot to enumerate" from a runtime exposure into a deploy-time failure.
- The audit log semantic change (was `ro-admin-audit` on every allowed; now split between `ro-admin-access` for owned-writes and `ro-admin-audit` for reads) is a quiet contract break for downstream dashboards. Worth a callout in the PR description and a notification to dashboard owners.

- Pluggable `isOwner` strategy for new resource types.
- Mount-time invariant check of routes vs. routeCapabilities.
- Resource-org vs user-org feature flag scoping (related to Important item 6).
- Configurable `paramAliases` (related to Important item 3).

## Assessment

Reasoning: The mapped-route happy paths are well-engineered and the fail-closed posture is consistently applied. The Critical bypass is a confused-deputy pattern reachable when a consumer adds a new resource-path route and forgets either route list - a security wrapper should fail closed on this class of mistake, not rely on a post-hoc log. The cleanest fix (which also addresses Important items 1 and 4 in one change): deny when `capability === null` regardless of ownership, and let `internalRoutes` go away. The other Important items are contract clarifications and config-robustness improvements that should land before this becomes load-bearing across more consumers.

## Next Steps

1. Address the Critical bypass and Important items 1 and 4 together via the recommended single change (deny on capability=null; collapse internalRoutes into routeCapabilities).
2. Tighten the routeCapabilities config-shape guard (Important item 2).
3. Decide and document the feature-flag tenant scoping semantics (Important item 6).
4. Document or enforce the parent-child path-param contract (Important item 5).
5. Make `spaceCatId` alias configurable (Important item 3).
6. Minor items can land in a polish commit.

PR #1596, opened 2026-05-08
feat: add rumConfig to Site config for deterministic RUM availability signal

Sugandhgyl:
Adds rumConfig sub-config to the Site config object with hasDomainKey and lastCheckedAt fields, following the existing fetchConfig/onboardConfig pattern. Includes Joi schema, getRumConfig/hasRumDomainKey/updateRumConfig accessors, toDynamoItem serialization, and full unit test coverage.

## Related Issues

Thanks for contributing!

### DIFF
--- a/packages/spacecat-shared-data-access/README.md
+++ b/packages/spacecat-shared-data-access/README.md
@@ -172,6 +172,25 @@ The `deliveryConfig` object on a Site stores delivery infrastructure details. It
 | `preferContentApi` | boolean | Whether to prefer the Content API for content retrieval |
 | `contentSourcePath` | string | AEM content root path for a site. Used to disambiguate multiple sites that share the same Cloud Manager program and environment. Corresponds to `/content/<site-name>` in the AEM repository. |
 
+## Site rumConfig
+
+The `rumConfig` object on a Site tracks Real User Monitoring (RUM) domain key availability. It is set automatically on site creation and refreshed weekly by the `rum-config-refresh` audit handler.
+
+| Property | Type | Description |
+|----------|------|-------------|
+| `hasDomainKey` | boolean | Whether the site's domain has an active RUM domain key registered |
+| `lastCheckedAt` | string (ISO 8601) | Timestamp of the most recent RUM domain key check |
+
+### Config model methods
+
+| Method | Returns | Description |
+|--------|---------|-------------|
+| `getRumConfig()` | `{ hasDomainKey, lastCheckedAt } \| undefined` | Returns the current rumConfig, or `undefined` if not yet set |
+| `hasRumDomainKey()` | `boolean` | Returns `true` if a RUM domain key is active, `false` otherwise |
+| `updateRumConfig(hasDomainKey)` | `void` | Sets `hasDomainKey` and updates `lastCheckedAt` to the current time |
+
+Sites created before the `rum-config-refresh` handler ran will have no `rumConfig` key — callers must treat `undefined` as "unknown" rather than "not set up".
+
 ## Architecture
 
 ```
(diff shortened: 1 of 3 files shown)


sandsinh (changes requested):
Hey @Sugandhgyl,

## Strengths
- The `rumConfig` sub-config is cleanly isolated within the existing `Config` factory pattern, avoiding any changes to the Site model's top-level interface.
- Joi schema validation covers the shape of the sub-config, catching malformed data at write time rather than silently storing garbage.
- The test suite covers `getRumConfig`, `hasRumDomainKey`, `updateRumConfig`, Joi validation paths, and `toDynamoItem` serialization - good coverage for the contract surface.
- Making `rumConfig` optional in the schema is the right default for backward compatibility with existing Site records that do not yet have this field.

## Issues

### Critical (Must Fix)
None.

### Important (Should Fix)

**1. `updateRumConfig` timestamps cannot be controlled in tests - packages/spacecat-shared-data-access/src/models/site/config.js**
`updateRumConfig` calls `new Date().toISOString()` directly. Any test that needs to assert on `lastCheckedAt` must either use loose assertions or freeze the clock globally. This is the kind of coupling that causes intermittent CI failures when two `Date` calls straddle a millisecond boundary. Add a `now` parameter (defaulting to `new Date()`) so callers and tests can inject the timestamp without system-wide clock manipulation.

**2. `updateRumConfig` accepts any value for `hasDomainKey` with no runtime guard - packages/spacecat-shared-data-access/src/models/site/config.js**
If a caller passes `undefined`, `null`, or a non-boolean, the Joi schema will only catch the error at `toDynamoItem()` time - after the mutation has already been applied to in-memory state. Add an early `typeof hasDomainKey !== 'boolean'` guard so the error surfaces at the mutation site, not later at serialization.

**3. Public methods have no JSDoc - packages/spacecat-shared-data-access/src/models/site/config.js**
`getRumConfig`, `hasRumDomainKey`, and `updateRumConfig` are exported as part of the public API of a shared library consumed by multiple services. Consumers must read the implementation to understand the contract. Add minimal JSDoc (param types, return type, what the method represents) so the interface is self-documenting.

**4. Time-dependent assertion in the test is a CI reliability risk - packages/spacecat-shared-data-access/test/unit/models/site/config.test.js**
The test verifying that `updateRumConfig` sets `lastCheckedAt` is likely capturing `new Date().toISOString()` at assertion time and comparing it to the timestamp written inside the mutator. These two calls can produce different values under CI load. Use `sinon.useFakeTimers()` or inject a fixed timestamp via the recommended `now` parameter.

### Minor (Nice to Have)

- `getRumConfig` returns a direct reference to internal state. If a caller mutates the returned object without going through `updateRumConfig`, the mutation bypasses the `lastCheckedAt` update. Consider returning a frozen copy or a plain object copy.

## Recommendations

## Assessment
Ready to merge? With fixes
The Important findings - clock injection, shared staleness constant, and input validation - are small changes but this is a shared library consumed by multiple services. A loose contract here propagates outward and is harder to fix after the downstream packages ship.

sandsinh (dismissed):
Hey @Sugandhgyl,

## Strengths

All four Important and one Minor finding from the previous review have been addressed in commit f7f94715:
- Previously flagged: injectable `now` parameter - getRumConfig now returns a shallow copy, updateRumConfig accepts injectable timestamp, TypeError guard added early on non-boolean input, JSDoc on all three public methods, and test timestamps are fully deterministic via injected values.

The implementation is clean and the test suite is comprehensive: isolated copies verified, TypeError paths covered, Joi schema paths exercised, and toDynamoItem serialization confirmed.

## Assessment
Ready to merge? Yes
All concerns from the previous review are resolved and no new issues introduced.

sandsinh (approved):

PR #1598, opened 2026-05-11
feat: add path-level prerender deploy support to tokowaka-client

ssilare-adobe:
### Path-level prerender suggestions
- Adds path-level prerender suggestion support to the Tokowaka client alongside existing domain-wide support
- `deployToEdge` classifies suggestions into pattern-based (domain-wide + path-level) and per-URL buckets, then deploys each via the appropriate code path
- Pattern suggestions are deployed by appending their `allowedRegexPatterns` to `metaconfig.prerender.allowList` (deduplicated, CDN write skipped when nothing changes)
- Per-URL suggestions in the same batch that are covered by a pattern are automatically skipped and marked with `coveredByDomainWide` or `coveredByPattern`

### Rollback
- `rollbackSuggestions` handles all rollback paths uniformly:
  - **Per-URL**: removes S3 patches, strips `edgeDeployed`/`tokowakaDeployed`
  - **Pattern (domain-wide + path-level)**: removes pattern from `metaconfig.prerender.allowList` (single in-memory pass, one upload), strips markers
  - **Covered cleanup**: clears `coveredByDomainWide` or `coveredByPattern` on per-URL suggestions that were shadowed by the rolled-back pattern

### Save strategy
- **Per-URL suggestions**: `saveMany()` batch upsert (deploy, rollback, same-batch skipped)
- **Pattern / domain-wide suggestions**: `Promise.allSettled(.save())` — individual saves in parallel (deploy + rollback)
- **Covered suggestions**: `Promise.allSettled(.save())` — individual saves in parallel (deploy marking + rollback cleanup)

### `updatedBy` audit trail
Context-specific fallback strings for traceability:
- Direct suggestions: `'tokowaka-rollback'`
- Per-URL covered by domain-wide: `'domain-wide-rollback'`
- Per-URL covered by path: `'path-rollback'`
- Caller-provided actor email (`updatedBy`) overrides all fallbacks

### Refactors
- **`metaconfig-utils.js`** (new): Extracted `removePatternFromMetaconfig` and `addPatternsToMetaconfig` — pure utility functions for metaconfig allowList manipulation
- **`suggestion-utils.js`**: Extracted `saveSuggestions`, `stripSuggestion`, `cleanupCoveredSuggestions`, `classifySuggestions`, `filterBatchCoveredSuggestions` out of the main class for testability and reuse
- **`pattern-utils.js`**: `buildUrlMatcher` for regex-based URL matching used in both deploy and rollback

## Deploy Flow (`deployToEdge`)

```mermaid
flowchart TD
    A[deployToEdge] --> B[Step 1: Classify suggestions]
    B --> C{Pattern or Per-URL?}

    C -->|Per-URL| D[Step 2: Filter batch-covered]
    D --> E[Step 3: Deploy via S3 patches]
    E --> F["Save via saveMany()"]

    C -->|Pattern / Domain-Wide| G[Step 4: Fetch metaconfig]
    G --> H["For each pattern suggestion (sequential)"]
    H --> I[addPatternsToMetaconfig]
    I --> J{allowList changed?}
    J -->|Yes| K[Upload metaconfig to S3]
    J -->|No| L[Skip CDN write]
    K --> M[Stamp edgeDeployed on suggestion]
    L --> M
    M --> N[Mark covered per-URL suggestions]
    N --> O["Save covered via Promise.allSettled(.save())"]
    O --> H

    H -->|All done| P["Batch-save pattern suggestions via Promise.allSettled(.save())"]

    D --> Q[Step 5: Same-batch skipped]
    Q --> R["Mark coveredByDomainWide / coveredByPattern"]
    R --> S["Save via saveMany()"]

    F --> T[Return results]
    P --> T
    S --> T
```

## Rollback Flow (`rollbackSuggestions`)

```mermaid
flowchart TD
    A[rollbackSuggestions] --> B[Separate pattern vs per-URL]

    B -->|Per-URL| C[Filter eligible suggestions]
    C --> D[Remove S3 patches per URL]
    D --> E["Strip edgeDeployed / tokowakaDeployed"]
    E --> F["Save via saveMany()"]

    B -->|Pattern / Domain-Wide| G[Fetch metaconfig once]
    G --> H["For each pattern: removePatternFromMetaconfig (in-memory)"]
    H --> I{metaconfig changed?}
    I -->|Yes| J[Upload metaconfig once]
    I -->|No| K[Skip upload]
    J --> L["Save all pattern suggestions via Promise.allSettled(.save())"]
    K --> L

    L --> M[For each succeeded pattern suggestion]
    M --> N{isDomainWide?}
    N -->|Yes| O["Find covered: coveredByDomainWide = suggestion.id"]
    N -->|No / Path| P["Find covered: coveredByPattern = suggestion.id"]
    O --> Q["cleanupCoveredSuggestions via Promise.allSettled(.save())"]
    P --> Q
    Q -->|Strip coveredByDomainWide| R[Done]
    Q -->|Strip coveredByPattern + edgeDeployed| R

    F --> S[Return results]
    R --> S
```

## Save Strategy

| Suggestion Type | Context | Save Method |
|---|---|---|
| Per-URL | Deploy / Rollback | `saveMany()` (batch upsert) |
| Pattern / Domain-Wide | Deploy / Rollback | `Promise.allSettled(.save())` |
| Covered by pattern | Deploy marking | `Promise.allSettled(.save())` |
| Covered by pattern | Rollback cleanup | `Promise.allSettled(.save())` |
| Same-batch skipped | Deploy | `saveMany()` |

## Test plan

- [x] All 736 unit tests pass with 100% coverage
- [x] `deployToEdge` path-level tests: pattern written to allowList, covered per-URL skipped, markers set
- [x] `rollbackSuggestions` per-URL tests: S3 patches removed, markers cleared
- [x] `rollbackSuggestions` pattern tests: allowList updated, covered suggestions cleaned, correct `updatedBy` fallbacks applied per context
- [x] Pattern suggestion save failure surfaced via Promise.allSettled
- [x] Covered suggestion save failure logged and swallowed gracefully


packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -47,6 +47,11 @@ function isEdgeDeployableSuggestionStatus(status) {
   return status === 'NEW' || status === 'PENDING_VALIDATION';
 }
 
+/** Returns true if the suggestion is a path-level prerender suggestion. */
+function isPathSuggestion(suggestion) {
```
nit23uec: lets use `coveredByPattern` prop in the schema for the segment related suggestions.
isCoveredByPathPattern should be a generic method that returns true if `suggestion.getData()?.coveredByPattern || suggestion.getData()?.isDomainWide`

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -817,33 +822,46 @@ class TokowakaClient {
       );
     }
 
-    // Validate which suggestions can be rolled back
+    // Separate metaconfig-based suggestions (path and domain-wide allowList entries)
+    // from per-URL suggestions — each group is rolled back via a different code path.
+    const metaconfigSuggestions = suggestions.filter((s) => {
```
nit23uec: nitpick: metaconfig is an impl detail.
lets use `patternCoveredSuggestions` to filter segment and domain wide suggestions

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -917,15 +935,64 @@ class TokowakaClient {
...
+      if (!patternToRemove) {
+        this.log.warn(`[edge-rollback] Suggestion ${suggestion.getId()} has no allowedRegexPatterns, skipping`);
+        ineligibleSuggestions.push({ suggestion, reason: 'Missing allowedRegexPatterns' });
+        // eslint-disable-next-line no-continue
+        continue;
+      }
+
+      try {
+        // eslint-disable-next-line no-await-in-loop
+        const metaconfig = await this.fetchMetaconfig(baseURL);
```
nit23uec: earlier also, we used to support a domain wide rollback - can you pls make sure we are usig the same code to rollback such suggestions. Pls sync with @dipratap

packages/spacecat-shared-tokowaka-client/src/index.js
nit23uec: better to reuse the generic method `coveredByPattern` proposed above to do any handling of such suggestions.

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1301,6 +1369,11 @@ class TokowakaClient {
         if (Array.isArray(allowedRegexPatterns) && allowedRegexPatterns.length > 0) {
           domainWideSuggestions.push({ suggestion, allowedRegexPatterns });
         }
+      } else if (isPathSuggestion(suggestion)) {
```
nit23uec: domainwide and pathbased should be handled exactly on a similar basis. there should be no separate if else required for them.
we did a mistake when we introduced domainWide prop for the `/*` suggestions - we should have just introduced a generic prop coveredByPattern to cover any regex case.

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: @nit23uec This checks if the suggestion is a path level suggestion and not if the suggestion is covered by path level suggestion. 

What this function checks: 
`https://adobe.com/products/* -> Path level suggestion -> returns true`


If we need to add `coveredByPattern`, it should come on the suggestions which are covered by pattern like `https://adobe.com/products/test` rather than on path level suggestion.

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: Done — renamed to `patternSuggestions` throughout (`metaconfigSuggestions` → `patternSuggestions`, `succeededMetaconfigSuggestions` → `succeededPatternSuggestions`, etc.).

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: The rollback path now uses the same unified `patternSuggestions` loop for both domain-wide and path-level suggestions — `isPatternSuggestion` detects both via `allowedRegexPatterns` presence (plus `isDomainWide: true` as a fallback for malformed suggestions). The rollback removes only the specific pattern from the allowList and deletes the `prerender` key when the list empties, same as before for domain-wide.

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: Done — the deploy loop is now fully unified. Both domain-wide (`isDomainWide: true, allowedRegexPatterns: ['/\*']`) and path-level (`allowedRegexPatterns: ['/products/\*']`) suggestions go through the same loop body. The only remaining `isDomainWide` check is for the coverage field name: `coveredByDomainWide` vs `coveredByPattern`. I've kept `coveredByDomainWide` for domain-wide suggestions for backward compatibility with existing consumers (tokowaka-utilities, spacecat-api-service). Path-level suggestions now use `coveredByPattern`. If you'd like to collapse both into `coveredByPattern`, that would require coordinated changes in those consumers — happy to do that as a follow-up.

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: To clarify the distinction: `isPatternSuggestion` is used to identify suggestions that ARE a pattern (domain-wide or path-level) — i.e., they deploy via metaconfig allowList rather than per-URL S3 config. The `coveredByPattern` and `coveredByDomainWide` fields are written on per-URL suggestions that are COVERED by a pattern suggestion. The updated implementation uses `coveredByPattern` for path-level coverage and keeps `coveredByDomainWide` for domain-wide coverage (backward compat). If you'd like a single unified field for both, I can do that in a follow-up once consumers are updated.

ssilare-adobe (commented):
### Self-review 

Hey @ssilare-adobe,

## Strengths

- Clean decomposition of monolithic methods into focused private helpers (`#rollbackPerUrlConfig`, `#stripAndSaveSuggestion`, `#classifySuggestions`, `#filterBatchCoveredSuggestions`, `#deployPerUrlSuggestions`, `#deployPatternSuggestion`). The main `rollbackSuggestions` and `deployToEdge` methods are significantly more readable now.
- Independent coverage layers (`coveredByDomainWide` vs `coveredByPattern`) with field-specific stripping during rollback is architecturally sound. DW rollback and path rollback are truly independent operations, well-tested with dedicated edge cases.
- Deploy now uses append-and-deduplicate (`[...new Set([...existing, ...new])]`) instead of overwriting the entire allowList. This is both a correctness fix and a defense against accidental data loss from concurrent deploys.
- Two-pass rollback for metaconfig (collect removals in memory, upload once) is the right batching strategy, avoiding N S3 writes for N patterns. Deploy-side mirrors this with a single metaconfig fetch shared across all pattern suggestions in the batch.
- Per-suggestion error isolation in rollback ensures one failure does not contaminate the succeeded/failed lists. `#cleanupCoveredSuggestions` uses `Promise.allSettled` so a single covered-suggestion save failure does not abort the rest.
- Same-batch skip logic correctly uses `coveredByDomainWide` or `coveredByPattern` based on the covering pattern type, preserving the independent coverage layer model.
- Sibling keys on `metaconfig.prerender` are preserved via spread in both deploy and rollback paths.
- Backward-compatible API change on `rollbackSuggestions` with defaults on the new options parameter.
- Thorough test coverage with ~1200 new test lines covering rollback independence, double-coverage preservation, error isolation, missing metaconfig, upload failures, invalid patterns, temporal ordering, cascade-skip scenarios, and same-batch coverage field correctness for both DW and path patterns.

## Issues

No blocking issues found.

### Minor (Nice to Have)

**1. ReDoS risk in regex fallback path (pre-existing)**
`src/index.js:91` - The `new RegExp(pattern)` fallback compiles user-provided patterns without ReDoS protection. This is pre-existing on main. The `/*` fast path mitigates the most common patterns. Consider a follow-up to add safe-regex validation or restrict patterns to the `/<path>/*` format.

## Assessment
Ready to merge? Yes.
Well-structured, well-tested change with comprehensive error handling and good architectural decisions around independent coverage layers.

MysticatBot-Dev (commented):
Hey @ssilare-adobe,

## Strengths

- Clean decomposition of monolithic `rollbackSuggestions` and `deployToEdge` into focused private helpers (`#rollbackPerUrlConfig`, `#stripAndSaveSuggestion`, `#classifySuggestions`, `#filterBatchCoveredSuggestions`, `#deployPerUrlSuggestions`, `#deployPatternSuggestion`). The main orchestrators now read as step-by-step narratives.
- Independent coverage layers (`coveredByDomainWide` / `coveredByPattern`) with layer-specific stripping during rollback is architecturally sound. Rolling back one type does not cascade into the other - well-tested with dedicated double-coverage and independence scenarios.
- Append-and-deduplicate deploy (`[...new Set([...existing, ...new])]`) replaces the old full-overwrite approach, eliminating a class of concurrent-deploy data-loss bugs. This is the right design for a shared CDN config.
- Two-pass rollback for metaconfig (collect removals in memory, upload once) avoids N S3 writes for N patterns. Efficient and correct.
- Error isolation via `Promise.allSettled` in `#cleanupCoveredSuggestions` ensures one failed save does not block other covered suggestion cleanups. Per-suggestion try/catch in the rollback loop gives the same guarantee at the pattern level.
- Backward-compatible API change on `rollbackSuggestions` with destructured options parameter and defaults preserves all existing call sites.
- `buildUrlMatcher` fast-path for `/*` patterns avoids regex compilation for the common case and eliminates ReDoS for that path.
- Comprehensive test coverage (~1200 new test lines) covering rollback independence, double-coverage preservation, error isolation, missing metaconfig, upload failures, temporal ordering, cascade-skip scenarios, and same-batch coverage field correctness.

## Issues

### Important (Should Fix)

**1. Rollback only removes `allowedRegexPatterns[0]`, silently dropping additional patterns**

`packages/spacecat-shared-tokowaka-client/src/index.js` - rollback pattern loop:
```js
const patternToRemove = data?.allowedRegexPatterns?.[0];
```

The deploy path correctly handles all entries in `allowedRegexPatterns` (iterates the full array for matchers and merges all into the allowList via `allowedRegexPatterns.flatMap(...)`). The rollback path takes only `[0]`. If a suggestion ever has multiple patterns (e.g., `['/products/*', '/services/*']`), rollback removes only the first - the rest remain orphaned in the CDN metaconfig allowList with no owning suggestion.

**Why it matters**: Deploy and rollback must be symmetric. An orphaned pattern means the CDN continues prerendering URLs the user intended to roll back, with no UI indication the rollback was partial. Even if current suggestions always have single-element arrays, the code neither validates nor documents that assumption.

**How to fix**: Iterate all patterns:
```js
const patternsToRemove = data?.allowedRegexPatterns ?? [];
if (patternsToRemove.length === 0) {
  this.log.warn(`[edge-rollback] Suggestion ${suggestion.getId()} has no allowedRegexPatterns, skipping`);
  ineligibleSuggestions.push({ suggestion, reason: 'Missing allowedRegexPatterns' });
  continue;
}
for (const p of patternsToRemove) {
  if (this.#removePatternFromMetaconfig(metaconfig, p)) {
    metaconfigChanged = true;
  }
}
```

### Minor (Nice to Have)

**2. Regex fallback matches against full URL string, not pathname**

`packages/spacecat-shared-tokowaka-client/src/index.js` - `buildUrlMatcher`:
```js
const regex = new RegExp(pattern);
return (url) => regex.test(url);
```

The `/*` fast path correctly extracts and matches against `new URL(url).pathname`. The regex fallback tests against the entire URL string (e.g., `https://example.com/products/item`). This means a pattern like `^/blog` would never match because the string starts with `https://`. The inconsistency between the two paths could surprise callers. Consider matching against `pathname` in both paths for consistency, or adding a JSDoc note documenting the contract.

**3. `getData()` re-read after `setData()` in rollback depends on entity implementation details**

In the rollback two-pass loop, Pass 1 calls `suggestion.setData(omitKeys(data, ['edgeDeployed', 'tokowakaDeployed']))`, then Pass 2 reads `suggestion.getData()?.isDomainWide`. This depends on `getData()` returning the value previously set. A safer approach would be to capture `isDomainWide` before `setData()`:

```js
toSave.push({ suggestion, isDomainWide: data?.isDomainWide === true });
```

**4. Two CDN invalidation error tests removed without clear reason**

Tests "should return error object if no CDN client available" and "should return error object if CDN invalidation fails" appear deleted from the `invalidateCdnCache` describe block. These don't seem related to the path-level feature. If they were moved elsewhere in the file, ignore this - but if they were accidentally dropped during the test reorganization, they should be restored.

## Recommendations

- Consider adding a test for rollback of a suggestion with `allowedRegexPatterns: ['/a/*', '/b/*']` (multiple patterns) to document the expected behavior and prevent regression once the `[0]`-only issue is fixed.
- The `buildUrlMatcher` regex fallback path remains exposed to user-provided patterns (noted in self-review as pre-existing). Now that the code has a clean separation point, a follow-up to add a complexity guard (reject patterns longer than N characters or restrict to the `/<path>/*` format) would be straightforward.

## Assessment

Ready to merge? With fixes.

The single Important finding - asymmetric handling of multi-pattern rollback vs deploy - represents a correctness gap that should be addressed. If the system currently only produces single-element arrays, a minimal fix would be to iterate all elements (small change). The Minor items are hardening suggestions that can optionally be addressed in a follow-up.

## Next Steps

1. Fix the `allowedRegexPatterns[0]`-only rollback to iterate all patterns.
2. Optionally capture `isDomainWide` before `setData()` in the rollback pass for robustness.
3. Verify the CDN invalidation test removals were intentional.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 28s | <b>Cost:</b> $4.92 | <b>Commit:</b> <code>5f007830eb11d3d7c8c0a4f35d8567f3b1b88a7a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -42,11 +42,58 @@ const HTTP_BAD_REQUEST = 400;
...
 function isEdgeDeployableSuggestionStatus(status) {
   return status === 'NEW' || status === 'PENDING_VALIDATION';
 }
 
+/**
+ * Returns true if this suggestion uses pattern-based allow-list deploy (domain-wide or path-level).
+ * Detected by the presence of a non-empty allowedRegexPatterns array, or by isDomainWide: true
+ * (which may have an empty array when malformed — handled gracefully in the deploy loop).
+ */
+function isPatternSuggestion(suggestion) {
```
dipratap: please move this to suggestions utils

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -42,11 +42,58 @@ const HTTP_BAD_REQUEST = 400;
 const HTTP_INTERNAL_SERVER_ERROR = 500;
 const HTTP_NOT_IMPLEMENTED = 501;
 
+/** Returns a shallow copy of obj with the specified keys removed. */
+function omitKeys(obj, keys) {
```
dipratap: pls move this to utils

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -42,11 +42,58 @@ const HTTP_BAD_REQUEST = 400;
...
+
+/**
+ * Builds a URL matcher function for a given allowedRegexPatterns entry.
+ * Patterns ending with `/*` use pathname prefix matching (e.g. `/products/*` matches `/products/`).
+ * Other patterns are compiled as regular expressions for backward compatibility.
+ * Returns null for patterns that cannot be compiled.
+ * @param {string} pattern
+ * @returns {((url: string) => boolean)|null}
+ */
+function buildUrlMatcher(pattern) {
```
dipratap: pls move this to utils, lets try to avoid adding code to index file wherever possible

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -797,15 +844,148 @@ class TokowakaClient {
...
 
+  /**
+   * Removes a single pattern from the metaconfig's prerender allowList in-place.
+   * Deletes the prerender key entirely when the resulting list would be empty.
+   * @param {Object} metaconfig - Metaconfig object (mutated in place)
+   * @param {string} pattern - Pattern to remove
+   * @returns {boolean} True if the allowList was changed
+   * @private
+   */
+  #removePatternFromMetaconfig(metaconfig, pattern) {
```
dipratap: move to utils, maybe create metaconfig-utils

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -797,15 +844,148 @@ class TokowakaClient {
...
+    }
+    const keysToRemove = fieldsToStrip;
+    const results = await Promise.allSettled(covered.map(async (cs) => {
+      cs.setData(omitKeys(cs.getData(), keysToRemove));
+      cs.setUpdatedBy(updatedBy ?? actorFallback);
+      return cs.save();
+    }));
+    results.forEach((result, i) => {
+      if (result.status === 'rejected') {
+        this.log.warn(`[rollback] Failed to clean covered suggestion ${covered[i].getId()}: ${result.reason?.message}`);
```
dipratap: can we also consolidate and log this once, and tag it with [edge-rollback-failed] so that it comes under alerting?

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -817,115 +997,159 @@ class TokowakaClient {
       );
     }
 
-    // Validate which suggestions can be rolled back
-    // For rollback, we use the same canDeploy check to ensure data integrity
+    // Separate pattern-based suggestions (path and domain-wide allowList entries)
+    // from per-URL suggestions — each group is rolled back via a different code path.
+    const patternSuggestions = suggestions.filter(isPatternSuggestion);
+    const perUrlSuggestions = suggestions.filter((s) => !isPatternSuggestion(s));
+    this.log.info(`[rollback] Site: ${baseURL}, total: ${suggestions.length}, pattern: ${patternSuggestions.length}, perUrl: ${perUrlSuggestions.length}, allSuggestions: ${allSuggestions.length}`);
```
dipratap: IIRC tag used is [edge-rollback] at other places in rollback flow

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -817,115 +997,159 @@ class TokowakaClient {
...
+    this.log.info(`Updated Tokowaka configs for ${s3Paths.length} URLs, removed ${totalRemovedCount} patches total`);
 
-        // Set prerender to false (keep other patches if they exist)
-        const updatedConfig = {
-          ...existingConfig,
-          prerender: false,
-        };
+    // Save all eligible per-URL suggestions, stripping deployment markers.
+    const savedEligibleSuggestions = await Promise.all(
+      eligibleSuggestions.map((s) => this.#stripAndSaveSuggestion(s, 'tokowaka-rollback', updatedBy)),
```
dipratap: use Suggestions collection saveMany method to save multiple suggestions at once

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -817,115 +997,159 @@ class TokowakaClient {
...
+      } else {
+        // Pass 1: remove all patterns from the in-memory metaconfig, stage suggestions for saving.
+        const toSave = [];
+        let metaconfigChanged = false;
+
+        for (const suggestion of patternSuggestions) {
+          const data = suggestion.getData();
+          const patternToRemove = data?.allowedRegexPatterns?.[0];
+          if (!patternToRemove) {
+            this.log.warn(`[edge-rollback] Suggestion ${suggestion.getId()} has no allowedRegexPatterns, skipping`);
```
dipratap, line 980: Please ensure we are flaggin these kind of rollback failures in the flow somewhere in a consolidated manner. if the user rollbacks 5 suggestions, and 2 of them fails then we should one log for the overall flow saying '[edge-rollback-failed] 2 out of 5 suggestions failed ....'. This is to ensure alerting is able to catch these failures and surface to us. Also, only one log should be there since we dont want to have multiple alerts fired for one execution flow.

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -817,115 +997,159 @@ class TokowakaClient {
...
+          }
 
-        totalRemovedCount += 1; // Count as 1 rollback
-        // eslint-disable-next-line no-continue
-        continue;
-      }
+          const existingAllowList = metaconfig.prerender?.allowList ?? [];
+          const changed = this.#removePatternFromMetaconfig(metaconfig, patternToRemove);
+          const updatedAllowList = metaconfig.prerender?.allowList ?? [];
+          this.log.info(`[rollback] Pattern ${patternToRemove}: allowList before=${JSON.stringify(existingAllowList)}, after=${JSON.stringify(updatedAllowList)}`);
```
dipratap: [edge-rollback]

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -817,115 +997,159 @@ class TokowakaClient {
...
-      );
+        // Pass 2: upload the mutated metaconfig once, then save each suggestion and clean up.
+        try {
+          if (metaconfigChanged) {
+            await this.uploadMetaconfig(baseURL, metaconfig);
+          }
+        } catch (error) {
+          this.log.error(`[edge-rollback] Metaconfig upload failed: ${error.message}`, error);
+          toSave.forEach((s) => failedPatternSuggestions.push({
+            suggestion: s, reason: error.message, statusCode: 500,
```
dipratap: will this reason surface in api response? if yes, lets not capture the exact error here, and rather say internal server error

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -817,115 +997,159 @@ class TokowakaClient {
...
+            const covered = allSuggestions.filter(
+              (s) => s.getData()?.[coverageField] === suggestion.getId(),
+            );
+            const fieldsToStrip = isDomainWide
+              ? ['coveredByDomainWide']
+              : ['edgeDeployed', 'tokowakaDeployed',
+                'coveredByPattern'];
+            if (covered.length > 0) {
+              // eslint-disable-next-line max-len
+              this.log.info(`[rollback] Cleaning ${covered.length} covered suggestion(s) for pattern ${suggestion.getId()} (isDomainWide=${isDomainWide}, fallback=${coveredFallback})`);
```
dipratap: [edge-rollback]

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1266,6 +1490,223 @@ class TokowakaClient {
...
+  /**
+   * Classifies a batch of target suggestions into pattern-based and per-URL buckets.
+   * Pattern suggestions (domain-wide or path-level) are returned as
+   * `{ suggestion, allowedRegexPatterns }` objects; per-URL suggestions are filtered
+   * to only those with a deployable status.
+   * @param {Array} targetSuggestions
+   * @returns {{ patternSuggestions: Array, validSuggestions: Array }}
+   * @private
+   */
+  #classifySuggestions(targetSuggestions) {
```
dipratap: move this and the other applicable ones to suggestions-utils

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1266,6 +1490,223 @@ class TokowakaClient {
...
+    const patternSuggestions = [];
+    const validSuggestions = [];
+
+    targetSuggestions.forEach((suggestion) => {
+      const data = suggestion.getData();
+      if (isPatternSuggestion(suggestion)) {
+        this.log.info(
+          `[deploy] Classified as PATTERN: ${suggestion.getId()} `
```
dipratap: please fix tags in this file - use [edge-deploy] in the info and warn ones. 
[edge-deploy-failed] should only be used for flow breaking error logs since alerting is set up on them.

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1266,6 +1490,223 @@ class TokowakaClient {
...
+    });
+
+    this.log.info(`[deploy] Pattern ${suggestion.getId()}: found ${covered.length} coverable per-URL suggestions (field=${coverageField})`);
+
+    if (covered.length > 0) {
+      try {
+        await Promise.all(covered.map(async (cs) => {
+          cs.setData({ ...cs.getData(), [coverageField]: suggestion.getId() });
+          cs.setUpdatedBy(updatedBy);
+          return cs.save();
```
dipratap: Dont use savemany here - saying since it was suggested in one other place in this PR, this is being handled through https://jira.corp.adobe.com/browse/LLMO-5003.

package-lock.json
```diff
@@ -834,6 +834,7 @@
       "resolved": "https://registry.npmjs.org/@aws-sdk/client-dynamodb/-/client-dynamodb-3.940.0.tgz",
       "integrity": "sha512-u2sXsNJazJbuHeWICvsj6RvNyJh3isedEfPvB21jK/kxcriK+dE/izlKC2cyxUjERCmku0zTFNzY9FhrLbYHjQ==",
       "license": "Apache-2.0",
+      "peer": true,
```
dipratap: package-lock file shouldn't have any changes, since no changes to package.json

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1291,187 +1732,93 @@ class TokowakaClient {
...
     if (skippedInBatch.length > 0) {
-      const results = await Promise.allSettled(skippedInBatch.map(async (s) => {
-        s.setData({
-          ...s.getData(),
-          coveredByDomainWide: 'same-batch-deployment',
+      const results = await Promise.allSettled(skippedInBatch.map(async (item) => {
+        const coverageField = item.isDomainWide ? 'coveredByDomainWide' : 'coveredByPattern';
+        item.suggestion.setData({
+          ...item.suggestion.getData(),
+          [coverageField]: 'same-batch-deployment',
```
dipratap, line 1672: the current flow has the suggestion id of domain wide suggestion in coveredByDomainWide field.
"coveredByDomainWide": "274efe36-be4b-42f5-8d8d-5d2638cadee9"
The new code changes it. To keep things consistent, lets keep the suggestion id of the pattern suggestion here in value.

dipratap (changes requested):

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: Current flow : https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-tokowaka-client/src/index.js#L1458

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: `#removePatternFromMetaconfig` is the only one that can move - it's a pure function with no this dependency. All the other metaconfig methods (`fetchMetaconfig`, `uploadMetaconfig`, `createMetaconfig`, `updateMetaconfig`, `#updateMetaconfigWithDeployedPaths`, `#fetchMetaconfigWithMetadata`) are instance methods that use this.s3Client, this.log, this.bucketName, etc.

Should this be moved given that it's just one function?

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: done

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: done

package-lock.json
ssilare-adobe: done

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: I don't think it is avoidable, as per the ticket, the optimization is based on the `chunkSize` while using it with `saveMany` as `100` instead of `25` (default)

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: Sure, moved [here](https://github.com/adobe/spacecat-shared/pull/1598/changes#diff-cb267e46e6e12d28e71646fb26060078c0269c1c182cf015d39319102a0cb438R1087)

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: done

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: Discussed over call, moved `removePatternFromMetaconfig`

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: Addressed — pattern/domain-wide suggestion saves now use `Promise.allSettled(toSave.map(s => s.save()))` (individual saves in parallel, not `saveMany`) in both deploy and rollback flows.

Per-URL suggestions continue using `saveSuggestions` → `saveMany` as before.

See commit 0508a632.

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: Resolved as discussed over call

packages/spacecat-shared-tokowaka-client/src/index.js
ssilare-adobe: Valid — the `reason` field was surfacing raw `error.message` in the API response via `item.reason` in `spacecat-api-service/src/controllers/suggestions.js:2414`.

Fixed all 4 occurrences in deploy/rollback flows to use `'Internal server error'` instead. The raw error is still logged server-side for debugging.

See commit e0d6124b.

dipratap (approved):

dipratap (approved):

MysticatBot (changes requested):
Hey @ssilare-adobe,

Re-reviewing after the 37+ commits since the last review. The refactoring and reviewer feedback incorporation are well done - the extracted utils, unified log prefixes, consolidated error logging, and independent coverage layers are all solid improvements. Two issues remain that should be addressed before merge.

## Strengths

- Clean decomposition into focused utility modules (`metaconfig-utils.js`, `pattern-utils.js`, expanded `suggestion-utils.js`) significantly improves readability and testability of the main orchestration class.
- Append-and-deduplicate deploy semantics via `[...new Set([...existing, ...patterns])]` prevents duplicate entries and correctly elides CDN writes when nothing changes.
- Independent coverage layers (`coveredByDomainWide` / `coveredByPattern`) with layer-specific stripping during rollback are architecturally sound. Tests verify DW rollback preserves `coveredByPattern` and vice versa - well-modeled independence.
- Single metaconfig fetch shared across all pattern suggestions in a batch (both deploy and rollback) reduces S3 read costs vs the prior per-suggestion approach.
- Per-suggestion error isolation ensures one failure does not contaminate the succeeded/failed lists. `cleanupCoveredSuggestions` swallows individual failures gracefully.
- All prior reviewer feedback addressed cleanly: `saveMany` with chunkSize 25, consolidated `[edge-rollback-failed]` alerting log, standardized log prefixes, raw error.message replaced with generic strings, utils extraction, Promise.allSettled for pattern suggestions.
- Comprehensive test coverage (~1200 new test lines) covering rollback independence, double-coverage preservation, error isolation, missing metaconfig, upload failures, temporal ordering, cascade-skip scenarios, and same-batch coverage field correctness.
- Previously flagged issues now addressed: getData() re-read after setData() is handled better, CDN invalidation tests reorganized, package-lock.json reverted.

## Issues

### Important (Should Fix)

**1. Rollback only removes `allowedRegexPatterns[0]`, silently dropping additional patterns**

`packages/spacecat-shared-tokowaka-client/src/index.js` (rollback pattern loop):
```js
const patternToRemove = data?.allowedRegexPatterns?.[0];
```

Deploy correctly handles all entries via `addPatternsToMetaconfig(metaconfig, allowedRegexPatterns)` and builds matchers using `allowedRegexPatterns.flatMap(...)`. Rollback takes only `[0]`. If a suggestion ever has multiple patterns (e.g., `['/products/*', '/services/*']`), rollback removes only the first - the rest remain orphaned in the CDN metaconfig allowList with no owning suggestion.

**Why it matters**: Deploy and rollback must be symmetric. An orphaned pattern means the CDN continues prerendering URLs the user intended to roll back, with no UI indication the rollback was partial.

**How to fix**: Iterate all patterns:
```js
const patternsToRemove = data?.allowedRegexPatterns ?? [];
if (patternsToRemove.length === 0) {
  this.log.warn(`[edge-rollback] Suggestion ${suggestion.getId()} has no allowedRegexPatterns, skipping`);
  ineligibleSuggestions.push({ suggestion, reason: 'Missing allowedRegexPatterns' });
  continue;
}
for (const pattern of patternsToRemove) {
  if (removePatternFromMetaconfig(metaconfig, pattern)) {
    metaconfigChanged = true;
  }
}
```

Also add a test for rollback of a suggestion with `allowedRegexPatterns: ['/products/*', '/blog/*']` to prevent regression.

**2. `addPatternsToMetaconfig` and `removePatternFromMetaconfig` overwrite sibling keys on `metaconfig.prerender`**

`packages/spacecat-shared-tokowaka-client/src/utils/metaconfig-utils.js`:
```js
metaconfig.prerender = { allowList: merged };   // line 52
metaconfig.prerender = { allowList: updated };  // line 34
```

Both functions replace the entire `prerender` object. If `metaconfig.prerender` has sibling keys beside `allowList` (e.g., `ttl`, `excludePatterns`), they are silently dropped. The PR description states "Sibling keys on `metaconfig.prerender` are preserved via spread" but no spread is present in the implementation.

**Why it matters**: The metaconfig is shared infrastructure. Silent data loss on deploy or rollback is hard to diagnose and recover from.

**How to fix**: Preserve sibling keys with spread:
```js
// addPatternsToMetaconfig:
metaconfig.prerender = { ...metaconfig.prerender, allowList: merged };

// removePatternFromMetaconfig (non-empty case):
metaconfig.prerender = { ...metaconfig.prerender, allowList: updated };
```

For the `delete metaconfig.prerender` path (empty allowList), consider checking whether other keys exist before deleting the entire object.

### Minor (Nice to Have)

**3. Regex fallback in `buildUrlMatcher` matches against full URL string, not pathname**

`packages/spacecat-shared-tokowaka-client/src/utils/pattern-utils.js`:
```js
// /* path: matches against pathname
return (url) => new URL(url).pathname.startsWith(prefix);
// regex fallback: matches against full URL string
return (url) => regex.test(url);
```

A regex like `^/blog` would never match because the URL string starts with `https://`. The inconsistency between the two code paths could confuse future callers. Consider matching against pathname in both branches for consistency, or adding a JSDoc note documenting the contract.

**4. `buildUrlMatcher` prefix matching may over-match path boundaries**

`packages/spacecat-shared-tokowaka-client/src/utils/pattern-utils.js`:
```js
const prefix = pattern.slice(0, -2); // '/products/*' -> '/products'
return (url) => new URL(url).pathname.startsWith(prefix);
```

Pattern `/products/*` produces prefix `/products`, which matches `/productsabc` and `/products-new` in addition to the intended `/products/...`. Consider appending a trailing slash to the prefix comparison (with a special case for `/*` where prefix is empty).

**5. `buildUrlMatcher` has no dedicated unit test file**

The function is a pure utility on the critical path for both deploy coverage and rollback. Its edge cases (pattern `/*`, empty strings, invalid regex, pathname vs full-URL inconsistency) deserve direct unit tests in `test/utils/pattern-utils.test.js` rather than relying solely on integration coverage.

## Recommendations

- Add the multi-pattern rollback test to explicitly document and guard against issue 1.
- The ReDoS concern in `buildUrlMatcher` (regex fallback compiles user-provided patterns) is pre-existing on main. The `/*` fast path mitigates the common case. Consider a follow-up to add a complexity guard or restrict patterns to the `/<path>/*` format.
- Consider formalizing the metaconfig schema contract (even a JSDoc typedef) as more features attach to `metaconfig.prerender.*`.

## Assessment

Ready to merge? With fixes.
Reasoning: The rollback asymmetry (issue 1) is a correctness gap that could leave orphaned patterns in production CDN configs, and the sibling-key overwrite (issue 2) contradicts the PR's documented contract. Both are small, focused fixes (5-10 lines total). The rest of the design is sound, well-tested, and addresses all prior reviewer feedback.

## Next Steps

1. Fix the `allowedRegexPatterns[0]`-only rollback to iterate all patterns.
2. Add spread to preserve sibling keys in both metaconfig-utils functions.
3. Add a test for multi-pattern rollback.
4. Minor items are optional - can be addressed in a follow-up.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 9s | <b>Cost:</b> $5.42 | <b>Commit:</b> <code>12b72fb5e050007eeb4054a5bc4c8da998464760</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1599, opened 2026-05-12
fix(http-utils): address RO admin wrapper review findings

ravverma:
## Summary

Follow-up to [PR #1595](https://github.com/adobe/spacecat-shared/pull/1595) addressing the Critical / Important / Minor findings from @solaris007's [review](https://github.com/adobe/spacecat-shared/pull/1595#pullrequestreview-4269055881).

### Critical (defense-in-depth adopted)

The reviewer's attack scenario premise - "developer adds a new route and forgets BOTH `routeCapabilities` AND `internalRoutes`" - is prevented by an existing in-place test in `spacecat-api-service` that asserts every mounted route is enumerated in one of the two maps. The attack is therefore not reachable in practice.

However, the reviewer's recommended fix is a strict improvement to the wrapper's contract regardless of the api-service test, so it has been adopted as defense-in-depth:

- **Body fallback now requires an explicit capability mapping.** When `capability === null` AND no path params are resolvable (route is in NEITHER `routeCapabilities` NOR `internalRoutes`), the wrapper denies up-front with `reason: 'unmapped-no-path-params'`. The body's siteId/orgId claim is no longer consulted in that case.
- Legitimate body-fallback routes (e.g. `POST /preflight/jobs` with body siteId) are unaffected because they have an explicit capability mapping (`'POST /preflight/jobs': 'preflight:write'`).
- The wrapper is now safe to drop into a consumer that does not have route-exhaustiveness tests.

### Important

1. **Permissive default for unmapped writes** - The ownership-first model is preserved for routes in `internalRoutes` (path-params resolvable). The Critical fix above closes the body-fallback escape hatch. `unmapped-route-allowed` drift warn still fires so consumers can detect routeCapabilities drift.
2. **`routeCapabilities` non-object input silently fails open** - `guardNonEmptyRouteCapabilities` tightened to require a non-empty plain object. Rejects undefined, null, arrays, strings, numbers. `s2sAuthWrapper` inherits this guard.
3. **`spaceCatId` alias couples shared utils to api-service** - Added `paramAliases` option (default `{}`). Consumers declare `{ spaceCatId: 'organizationId' }` at the call site. The wrapper no longer hardcodes a service-specific routing convention.
4. **`internalRoutes` parallel source of truth** - Kept as designed. The api-service test prevents drift; collapsing to a single map would require a sentinel value scheme that adds its own complexity. Documented the relationship in JSDoc.
5. **Parent-child resource ambiguity** - Documented explicit handler contract in JSDoc: handlers behind path-param authorization MUST verify nested resource ids belong to the authorized parent before mutating them.
6. **Feature flag scoped to `tenantIds[0]`** - Documented the "user-org gate" semantics. Added a test pinning multi-tenant ordering behavior so the choice cannot silently flip later.

### Minor

1. **Silent failure on missing accessor** - `dataAccess.Site` / `dataAccess.Organization` absence now emits `log.error`.
2. **siteId-over-organizationId precedence** - Documented in `isOwnerOfResource` JSDoc.
3. **Stack-trace leakage** - Error logs now use `{ errMessage: err.message, errName: err.name }` instead of `{ err }` to avoid DB driver internals (table names, region, connection ids) leaking through the structured logger.
4. **Coverage refinements** - Body-fallback access-log test asserts `ro-admin-audit` is NOT emitted (suppression contract) and asserts `resolvedOrgId`. New test for org-route access log with `resolvedOrgId`.
5. **organizationId precedence over spaceCatId not pinned** - Added two tests: canonical wins over alias when both are set in body; spaceCatId is ignored when `paramAliases` is not configured.

### Recommendations addressed

- Cross-domain consistency note: JSDoc cross-references `s2sAuthWrapper` as a deliberately-different-policy sibling.
- Performance note: JSDoc documents per-write DB round-trip cost.
- FF-disabled denial now includes `reason: 'feature-flag-disabled'` field for SIEM filterability (also covers a prior Minor).

### Recommendations not adopted in this PR

- **Pluggable `isOwner` strategy** - Larger refactor; worth its own PR.
- **Startup-time invariant check of mounted routes vs `routeCapabilities`** - Already enforced at the consumer (api-service) level; not adding a duplicate check to the shared wrapper.

## Test plan

- [x] 344 tests passing, 100% lines / statements coverage, 97.64% branches (threshold 97%).
- [x] Lint clean.
- [ ] Verify api-service tests still pass against this version (test the wrapper with the existing routeCapabilities + internalRoutes maps; add `paramAliases: { spaceCatId: 'organizationId' }` at the call site).

### DIFF
--- a/packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
+++ b/packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
@@ -11,7 +11,6 @@
  */
 
 import { Response } from '@adobe/fetch';
-import { isObject } from '@adobe/spacecat-shared-utils';
 import { LaunchDarklyClient } from '@adobe/spacecat-shared-launchdarkly-client';
 
 import { FF_READ_ONLY_ORG } from './constants.js';
@@ -29,17 +28,40 @@ function forbidden(message) {
 }
 
 /**
+ * Resolves a canonical identifier (e.g. 'siteId', 'organizationId') from a source map,
+ * preferring the canonical key, then trying any configured aliases that map to it.
+ *
+ * @param {Object|undefined} source - Path params or context.data
+ * @param {string} canonical - Canonical key name ('siteId' or 'organizationId')
+ * @param {Object<string, string>} aliases - Alias map (alias name -> canonical name)
+ * @returns {string|null} The resolved id, or null
+ */
+function resolveId(source, canonical, aliases) {
+  if (!source) return null;
+  if (source[canonical] !== undefined) return source[canonical];
+  for (const [alias, target] of Object.entries(aliases)) {
+    if (target === canonical && source[alias] !== undefined) {
+      return source[alias];
+    }
+  }
+  return null;
+}
+
+/**
  * Checks whether the authenticated read-only admin user owns the resource identified
  * by the path parameters or request body. Ownership is determined by whether the user's
  * IMS org matches the organization that owns the referenced site or organization entity.
  *
  * ID resolution order:
- * 1. Named path params (e.g. :siteId, :organizationId, :spaceCatId) — always preferred.
- *    spaceCatId is an alias for organizationId used by /v2/orgs/:spaceCatId/* routes
- *    in spacecat-api-service.
+ * 1. Named path params (e.g. :siteId, :organizationId, or any alias configured via
+ *    paramAliases such as :spaceCatId -> organizationId) — always preferred.
  * 2. context.data (parsed request body) — used only when the matched route has no path
  *    params at all (e.g. POST /preflight/jobs passes siteId in the body). Requires that
  *    the body-parser wrapper runs before readOnlyAdminWrapper in the .with() chain.
+ *
+ * Precedence: siteId wins over organizationId when both are resolvable; the site lookup
+ * is performed and organizationId is ignored. Routes that surface both should rely on
+ * siteId as the canonical owner reference.
  *
  * Fail-closed: returns false if dataAccess is unavailable, the entity is not found,
  * no recognizable ID can be resolved, or any lookup throws.
@@ -47,9 +69,10 @@ function forbidden(message) {
  * @param {Object} context - Universal context (must have context.dataAccess and context.log)
  * @param {Object} authInfo - The AuthInfo instance for the current user
  * @param {Object<string, string>} params - Named path params extracted from the route pattern
+ * @param {Object<string, string>} paramAliases - Alias map for canonical id resolution
  * @returns {Promise<boolean>}
  */
-async function isOwnerOfResource(context, authInfo, params) {
+async function isOwnerOfResource(context, authInfo, params, paramAliases) {
   const { dataAccess, log } = context;
   if (!dataAccess) {
     log.error({ tag: 'ro-admin' }, 'isOwnerOfResource: dataAccess not on context — ensure dataAccessWrapper runs before readOnlyAdminWrapper');
@@ -60,14 +83,17 @@ async function isOwnerOfResource(context, authInfo, params) {
   // When params has keys, the route does have path identifiers; relying on body data in that
   // case could allow a caller to spoof ownership by naming params differently than :siteId.
   const hasPathParams = Object.keys(params).length > 0;
-  const siteId = hasPathParams ? params.siteId : context.data?.siteId;
-  const organizationId = hasPathParams
-    ? (params.organizationId ?? params.spaceCatId)
-    : (context.data?.organizationId ?? context.data?.spaceCatId);
+  const source = hasPathParams ? params : context.data;
+  const siteId = resolveId(source, 'siteId', paramAliases);
+  const organizationId = resolveId(source, 'organizationId', paramAliases);
 
   try {
     if (siteId) {
-      const site = await dataAccess.Site?.findById(siteId);
+      if (!dataAccess.Site) {
+        log.error({ tag: 'ro-admin' }, 'isOwnerOfResource: dataAccess.Site accessor is missing');
+        return false;
+      }
+      const site = await dataAccess.Site.findById(siteId);
       if (!site) {
         return false;
       }
@@ -84,7 +110,11 @@ async function isOwnerOfResource(context, authInfo, params) {
     }
 
     if (organizationId) {
-      const org = await dataAccess.Organization?.findById(organizationId);
+      if (!dataAccess.Organization) {
+        log.error({ tag: 'ro-admin' }, 'isOwnerOfResource: dataAccess.Organization accessor is missing');
+        return false;
+      }
+      const org = await dataAccess.Organization.findById(organizationId);
       if (!org) {
         return false;
       }
@@ -96,7 +126,11 @@ async function isOwnerOfResource(context, authInfo, params) {
       return authInfo.hasOrganization(imsOrgId);
     }
   } catch (err) {
-    log.error({ tag: 'ro-admin', err }, 'Error checking resource ownership for RO admin');
+    log.error({
+      tag: 'ro-admin',
+      errMessage: err.message,
+      errName: err.name,
+    }, 'Error checking resource ownership for RO admin');
     return false;
   }
 
@@ -112,6 +146,13 @@ async function isOwnerOfResource(context, authInfo, params) {
  * Evaluates the read-only admin feature flag for the authenticated user's IMS org.
  * Uses {@link AuthInfo#getTenantIds} to resolve the org and
  * {@link LaunchDarklyClient#isFlagEnabledForIMSOrg} for evaluation.
+ *
+ * Tenant scoping semantics: this is a USER-ORG GATE. The flag is evaluated against the
+ * user's primary tenant (`getTenantIds()[0]`). For a multi-tenant user whose tenant list
+ * is `[A, B]`, the flag state for A decides access; B is not consulted. This is order-
+ * dependent: the same user with tenant list `[B, A]` would be gated by B's flag state.
+ * Resource-level org membership is enforced separately by the subsequent ownership check.
+ *
  * Fail-closed: returns false when the client/org is unavailable or evaluation errors.
  *
  * @param {Object} context - Universal context (lambda context)
@@ -135,7 +176,11 @@ async function evaluateFeatureFlag(context, authInfo) {
     const imsOrgId = `${ident}@AdobeOrg`;
     return await ldClient.isFlagEnabledForIMSOrg(FF_READ_ONLY_ORG, imsOrgId);
   } catch (err) {
-    log.error({ tag: 'ro-admin', err }, 'Feature flag evaluation failed for RO admin; defaulting to deny');
+    log.error({
+      tag: 'ro-admin',
+      errMessage: err.message,
+      errName: err.name,
+    }, 'Feature flag evaluation failed for RO admin; defaulting to deny');
     return false;
   }
 }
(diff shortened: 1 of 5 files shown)


packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js
```diff
@@ -112,6 +150,13 @@ async function isOwnerOfResource(context, authInfo, params) {
  * Evaluates the read-only admin feature flag for the authenticated user's IMS org.
  * Uses {@link AuthInfo#getTenantIds} to resolve the org and
  * {@link LaunchDarklyClient#isFlagEnabledForIMSOrg} for evaluation.
```
solaris007, line 152: **Important 1 - Multi-tenant FF JSDoc oversells cross-tenant safety.**

The sentence "Resource-level org membership is enforced separately by the subsequent ownership check" is misleading. The ownership check (`authInfo.hasOrganization(imsOrgId)`) verifies the user is a MEMBER of the resource's owning org. It does NOT verify that org has `FT_READ_ONLY_ORG` enabled.

Concrete scenario: user with `tenants=[A, B]`, `FF(A)=ON`, `FF(B)=OFF`. The FF check on A passes, then the user can write to a B-owned resource because they're a member of B. B's deliberate FF-off stance is bypassed.

Either honestly document the gap, or evaluate the FF against the resource's tenant after ownership resolves (larger change, can be follow-up).

packages/spacecat-shared-http-utils/test/auth/s2s-wrapper.test.js
```diff
@@ -353,18 +363,11 @@ describe('s2sAuthWrapper', () => {
...
-        method: 'GET',
-        suffix: '/sites',
-        headers: { authorization: `Bearer ${token}` },
-      };
-      const wrapped = s2sAuthWrapper(handler);
-      const result = await wrapped({}, context);
-
-      expect(result).to.deep.equal({ status: 200 });
-      expect(handler.calledOnce).to.be.true;
+    it('requires routeCapabilities at construction time (no implicit pass-through)', () => {
```
solaris007, line 366: **Important 2 - Undocumented breaking change to `s2sAuthWrapper`.**

This test replaces the previously-passing "passes through when no routeCapabilities is provided" test. `s2sAuthWrapper` is a public export; any consumer calling `s2sAuthWrapper(handler)` without options will now throw at module load.

The hardening is correct, but the PR body says only that the guard is "inherited" and doesn't characterize this as breaking. Either add a `BREAKING CHANGE:` line to the commit body for semantic-release, or note explicitly in the PR description that no external callers were found (`org:adobe` code search confirms no callers in this org - worth stating).

packages/spacecat-shared-http-utils/src/auth/route-utils.js
```diff
@@ -104,15 +104,16 @@ export function extractRouteParams(context, routeMap, fallbackRoutes = []) {
...
  * @param {string} wrapperName - Name of the wrapper (for the error message)
  * @param {Object} routeCapabilities - The route capabilities map
  */
 export function guardNonEmptyRouteCapabilities(wrapperName, routeCapabilities) {
-  if (isObject(routeCapabilities) && Object.keys(routeCapabilities).length === 0) {
-    throw new Error(`${wrapperName}: routeCapabilities must not be an empty object`);
+  if (!isObject(routeCapabilities) || Object.keys(routeCapabilities).length === 0) {
+    throw new Error(`${wrapperName}: routeCapabilities must be a non-empty object`);
   }
 }
```
solaris007, line 119: **Important 3 - Cleanup is incomplete in `s2s-wrapper.js`.**

(Anchoring here because the related file is not in this hunk - the actual cleanup gap is in `packages/spacecat-shared-http-utils/src/auth/s2s-wrapper.js:14` and `s2s-wrapper.js:109`.)

This PR removed the redundant `if (isObject(routeCapabilities))` block from `read-only-admin-wrapper.js` because the construction guard makes it unreachable when false. The identical block at `s2s-wrapper.js:109` was not removed, and the `isObject` import on `s2s-wrapper.js:14` is now unused.

Both wrappers share the same guard; the dead branch in s2s-wrapper reads as if s2s is still permissive about missing capabilities - exactly the confusion the cleanup was meant to eliminate.

solaris007 (approved):
Hey @ravverma,

Thorough follow-up. The prior Critical body-fallback IDOR is closed, the defense-in-depth deny block is placed correctly (before ownership check, with a precise condition that admits legitimate body-fallback routes via their explicit non-null capability mapping), `paramAliases` is well-designed with canonical-wins precedence robust to attacker-supplied source ordering, and the JSDoc work is substantial. Both pushbacks (permissive default for unmapped writes, internalRoutes parallel source of truth) hold up - no new exploit surface identified.

Approving with a handful of follow-ups: one auth-correctness JSDoc gap that's worth a quick fix in this PR, two code/release-hygiene items, and some polish. I trust you to handle these; nothing here needs another full review cycle. (CI lint is understood to be coming separately.)

## Strengths (previously flagged items, now resolved)

- **Body-fallback IDOR (prior Critical) closed.** Early-deny `if (capability === null && !hasPathParams)` fires before `isOwnerOfResource`. New test asserts `Site.findById` is never reached, proving deny is up-front rather than after a wasted DB round-trip.
- **`paramAliases` cleanly decouples the shared library from spacecat-api-service's `:spaceCatId` convention.** Default `{}` keeps the wrapper neutral; api-service supplies the alias at the call site. "Prefers canonical organizationId over spaceCatId alias" and "ignores spaceCatId in body when paramAliases is not configured" pin the contract.
- **`guardNonEmptyRouteCapabilities` tightened** to reject undefined/null/array/string/empty-object. The silent-passthrough class that the old `if (isObject(...))` soft-skip enabled is gone.
- **Error log shape standardized** (`errMessage`/`errName` over `err`) across all three call sites - SIEM-friendly, no stack-trace leakage.
- **Multi-tenant FF ordering pinned by test** ("pins multi-tenant ordering: only the first tenant is checked"). Prevents silent semantic drift; asserts `calledOnce` so `tenantIds[1]` cannot leak in.
- **Accessor-missing fail-closed and logged.** `dataAccess.Site` / `dataAccess.Organization` absent emits `log.error` and returns false. Tests verify both the 403 and the log emission.
- **JSDoc is substantially clearer.** Handler contracts for body-fallback and nested-path-params, sibling-wrapper cross-reference, per-write DB cost, user-org gate semantics - all explicit.
- **IMS org format verified correct.** `getTenantIds()` returns bare idents; `${ident}@AdobeOrg` constructs canonical IMS org IDs; `AuthInfo.hasOrganization()` strips the suffix before comparing.
- **All five prior Minor findings addressed.** Stack-trace shape, siteId-over-organizationId precedence, accessor missing, audit-log suppression on body-fallback, organizationId precedence pinned by test.
- **Pushbacks engaged with substantively.** Important 1 (permissive default): body-fallback escape closed; remaining ownership-on-unmapped-via-internalRoutes path requires deliberate consumer enumeration plus drift warn. No exploitable scenario survives. Important 4 (internalRoutes parallel source of truth): dual-map kept with consumer-side test enforcement plus JSDoc; sentinel-collapse alternative trades comparable complexity.

## Follow-ups (please address before/with the lint fix)

### Important

1. **Multi-tenant FF JSDoc oversells the cross-tenant safety** (`packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js`, `evaluateFeatureFlag` JSDoc). The sentence "Resource-level org membership is enforced separately by the subsequent ownership check" is misleading. The ownership check (`authInfo.hasOrganization(imsOrgId)`) verifies the user is a MEMBER of the resource's owning org. It does NOT verify that org has `FT_READ_ONLY_ORG` enabled. Concrete scenario: user with `tenants=[A, B]`, `FF(A)=ON`, `FF(B)=OFF`. The FF check on A passes, then the user can write to a B-owned resource because they're a member of B. B's deliberate FF-off stance is bypassed. Either honestly document the gap ("the wrapper does NOT verify the resource's owning org has FT_READ_ONLY_ORG enabled; a multi-tenant user gated on tenant A can act on resources in tenants without the FF enabled if they are members of those tenants") or evaluate the FF against the resource's tenant after ownership resolves (larger change, can be follow-up).

2. **`s2sAuthWrapper` passthrough removal is an undocumented breaking change** (`packages/spacecat-shared-http-utils/test/auth/s2s-wrapper.test.js`, replaced "passes through when no routeCapabilities is provided" with "requires routeCapabilities at construction time"). `s2sAuthWrapper` is a public export; any consumer calling `s2sAuthWrapper(handler)` without options will now throw at module load. The hardening is correct posture, but the PR body says only that the guard is "inherited" - it does not characterize this as breaking. Either add a `BREAKING CHANGE:` line to the commit body so semantic-release picks the right bump, or note explicitly in the PR description that no external callers were affected (staff-engineer noted `org:adobe` code search found no callers - worth confirming in the description).

3. **Dead code left in `s2s-wrapper.js` after parallel cleanup** (`packages/spacecat-shared-http-utils/src/auth/s2s-wrapper.js:14, 109`). This PR removed the redundant `if (isObject(routeCapabilities))` block from `read-only-admin-wrapper.js` because the construction guard makes it unreachable when false, but the identical block at `s2s-wrapper.js:109` was not removed and the `isObject` import on line 14 is now unused. Both wrappers share the same guard; the dead branch in s2s-wrapper reads as if s2s is still permissive about missing capabilities - exactly the confusion the cleanup was meant to eliminate.

### Minor

4. **`resolveId` is recomputed for the access log** (`packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js`, around the `ro-admin-access` log emission). After `isOwnerOfResource` already resolved `siteId` and `organizationId`, the wrapper re-resolves both. The risk isn't cost - it's drift between the auth decision and the audit trail if `resolveId` or its inputs change in one site but not the other. See the Recommendations.

5. **Inconsistent log shape across the three warn sites** (`packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js`). `not-owner` includes `email`, `method`, `suffix`, `org`, `reason`; `unmapped-route-allowed` and `unmapped-no-path-params` omit `email`. Same logical concern (denied/allowed admin route), different field sets. Pick a canonical shape and apply uniformly for SIEM filtering parity.

6. **`reason: 'feature-flag-disabled'` is logged but not asserted in any test** (`packages/spacecat-shared-http-utils/test/auth/read-only-admin-wrapper.test.js`, FF-disabled denial test). The three other new reason fields (`unmapped-no-path-params`, `not-owner`, `unmapped-route-allowed`) are all asserted in tests. This one isn't. A refactor could quietly drop the reason and break SIEM filters without test failure. Fix: add `reason: 'feature-flag-disabled'` to the `calledWithMatch` arg.

7. **Test name leaks review history** (`packages/spacecat-shared-http-utils/test/auth/read-only-admin-wrapper.test.js`, "denies access on unmapped route with no path params even when body claims ownership (Critical fix)"). Drop the `(Critical fix)` parenthetical. Tests should describe behavior, not git history.

8. **Body-fallback handler contract should explicitly forbid reading alias keys** (`packages/spacecat-shared-http-utils/src/auth/read-only-admin-wrapper.js`, "Handler contracts" JSDoc section). When `paramAliases` is configured and body fallback applies, the wrapper authorizes against the canonical key (`resolveId` precedence). If an attacker sends `{ organizationId: 'owned-A', spaceCatId: 'victim-B' }`, the wrapper authorizes A but a handler mistakenly reading `body.spaceCatId` would operate on B. Add: "When `paramAliases` is configured, handlers MUST read the canonical key (`siteId`/`organizationId`) for the resource being mutated, NOT the alias key - the wrapper authorizes against canonical."

## Recommendations

- **Have `isOwnerOfResource` return `{ isOwner, resolvedSiteId, resolvedOrgId }`** so the wrapper logs exactly what was authorized against. Single source of truth; resolves Minor 4.
- **Export a `validateRoAdminWrapperConfig(routeCapabilities, internalRoutes, paramAliases, mountedRoutes)` helper** for consumer test suites. With three config maps and the integrity invariant enforced only at the consumer, centralizing the contract scales better as more services adopt the wrapper. Not blocking - worth tracking for the second consumer that adopts this.

## Assessment

Approving conditionally. The substantive work is done and the substantive concerns from the prior review are properly addressed. The three Important items are text-only fixes (one JSDoc clarification, one release-hygiene callout, one cleanup pass) and the rest is polish. CI lint is understood to be coming separately.

PR #1602, opened 2026-05-12
feat(data-access): add semrushWorkspaceId attribute to Organization

solaris007:
## Summary

Adds `semrushWorkspaceId` as a top-level string attribute on the `Organization` model in `spacecat-shared-data-access`. Auto-generates `findBySemrushWorkspaceId` / `allBySemrushWorkspaceId` collection accessors via `addAllIndex`. Mirrors the `imsOrgId` pattern.

## Part of

Adobe IMS to Semrush integration (first pass). Parent proposal: **adobe/mysticat-architecture#61**.

Step 2 of a multi-repo sequence:
- ⏳ Step 1 - DB column: **adobe/mysticat-data-service#593** (parallel; needs to merge before this lands for the IT test to be un-skipped)
- 👉 Step 2 - this PR
- Step 3 - spacecat-api-service consumes the new shared-data-access version (follow-up)
- Step 4 - project-elmo-ui consumes the new field (follow-up)

## What's changed

- `organization.schema.js` - `addAttribute('semrushWorkspaceId', { type: 'string' })` + `addAllIndex(['semrushWorkspaceId'])`
- `index.d.ts` - TS declarations for getter / setter / `findBy*` / `allBy*`
- `organization.model.test.js` - unit tests for getter / setter
- `organization.collection.test.js` - minimal "method exists" tests for the auto-generated accessors
- `organizations.fixture.js` - fixture[0] gets `semrushWorkspaceId: 'ws_fixture_001'`
- `organization.test.js` (IT) - skipped IT test for `findBySemrushWorkspaceId` (un-skipped after Step 1 image is published)

## Open follow-up on this branch (not yet committed)

After **adobe/mysticat-data-service#593** merges and the new Docker image is published:
- Bump `MYSTICAT_DATA_SERVICE_TAG` default in `test/it/postgrest/docker-compose.yml`
- Remove `it.skip(...)` on the new IT test

## Local verification

- `npm run lint -w packages/spacecat-shared-data-access` ✅
- `npm test -w packages/spacecat-shared-data-access` ✅ - 1973 passing, coverage 100% lines / 97%+ branches
- IT test for new accessor is intentionally `it.skip`'d pending Step 1

### DIFF
--- a/docs/plans/2026-05-12-data-access-organization-add-semrush-workspace-id.md
+++ b/docs/plans/2026-05-12-data-access-organization-add-semrush-workspace-id.md
@@ -1,0 +1,589 @@
+# `spacecat-shared-data-access`: Add `semrushWorkspaceId` to Organization - Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Add a top-level `semrushWorkspaceId` string attribute to the `Organization` model in `spacecat-shared-data-access`, with auto-generated `findBySemrushWorkspaceId` / `allBySemrushWorkspaceId` collection accessors via `addAllIndex`. Mirrors the `imsOrgId` pattern.
+
+**Architecture:** Single SchemaBuilder `.addAttribute` + `.addAllIndex` call on the Organization schema. TypeScript declarations follow. Unit tests use `createElectroMocks` (no DB). Integration test follows the existing `findByImsOrgId` IT pattern but is gated on the parallel DB migration landing in `mysticat-data-service` and a new Docker image being published.
+
+**Tech Stack:** Node.js, ESM, ElectroDB-shaped SchemaBuilder, sinon (for unit-test mocks), mocha + chai (test runner), c8 (coverage; 100% lines / 97% branches enforced), semantic-release (versioning), husky + lint-staged (pre-commit), ESLint.
+
+**Parent proposal:** [adobe/mysticat-architecture#61](https://github.com/adobe/mysticat-architecture/pull/61) - "Adobe IMS Integration with Semrush (First Pass)"
+
+**Position in sequence:**
+- Step 1 = `mysticat-data-service` PR ([adobe/mysticat-data-service#593](https://github.com/adobe/mysticat-data-service/pull/593)) - adds the DB column
+- **Step 2 = this plan** - exposes the attribute on the data-access model
+- Step 3 = `spacecat-api-service` consumes the new shared-data-access version (separate plan)
+- Step 4 = `project-elmo-ui` consumes new `semrushWorkspaceId` from org response (separate plan)
+
+**Dependency on Step 1:** unit tests + schema + types + fixture changes can ship and merge independently. The integration test for `findBySemrushWorkspaceId` depends on the data-service Docker image (`MYSTICAT_DATA_SERVICE_TAG` in `test/it/postgrest/docker-compose.yml`, currently pinned to `v5.1.1`) being rebuilt with the new column. The IT test is added with `it.skip(...)` plus a TODO; once Step 1 merges and a new image is tagged, a small follow-up bumps the tag and un-skips the test.
+
+**Workshop context:** Workshop is happening this week (Basel, May 2026). Step 1 PR is open. Step 2 can be authored and reviewed in parallel.
+
+---
+
+### Task 1: Branch off `origin/main`
+
+**Files:** none yet
+
+- [ ] **Step 1: Verify clean working tree**
+
+```bash
+cd /Users/dj/adobe/github/adobe/spacecat-shared
+git status
+git fetch origin main
+```
+
+Expected: clean, `origin/main` up to date.
+
+- [ ] **Step 2: Create feature branch from `origin/main`**
+
+```bash
+git switch -c feat/data-access-organization-semrush-workspace-id origin/main
+```
+
+---
+
+### Task 2: Write failing unit tests for getter / setter
+
+**Files:**
+- Modify: `packages/spacecat-shared-data-access/test/unit/models/organization/organization.model.test.js`
+
+- [ ] **Step 1: Add the test block after the existing `imsOrgId` block (line 86)**
+
+In `packages/spacecat-shared-data-access/test/unit/models/organization/organization.model.test.js`, find the existing `describe('imsOrgId', ...)` block (around line 76) and add this new block immediately after it:
+
+```javascript
+  describe('semrushWorkspaceId', () => {
+    it('gets semrushWorkspaceId', () => {
+      expect(instance.getSemrushWorkspaceId()).to.equal('ws_test_existing');
+    });
+
+    it('sets semrushWorkspaceId', () => {
+      instance.setSemrushWorkspaceId('ws_test_new');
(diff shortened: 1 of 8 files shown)


ekremney (approved):

packages/spacecat-shared-data-access/test/unit/models/organization/organization.model.test.js
rainer-friederich, line 1: **[FIX — score 85] Missing absent-field unit test**

The plan document committed with this PR (Task 6) explicitly specifies a third test case that was not implemented:

```js
it('returns undefined when semrushWorkspaceId is absent on the record', () => {
  // use organizations[1] or organizations[2], which have no semrushWorkspaceId
  expect(instance2.getSemrushWorkspaceId()).to.be.undefined;
});
```

Most production orgs today have no `semrushWorkspaceId`, so the absent-field path is the dominant real-world state at rollout. The analogue PR for `llmBackend` (1588) required this exact coverage before merge. The two fixtures without the field are already in place — the test just needs to use one of them.

packages/spacecat-shared-data-access/src/models/organization/organization.schema.js
rainer-friederich, line 1: **[FIX — score 75] `findBySemrushWorkspaceId` limit:1 on a non-unique index**

`addAllIndex` generates `findBySemrushWorkspaceId` backed by `findByIndexKeys`, which issues a PostgREST query with `limit=1`. If two orgs ever share a workspace ID, the accessor silently returns one arbitrary row — no error, no warning.

`imsOrgId` has the same dual-accessor pattern, but Adobe IMS enforces global uniqueness on that identifier. Semrush workspace IDs have no such system-enforced guarantee.

Two options:

- Confirm that `mysticat-data-service#593` adds a `UNIQUE` constraint on the `semrush_workspace_id` column, and add a comment in this schema file pointing at that constraint so the invariant is documented at both layers.
- If uniqueness cannot be guaranteed, remove `findBySemrushWorkspaceId` from the generated/declared surface and keep only `allBySemrushWorkspaceId`, forcing callers to handle 0/1/N results explicitly.

Without one of these, a workspace-ID collision would produce a silent cross-tenant data leak.

rainer-friederich, line 1: **[FIX — score 75] No `validate` function on the attribute**

Every other lookup-key attribute on this schema has a validator: `imsOrgId` uses a regex, `llmBackend` uses an enum check, `fulfillableItems` and `config` use structural guards. `semrushWorkspaceId` ships with `{ type: 'string' }` only, which allows empty string `''`, whitespace-only values, and arbitrarily large payloads.

The plan notes that format validation is deferred pending Semrush confirming the canonical ID shape — that is reasonable. But a minimum non-empty guard is orthogonal to the format question:

```js
.addAttribute('semrushWorkspaceId', {
  type: 'string',
  validate: (v) => !v || (typeof v === 'string' && v.trim().length > 0),
})
```

Without this, an empty string stored in the field would pollute `findBySemrushWorkspaceId('')` lookups once the IT is wired up.

docs/plans/2026-05-12-data-access-organization-add-semrush-workspace-id.md
rainer-friederich, line 1: **[FIX — score 75] Implementation plan document should not be committed to this repo**

This document contains content that is not appropriate for the shared library's git history:

- A hardcoded absolute local path (`/Users/dj/adobe/github/adobe/spacecat-shared`) that is not reproducible for any other contributor
- Internal agentic tooling directives (`REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development`) that are meaningless to external contributors of this Apache-licensed package
- Ephemeral workshop context ("Workshop is happening this week (Basel, May 2026)") that will be stale the moment this merges
- Checkbox task-tracking scaffolding that won't be updated in the repo after merge

Per the workspace documentation hierarchy, cross-repo implementation plans belong in `mysticat-architecture`, not in individual service or library repos. This is a 4-step multi-repo sequence — the plan covers all four steps and should live in `mysticat-architecture`.

The code changes are clean and can merge on their own. Recommend dropping this file from the PR. Any durable context worth preserving (architecture rationale, the rollout sequencing) can go in the commit message body, which semantic-release surfaces in the changelog.

packages/spacecat-shared-data-access/src/models/organization/index.d.ts
rainer-friederich, line 1: **[CONSIDER — score 50] `setSemrushWorkspaceId` has no typed way to clear the value**

The setter is declared as `(semrushWorkspaceId: string): Organization`, which gives TypeScript callers no type-safe path to unlink a workspace (clear the field). If an org needs to be delinked from Semrush, callers would need a cast.

`setImsOrgId` uses the same pattern, so this is consistent with existing conventions. If the unlink use case isn't planned for Steps 3–4, this is fine to defer. If it is, changing the type now (`string | null`) avoids a semver-minor bump later.

Worth confirming with the Semrush integration design before Step 3 locks in the API contract.

rainer-friederich, line 1: **[SKIP — score 25] `getSemrushWorkspaceId(): string` return type is non-nullable**

The attribute has no `required: true` and no default, so existing orgs will return `undefined` at runtime. The declared type `string` tells TypeScript callers it is always present, which will cause them to skip null-checks they need.

The correct type would be `string | undefined`.

Noting this as low-priority because `getImsOrgId(): string` has the same incorrect signature on `main` — this PR is propagating an established (wrong) convention rather than introducing a new one. Fixing both would be the right long-term move but is out of scope for this PR.

rainer-friederich, line 1: **[SKIP — score 25] Cardinality of the Org → Semrush workspace relationship**

Both `findBySemrushWorkspaceId` (returns one) and `allBySemrushWorkspaceId` (returns many) are generated and declared. This is the standard output of `addAllIndex` — the same pattern `imsOrgId` uses — so no action is needed here structurally.

Worth noting for the Step 3 and Step 4 reviewers: the data-access layer takes no position on whether the relationship is 1:1 or 1:N. The API and UI layers will need to pick a convention. If the domain decision is 1:1, that should be enforced at the DB level (UNIQUE constraint in mysticat-data-service#593) and documented in a comment on the schema attribute, not just implied by which accessor the API chooses to call.

rainer-friederich (changes requested):
Review via local multi-agent team (data-engineer, senior-engineer, qa-engineer, senior-technical-writer, domain-expert).

**4 items to fix before merge:**

1. Add the absent-field unit test (`getSemrushWorkspaceId()` returns `undefined` when the field is not set) — specified in the committed plan doc but dropped from the implementation
2. Resolve the uniqueness contract for `findBySemrushWorkspaceId`: confirm mysticat-data-service#593 adds a UNIQUE constraint and document it here, or remove `findBy*` and keep only `allBy*`
3. Add a minimum non-empty `validate` function to the `semrushWorkspaceId` attribute — format validation can stay deferred, but empty strings should be rejected now
4. Remove the plan document from this PR — it contains local absolute paths, internal agentic directives, and ephemeral workshop context; the code changes are clean and can merge independently

**2 lower-priority items (non-blocking):**

5. Consider widening `setSemrushWorkspaceId` to `string | null` if the unlink use case is planned for Steps 3–4
6. `getSemrushWorkspaceId(): string` should be `string | undefined` — pre-existing issue with `getImsOrgId` too, not introduced here
7. Cardinality note for Step 3/4 reviewers: the data-access layer is silent on 1:1 vs 1:N; that decision should be enforced at the DB level if 1:1 is the intent

PR #1603, opened 2026-05-13
fix: TierClient.getFirstEnrollment() to return entitlement for paid o…

sandsinh:
## Summary
- \`getFirstEnrollment()\` was incorrectly returning \`{ entitlement: null }\` in all cases where no matching \`SiteEnrollment\` row was found, even when a valid org-level entitlement exists. This caused \`/sites-resolve\` to treat PAID orgs as having no entitlement, showing the PLG onboarding gate to paid customers.
- Fixed three return sites in \`getFirstEnrollment()\` to always pass through the \`entitlement\` object when it exists, only nulling \`enrollment\` and \`site\` when those are absent.
- Updated three corresponding unit tests to assert the correct behavior: entitlement is returned even when enrollment/site is null.

## Test plan
- [ ] Run \`npm test -w packages/spacecat-shared-tier-client\` — all tests should pass
- [ ] Verify a PAID org with no \`SiteEnrollment\` rows no longer triggers PLG rejection in \`/sites-resolve\`
- [ ] Confirm an org with no entitlement at all still returns \`{ entitlement: null, enrollment: null, site: null }\`


Sugandhgyl (dismissed):
The fix is correct and directly addresses the reported bug. The `resolveByOrg` logic in `spacecat-api-service `was already structured to handle { entitlement, site: null } — it just was never getting that shape from **getFirstEnrollment** before. 

Tests are appropriately updated. The **PLG/FREE_TRIAL** orphaned-entitlement edge case is a behavioral change but arguably an improvement.

Consider **adding a short inline comment** to each of the three return sites explaining the semantic — `"entitlement exists but no matching enrollment/site found" `— so future readers understand why entitlement is deliberately non-null in an otherwise "not found" scenario.

ravverma (commented):
Hey @sandsinh,

## Strengths

- All three return sites in `getFirstEnrollment()` are updated in lockstep with matching test assertions (`packages/spacecat-shared-tier-client/src/tier-client.js:315,327,342` and `test/tier-client.test.js:1186,1212,1274`).
- The unchanged "no entitlement at all" early return (`src/tier-client.js:309`) still asserts `entitlement: null` in tests (`test/tier-client.test.js:1173-1184`), so the "no entitlement" signal remains distinguishable from the new "entitlement-without-enrollment" signal.
- The fix brings `getFirstEnrollment()` into semantic alignment with the sibling `getAllEnrollment()` (lines 213-281), which already preserves the entitlement when the enrollment collection is empty. The contract divergence is repaired without expanding scope.
- Test descriptions were renamed from the vague "should return nulls when ..." to the contract-accurate "should return entitlement with null enrollment and site when ...", which will not mislead future readers.

## Issues

### Minor (Nice to Have)

1. `packages/spacecat-shared-tier-client/src/tier-client.js:315,327,342` - the three return sites now carry non-obvious semantics ("entitlement exists but no matching enrollment/site"). A future reader scanning this method may not immediately understand why `entitlement` is deliberately non-null in what otherwise looks like a "not found" branch, and may "normalize" the returns back to the old shape. Pick up Sugandhgyl's suggestion either as inline one-liners at each return or as a JSDoc note on the method (lines 296-301) documenting the three possible return shapes: `{ entitlement, enrollment, site }` (full match), `{ entitlement, enrollment: null, site: null }` (entitlement without enrollment), `{ entitlement: null, enrollment: null, site: null }` (no entitlement).

2. `packages/spacecat-shared-tier-client/test/tier-client.test.js` - the post-loop return at `src/tier-client.js:342` is reachable two ways in the org-only iteration: every site returns null from `findById` (the `if (!site)` warn branch), or every site returns a non-null site with a mismatching `getOrganizationId()` (the `else if` branch). Only the first path is exercised by the existing tests (the `Site.findById.resolves(null)` test at line 1200). A test with a non-null site whose org ID does not match, exhausting the loop via the mismatch branch, would prevent a future regression where someone reverts line 342 to `entitlement: null` without breaking any current test. About 15 lines of test code.

## Recommendations

- Add JSDoc to `getFirstEnrollment()` at `src/tier-client.js:296-301` documenting the three-state return contract. The new "entitlement non-null, enrollment null, site null" shape is now load-bearing semantics that the PLG paywall gate in `/sites-resolve` depends on; making it explicit at the API boundary protects against unintentional drift.
- Out of scope, worth tracking: `getFirstEnrollment()` issues N sequential `findById` calls in the org-only loop (lines 333-342) to dodge PostgREST URL-length limits, while `getAllEnrollment()` solves the same constraint with chunked `batchGetByKeys` of 50. Two different solutions to the same constraint in adjacent methods is the kind of accidental divergence that is cheap to align now. Not a finding here.

## Assessment

Ready to merge? Yes.

Reasoning: This is a correct, well-scoped bug fix with matching test updates, an existing MEMBER approval, and downstream compatibility already confirmed in the prior discussion (`resolveByOrg` was already structured to handle `{ entitlement, site: null }`). The Minor items are documentation and an additional test path - useful but non-blocking.

ravverma (dismissed):

MysticatBot (approved):
Hey @sandsinh,

## Strengths

- `src/tier-client.js:300-308` - The JSDoc documenting the three return shapes is well-written. It names the consumer (`/sites-resolve`) and the semantic distinction each shape represents. This protects against future regression by a developer who might normalize the returns back to all-nulls. Directly addresses feedback from both prior reviewers.
- `src/tier-client.js:321,333,348` - The fix is minimal and correct. Three one-token changes that preserve the already-fetched entitlement object without altering any behavioral paths beyond the return value shape.
- The fix aligns `getFirstEnrollment()` with the sibling `getAllEnrollment()`, which already preserves the entitlement when the enrollment collection is empty. This repairs a contract divergence between two methods callers reasonably expect to behave consistently.
- `test/tier-client.test.js:1186` - Test descriptions renamed from the vague "should return nulls when..." to the contract-accurate "should return entitlement with null enrollment and site when..." - self-documenting test output.
- `test/tier-client.test.js:1282-1318` - New test covers the loop-exhaustion path through the org-ID mismatch branch, directly addressing the prior reviewer's suggestion.
- No new attack surface introduced. The entitlement object was already returned in the happy path, so returning it in the no-match paths does not leak data that was not already accessible. The fix actually improves security posture by ensuring paid orgs are not incorrectly downgraded.

## Issues

### Minor (Nice to Have)

1. `test/tier-client.test.js:1283` - In the new test, `wrongOrgSite1` reuses the outer `siteId` variable via `getId: () => siteId`, while `wrongOrgSite2` uses `'other-site-id'`. The test works correctly because the assertion is about org ID mismatch, but reusing `siteId` for a "wrong" site could confuse a future reader into thinking the test exercises site-ID matching. Using a distinct ID like `'wrong-site-id-1'` would make the intent clearer at a glance.

## Assessment

Ready to merge? Yes.

This is a correct, well-scoped bug fix with matching test updates, explicit JSDoc documenting the new three-state contract, and a new test filling a previously-uncovered branch. Two prior reviewers approved, all feedback has been addressed, and CI is green.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 29s | <b>Cost:</b> $2.19 | <b>Commit:</b> <code>2b6909718da647cb2ede03e575f89c40924c96b3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1604, opened 2026-05-13
feat: www.krisshop.com.au onboarding failed

tkotthakota-adobe:
https://jira.corp.adobe.com/browse/SITES-44465

Problem: www.krisshop.com.au onboarding failed with APPRUM ❌, Scraping ❌, SEO Import ❌.
        
Description:                                                                                                                                     
wwwUrlResolver resolves a site's effective hostname by checking which variant (www. or non-www.) has RUM data. For www.krisshop.com.au, it toggled to krisshop.com.au (apex), found a RUM key there from old traffic, and returned it — without checking whether the apex has a valid SSL cert. It doesn't. Every downstream consumer (scraper/Puppeteer, PSI client, audit worker) then tried to connect to https://krisshop.com.au and failed with ERR_CERT_AUTHORITY_INVALID / unable to verify the first certificate.

Fix (in packages/spacecat-shared-utils/src/url-helpers.js):                                                                                

Added isHttpsReachable(hostname) — a HEAD request with 5s timeout and redirect: 'manual' that returns true/false. After wwwUrlResolver finds a RUM key on the toggled hostname, it now calls isHttpsReachable. If SSL fails, it logs a warning and falls through to try the original hostname (www.krisshop.com.au) instead.                                                                                           
                                                            
After release: bump @adobe/spacecat-shared-utils in both spacecat-audit-worker and spacecat-api-service.


packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js
```diff
@@ -243,20 +285,21 @@ class Configuration {
...
-      } else {
-        handler.enabled[entityKey] = Array
-          .from(new Set([...(handler.enabled[entityKey] || []), entityId]));
-      }
+      // Always remove from disabled first (handles re-enabling a previously disabled entry).
+      handler.disabled[entityKey] = (handler.disabled[entityKey] || [])
+        .filter((id) => id !== entityId);
+      // Always add to enabled so isHandlerEnabledForSite can find the site in enabled.sites
+      // and return true even when the org is in disabled.orgs (site-level override).
+      handler.enabled[entityKey] = Array
```
solaris007: **Critical 1 (see review body):** This writer now adds the site to `enabled.sites` for `enabledByDefault=true` handlers too. Before this PR, this branch only mutated `disabled.sites`. Downstream consequence: `getEnabledSiteIdsForHandler('alt-text-decorative-agent')` flips from empty -> non-empty on the first admin-driven site enable, which silently inverts the canary gate in `spacecat-audit-worker/src/image-alt-text/handler.js:32-37` (`return !enabledSites.length`). Either keep the writer asymmetric for default-enabled handlers, split the consumer's canary onto a dedicated method, or sequence the audit-worker fix first.

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -283,12 +283,25 @@ function toggleWWWHostname(hostname) {
...
 
+async function isHttpsReachable(hostname) {
+  try {
+    await fetch(`https://${hostname}`, {
+      method: 'HEAD',
+      signal: AbortSignal.timeout(5000),
+      redirect: 'manual',
+    });
+    return true;
+  } catch {
```
solaris007: **Important 3:** `catch {}` collapses DNS NXDOMAIN, TLS errors, TCP timeouts, AbortError, 405 / 403 / 5xx into one boolean. The caller's log line says nothing about why. Pass `log` into this helper and `log.warn` (or debug) the `e.name`+`e.message` inside the catch. Add a test that distinguishes timeout from TLS error - today both collapse to a single failure mode.

packages/spacecat-shared-data-access/test/unit/models/configuration/configuration.model.test.js
```diff
@@ -258,6 +359,7 @@ describe('ConfigurationModel', () => {
     it('updates handler sites for a handler enabled by default', () => {
       instance.updateHandlerSites('404', site.getId(), true);
       expect(instance.getHandler('404').disabled.sites).to.not.include(site.getId());
+      expect(instance.getHandler('404').enabled.sites).to.include(site.getId());
```
solaris007: **Important 4:** Good new assertion on the site side. The symmetric test on the org side (`updates handler orgs for a handler enabled by default with enabled`) was not updated and lacks the matching `expect(...).enabled.orgs.to.include(orgId)` assertion. Since `#updatedHandler` is shared between sites and orgs, a regression that flips just the org-side writer would not be caught by the current suite. Two-line fix.

packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js
```diff
@@ -183,26 +198,53 @@ class Configuration {
...
+    // Site-level takes priority over org-level: an explicit site enable/disable
+    // overrides the org-level setting, allowing a single site to be upgraded to
+    // a paid profile even if the org is in disabled.orgs.
+    if (disabledSites.includes(siteId)) {
+      return false;
+    }
+    if (enabledSites.includes(siteId)) {
+      if (disabledOrgs.includes(orgId)) {
+        // TODO(SITES-44467): demote to debug after 30-day canary window
+        this.log.info('[isHandlerEnabledForSite] site-override: site in enabled.sites overrides org in disabled.orgs', { type, siteId, orgId });
```
solaris007: **Important 5:** Two concerns with this hot-path log:

1. `isHandlerEnabledForSite` is called in tight per-handler-per-site loops in jobs-dispatcher, audit-worker, and reporting-worker. 30 days of unsampled INFO at this cardinality is meaningful Coralogix volume. Prefer a counter (dashboardable, alertable) or sampled log (1/N).
2. Constructor takes `log` positionally with no default. Pre-PR this method had no log calls. Any caller passing a partial logger (test stubs, scripts) will now throw `TypeError: this.log.info is not a function`. Defensive `this.log?.info?.(...)` is cheap insurance during the canary, or move the diagnostic to the writer (it fires when the override is set - rare and intentional - rather than every read).

Confirm SITES-44467 is on someone's queue before this lands in prod.

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -310,8 +323,11 @@ async function wwwUrlResolver(site, rumApiClient, log) {
   try {
     const wwwToggledHostname = toggleWWWHostname(hostname);
     await rumApiClient.retrieveDomainkey(wwwToggledHostname);
-    log.debug(`Resolved URL ${wwwToggledHostname} for ${baseURL} using RUM API Client`);
-    return wwwToggledHostname;
+    if (await isHttpsReachable(wwwToggledHostname)) {
```
solaris007: **Important 6:** `wwwUrlResolver` is on the audit-worker / api-service hot path. This 5s probe pays its full ceiling on every failure, every invocation - no module-scoped cache, no memoization. SSL state changes on the order of months; re-probing every call is wasteful and risks tipping audit batches over their per-Lambda time budget. Add a module-scoped `Map<hostname, {result, expiresAt}>` with ~1h TTL.

solaris007 (changes requested):
Hey @tkotthakota-adobe,

This PR bundles two distinct changes under an SSL-fix banner: the `wwwUrlResolver` HEAD probe for the krisshop scenario, and a rewrite of `isHandlerEnabledForSite` precedence + the `#updatedHandler` writer in `configuration.model.js`. The SSL probe is fine in shape but is missing observability and a reachability check on the fallback path. The configuration model rewrite is the more impactful change and is silently bundled - it changes the meaning of `getEnabledSiteIdsForHandler` for `enabledByDefault=true` handlers, which breaks an existing canary gate in spacecat-audit-worker. Detailed findings below.

## Strengths

- New JSDoc on `isHandlerEnabledForSite` and `isHandlerEnabledForOrg` (configuration.model.js:174-188, 220-237) makes the precedence ladder and the intentional site-vs-org asymmetry explicit and discoverable.
- The writer in `#updatedHandler` is now symmetric and idempotent: enable always removes from disabled and adds to enabled, disable always does the reverse for default-enabled handlers. This restores the disjoint-list invariant that the new precedence depends on.
- `isHttpsReachable` uses safe defaults: HEAD only, 5s timeout, `redirect: 'manual'`, hardcoded https scheme.
- Test coverage of the site-side truth table is thorough: site-in-both-lists, site-disabled, site-enabled-with-org-disabled, site-enabled, org-disabled, default-enabled, and explicit idempotency.
- `nock.cleanAll()` added in `afterEach` (url-helpers.test.js) is good hygiene and prevents test-order coupling.

## Issues

### Critical (Must Fix)

**1. Writer change silently flips `getEnabledSiteIdsForHandler` semantics for default-enabled handlers, breaking the decorative-agent canary gate**

`packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js:285-301`

Before this PR, for `enabledByDefault=true` handlers the writer never populated `enabled.sites` (the old `if (handler.enabledByDefault)` branch only removed from disabled). After this PR, every `updateHandlerSites(type, siteId, true)` adds the site to `enabled.sites` regardless of `enabledByDefault`.

`getEnabledSiteIdsForHandler(type)` returns `handler.enabled.sites || []`. It is unchanged in this PR, but its meaning just shifted.

Downstream consequence in `spacecat-audit-worker/src/image-alt-text/handler.js:27-37`:

```javascript
// Enabled when alt-text-decorative-agent handler has no enabled.sites (empty = enabled for all).
export async function isDecorativeAgentEnabled(context) {
  const { Configuration } = context.dataAccess;
  const configuration = await Configuration.findLatest();
  const enabledSites = configuration.getEnabledSiteIdsForHandler('alt-text-decorative-agent');
  return !enabledSites.length;
}
```

Today, an admin calling `updateHandlerSites('alt-text-decorative-agent', siteId, true)` on a default-enabled handler is a no-op for this canary (`enabled.sites` stays empty, `isDecorativeAgentEnabled` returns `true`). After this PR ships and audit-worker bumps its dep, the same admin action populates `enabled.sites` with one site - and `isDecorativeAgentEnabled` flips to `false` across all sites. The decorative agent silently stops running everywhere it was previously enabled-by-default.

Mitigation options - pick one:
1. Keep the writer asymmetric for `enabledByDefault=true` handlers (do not add to enabled.sites - encode the site-override via a separate `overrides.sites` field).
2. Rename `getEnabledSiteIdsForHandler` to `getExplicitlyEnabledSiteIdsForHandler` and give the audit-worker canary a dedicated method like `getOverrideSitesForHandler` - clean semantic split.
3. Fix the consumer first: ship audit-worker with a different canary mechanism, validate, then ship this PR.

Without one of these, this PR ships a downstream regression with no test in this repo that would catch it.

### Important (Should Fix)

**2. PR bundles two unrelated changes; title and description omit the configuration model rewrite**

The PR title is `feat: www.krisshop.com.au onboarding failed` and the body describes only the SSL probe. The configuration model rewrite is a ~90-line policy change with cross-service runtime semantics, mentioned only in a single commit message ("site-level handler overrides were ignored, org-level disable always won") which is not visible in the merged PR title or body.

Two consequences: (a) reviewers anchor on the title topic and the bundled change gets less scrutiny than it deserves, (b) future bisect/blame for a handler-enable regression will not find this PR by message search. The two changes also have different blast radii - utils is consumed via direct upgrade; data-access carries cross-service runtime semantics in a singleton S3 config - and should ride different release cadences.

Fix: split into two PRs. SSL probe as a patch to spacecat-shared-utils. Configuration model change as a semver-minor of spacecat-shared-data-access with a release-note callout for the precedence change and the data-shape change. Either way, update the PR title and body to explicitly call out the policy change.

**3. `isHttpsReachable` swallows all errors with bare `catch {}`**

`packages/spacecat-shared-utils/src/url-helpers.js:286-296`

```javascript
async function isHttpsReachable(hostname) {
  try {
    await fetch(`https://${hostname}`, { method: 'HEAD', signal: AbortSignal.timeout(5000), redirect: 'manual' });
    return true;
  } catch {
    return false;
  }
}
```

DNS NXDOMAIN, TLS errors, TCP timeouts, AbortError, 405 HEAD-rejected origins, anti-bot 403s, transient 5xx - all collapse to the same `false`. The caller's log message ("HTTPS check failed; trying X") says nothing about why. The next krisshop-style incident is harder to triage than the original was, because the diagnostic signal is lost.

Minimum fix: pass `log` into `isHttpsReachable` and log the error class inside the catch:

```javascript
async function isHttpsReachable(hostname, log) {
  try { await fetch(...); return true; }
  catch (e) { log.warn(`HTTPS check failed for ${hostname}: ${e.name} ${e.message}`); return false; }
}
```

Add a test that distinguishes a timeout (e.g. `nock(...).delay(6000)` against the 5s signal) from a TLS error (`replyWithError('CERT_AUTHORITY_INVALID')`) - currently both are exercised as a single failure mode.

**4. Org-side writer behavior is untested**

`packages/spacecat-shared-data-access/test/unit/models/configuration/configuration.model.test.js`

`#updatedHandler` is shared by `updateHandlerSites` and `updateHandlerOrgs`. Both code paths changed. The site-side gained a new assertion (`expect(...).enabled.sites).to.include(site.getId())` after enabling a default-enabled handler), but the existing org-side test for the same scenario was not updated:

```javascript
it('updates handler orgs for a handler enabled by default with enabled', () => {
  instance.updateHandlerOrgs('lhs-mobile', org.getId(), true);
  expect(instance.getHandler('lhs-mobile').enabled.orgs).to.include(org.getId());
  // ... no symmetric assertion for the default-enabled-handler case ...
});
```

Add the symmetric assertion - two lines - so a regression to the org-side writer is caught.

**5. Hot-path `log.info` line with 30-day canary; also raises TypeError risk for stub loggers**

`packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js:213-216`

`isHandlerEnabledForSite` is called in `getEnabledAuditsForSite` and `getDisabledAuditsForSite` (one call per handler per site) and in every per-site, per-handler dispatch loop in jobs-dispatcher, audit-worker, and reporting-worker. The new line fires whenever a site is in `enabled.sites` AND its org is in `disabled.orgs`. For any affected site, this fires N times per cycle.

Two follow-on concerns:
- 30 days of unsampled INFO logs at this cardinality is meaningful Coralogix volume. Either replace with a counter (cheaper, dashboardable, alertable) or sample 1/N.
- `isHandlerEnabledForSite` had no log calls before this PR. The constructor takes `log` positionally with no default. Any existing caller that passes a partial logger (test fixtures, scripts, downstream service init) will now throw `TypeError: this.log.info is not a function` instead of returning a boolean. Defensive `this.log?.info?.(...)` is cheap insurance during the canary, or move the diagnostic to the writer (it fires when the override is set, which is rare and intentional) rather than the reader (every audit cycle).

Confirm SITES-44467 exists and is on someone's queue before this lands in prod, otherwise the TODO ships and stays at INFO indefinitely.

**6. 5-second synchronous HTTPS probe on a hot path with no cache**

`packages/spacecat-shared-utils/src/url-helpers.js:286-336`

`wwwUrlResolver` is called by audit-worker and api-service on every site resolution. The probe pays up to 5s on failure on every call - no module-scoped cache, no request-scoped memoization. SSL state for a domain changes on the order of months, so re-probing every invocation is pure waste. For an audit batch resolving many sites with bad-SSL toggled hostnames, the additive cost can push Lambdas over their per-invocation budget.

Fix: add a module-scoped `Map<hostname, {result, expiresAt}>` with a TTL on the order of 1 hour. The customer-domain set is bounded; an unbounded cache is acceptable, an LRU is nicer.

**7. Cross-version rollout creates a forward-only data divergence with no rollback runbook**

The configuration is a singleton in S3, read by api-service, audit-worker, jobs-dispatcher, autofix-worker, reporting-worker on every invocation. After this PR ships and downstream services bump independently, a configuration written by a new-version writer (site in `enabled.sites` + org in `disabled.orgs` for a default-enabled handler) will be read inconsistently:
- New reader: returns `true` (precedence step 2)
- Old reader: returns `false` (`handler.disabled.orgs.includes(orgId)` wins)

Same data, different decision. There is no `schemaVersion` field on the handler shape and no rollback runbook. If you need to revert the library, the data already written with the new shape stays. S3 versioning helps only if ops knows which version ID to pin to.

Fix: add a CHANGELOG entry calling out the data-shape change as forward-only; add a brief rollback note in SITES-44467 covering the S3 version-pin procedure.

### Minor (Nice to Have)

**8. Fallback path returns original hostname without HTTPS check**

`packages/spacecat-shared-utils/src/url-helpers.js:323-344`

The fix validates HTTPS only on the toggled hostname. If the toggled has no RUM key but the original has bad SSL, `wwwUrlResolver` returns the broken original anyway - the same class of bug it is fixing, applied to the symmetric path. No test covers "both unreachable" or "RUM key on original, original has bad SSL". A domain where both apex and www have bad SSL still produces the same opaque scraper/PSI failure this PR set out to fix.

Either HTTPS-check the second branch too, or emit a counter / warn when neither variant has been verified.

**9. Pre-deploy audit: enumerate existing (enabled.sites, disabled.orgs) overlap**

Any (handler, siteId) pair already in `enabled.sites` for a site whose org is in `disabled.orgs` will flip from "handler disabled" to "handler enabled" the moment this code ships. Sensitive handlers in scope: `security-permissions`, `security-csp`, `security-csp-auto-suggest`, `security-vulnerabilities-auto-suggest`, `summit-plg`, the `*-auto-fix` handlers, `cdn-logs-analysis`. The intent per the inline comment is "free-trial -> paid migration," but the precedence applies retroactively to all existing rows.

Mitigation: admin capability gating + entitlement-layer enforcement (`SiteEnrollment`) means this is not a full bypass. But before deploy, run a one-time query against the production Configuration S3 object to enumerate every (handler, siteId) where siteId is in `enabled.sites` AND the site's org is in `disabled.orgs` for that handler. Operator should confirm each match is intended.

**10. SSRF defense in depth on `isHttpsReachable`**

`packages/spacecat-shared-utils/src/url-helpers.js:286-296`

The hostname is derived from `site.getBaseURL()`. `isValidUrl` only checks the protocol; nothing rejects `https://10.0.0.1`, `https://127.0.0.1`, `https://169.254.169.254`, or `*.internal` hostnames. Exploitability today is low (audit-worker Lambdas are not VPC-attached, AWS IMDS is HTTP not HTTPS, and the HEAD probe is gated by `rumApiClient.retrieveDomainkey`) but the moment audit-worker moves into a VPC for MCT, or someone reuses `isHttpsReachable` in a VPC-attached Lambda, the SSRF surface opens up. The neighboring `composeAuditURL` and `resolveCanonicalUrl` have the same risk profile (pre-existing).

Out of immediate scope for this PR, but worth a defense-in-depth helper that rejects RFC1918 / loopback / link-local / `*.internal`.

## Recommendations

- Consider moving the "paid site in unpaid org" override out of `Configuration` (a capability switch) and into the `Entitlement` layer. Today's precedence change uses the wrong primitive - capability config will accrete entitlement exceptions over time. The `Entitlement` model is already imported in this file.
- Replace the 30-day INFO log with either a counter (for dashboards/alerts) or a sampled log (1/N).
- Add CHANGELOG entries for both spacecat-shared-utils and spacecat-shared-data-access. The data-access entry needs to call out the precedence change, the data-shape change, and the rollback note.
- Add an explicit test for the "default-disabled handler, no site-level entries, org in `enabled.orgs`" row and a disable-idempotence test counterpart to the enable case.
- Clarify in the `isHttpsReachable` JSDoc that it is a TLS-reachability probe, not a content-correctness probe (returns true for 4xx/5xx responses; HEAD reachability does not imply the apex serves the same content as the www).
- Rebase the lockfile `"peer": true` metadata changes into a separate `chore: regenerate lockfile` PR so a `git revert` of the feature does not also revert the lockfile metadata.
- Document the site-vs-org precedence asymmetry as a platform invariant in `mysticat-architecture` so consumers across api-service, audit-worker, jobs-dispatcher, reporting-worker, fulfillment-worker share the same mental model.

- Audit of `getEnabledSiteIdsForHandler` consumers across spacecat-audit-worker (`isDecorativeAgentEnabled`) and spacecat-jobs-dispatcher (`scrapes.js:35`) for the new data-shape semantics. Belongs in those repos.
- Audit of `isHandlerEnabledForSite` / `isHandlerEnabledForOrg` consumers across the platform to confirm correct site-vs-org evaluator choice for each workflow.
- Adding the SSRF-defense hostname guard to `composeAuditURL` and `resolveCanonicalUrl` in the same file (pre-existing risk).
- A platform-wide rollout-coordination story for configuration-schema-affecting library bumps - worth an ADR.

## Assessment

Reasoning: The SSL probe fix is correct in spirit and the precedence rewrite is well-documented in JSDoc. However, the writer change silently breaks a load-bearing canary gate in audit-worker (critical finding 1), and the rollout / observability / error-handling concerns mean this should not ship as-is.

Minimum bar before merge:
1. Resolve the canary inversion (critical finding 1) either in this PR or by sequencing audit-worker first.
2. Add the org-side writer test assertion (important finding 4).
3. Fix the `isHttpsReachable` error suppression (important finding 3).
4. Update the PR title and description to call out the configuration model rewrite (important finding 2) - or split the PR.

The latency / cache, rollback runbook, and pre-deploy audit items are also material but tractable in the same review cycle.

## Next Steps

1. Address the critical decorative-agent canary inversion - this is the load-bearing blocker.
2. Resolve the bundled-PR concern (split, or rewrite the title/body to surface the config change).
3. Fix the seven important items in order of how easy they are to land.
4. Treat the minor items and recommendations as follow-up work; ship via separate PRs.

packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js
tkotthakota-adobe: removed it.

packages/spacecat-shared-data-access/test/unit/models/configuration/configuration.model.test.js
tkotthakota-adobe: removed it

packages/spacecat-shared-data-access/src/models/configuration/configuration.model.js
tkotthakota-adobe: removed it

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -378,4 +408,5 @@ export {
   hasNonWWWSubdomain,
   toggleWWWHostname,
   wwwUrlResolver,
+  HTTPS_REACHABLE_CACHE,
 };
```
solaris007, line 495: **Important 1:** Exporting `HTTPS_REACHABLE_CACHE` makes the Map shape (`Map<string, { result, expiresAt }>`) part of `spacecat-shared-utils`'s public API. Any future caching change (LRU, sized cap, distributed cache, swap to a class) becomes a breaking change for the package. Replace the export with a narrow test helper: `export function resetHttpsReachableCacheForTests() { HTTPS_REACHABLE_CACHE.clear(); }` and drop the Map from the export block. Or extract `isHttpsReachable` + cache into its own module (`https-reachability.js`).

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -283,12 +283,39 @@ function toggleWWWHostname(hostname) {
   return hostname.startsWith('www.') ? hostname.replace('www.', '') : `www.${hostname}`;
 }
 
+const HTTPS_REACHABLE_CACHE = new Map();
+const HTTPS_REACHABLE_TTL_MS = 60 * 60 * 1000; // 1 hour — SSL state changes on the order of months
```
solaris007: **Important 2:** Symmetric 1h TTL for `result: true` and `result: false` caches negative results too long. The comment is right that SSL state changes on the order of months for stable certs - but a transient DNS failure, 5s timeout on a slow site, or brief WAF block sticks a hostname as `result: false` for an hour. For the broken-cert recovery case (which the cache exists to handle), we want quick self-healing. Use asymmetric TTLs: keep 1h for success, use 5-10 min for failure.

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -283,12 +283,39 @@ function toggleWWWHostname(hostname) {
...
+
+async function isHttpsReachable(hostname, log) {
+  const cached = HTTPS_REACHABLE_CACHE.get(hostname);
+  if (cached && Date.now() < cached.expiresAt) {
```
solaris007, line 354: **Important 4:** The cache expiry branch (`Date.now() < cached.expiresAt` evaluating false) is never exercised by any test. All cache-HIT tests use a future `expiresAt`. If `HTTPS_REACHABLE_TTL_MS` were a typo (negative or wrong magnitude), no test catches it. Add a test that pre-populates an expired entry (`expiresAt: Date.now() - 1000`), sets `nock(...).head("/").reply(200)`, and asserts a fresh HEAD fires.

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -283,12 +283,39 @@ function toggleWWWHostname(hostname) {
...
+    return cached.result;
```
solaris007, line 359: **Important 5:** Cached `result: false` returns silently here with no log. If a site recovered from a brief SSL hiccup, the cache may still say `unreachable` for up to 59 minutes - same user-visible symptom as the krisshop bug this PR fixes, but caused by stale cache instead of bad RUM. Add a debug log on negative-cache hits so operators can correlate user complaints with stale cache decisions, e.g. `log.debug("isHttpsReachable: ${hostname} cached as unreachable, expires in Xs")`.

packages/spacecat-shared-utils/test/url-helpers.test.js
```diff
@@ -861,10 +871,88 @@ describe('URL Utility Functions', () => {
...
+
+      site.getBaseURL.returns('https://cachetest.com');
+      rumApiClient.retrieveDomainkey.withArgs(toggledHostname).resolves('domain-key');
+      // No nock needed — cache prevents the HTTP request entirely
+
+      const result = await wwwUrlResolver(site, rumApiClient, log);
+
+      expect(result).to.equal(toggledHostname);
+      expect(HTTPS_REACHABLE_CACHE.get(toggledHostname).result).to.be.true;
     });
```
solaris007, line 1151: **Important 3:** Only the `result: true` HIT path is tested. Add a test that pre-populates `HTTPS_REACHABLE_CACHE.set("cached-negative.com", { result: false, expiresAt: Date.now() + 60000 })` and asserts the resolver falls through without firing a HEAD request. The stale-negative-cache branch is the most likely production-incident path - worth a dedicated regression test.

solaris007 (changes requested):
Hey @tkotthakota-adobe,

Thanks for the quick turnaround. The scope is now clean (data-access changes reverted), and the two substantive prior findings - error logging on `isHttpsReachable` and caching to amortize the 5s probe - are genuinely fixed, not papered over. The cache implementation introduces a few patterns worth tightening before this lands.

## Strengths

- **Prior Important 3 (catch swallowing errors) is genuinely fixed** (url-helpers.js:304-306). `isHttpsReachable` takes `log`, calls `log.warn` with structured `{ fn, hostname, errorName, errorMessage }`, and tests at url-helpers.test.js:914-941 assert both TLS-failure and network-timeout paths produce the structured fields.
- **Prior Important 6 (no cache) is genuinely fixed** (url-helpers.js:286-308). Module-scoped Map with 1h TTL, both success and failure cached, cache-hit branch covered by a test that seeds the cache directly. Pre-populating is cleaner than stubbing twice with conflicting responses.
- **Scope is now coherent.** Prior Important 2 (bundled changes) resolved - title, description, and diff all describe the same change now.
- **Sensible probe defaults.** `redirect: 'manual'` is the right SSRF-defense choice for a TLS reachability probe; the response is never followed and never inspected, so a 3xx pointing to internal infrastructure cannot be chained into anything. `AbortSignal.timeout(5000)` bounds the latency cost.
- **Krisshop scenario tested end-to-end** at url-helpers.test.js:881-895, plus the symmetric "toggled has bad SSL, original has no RUM" case at url-helpers.test.js:897-912.
- **Test hygiene additions** (url-helpers.test.js:688-690): `HTTPS_REACHABLE_CACHE.clear()` and `nock.cleanAll()` in `afterEach`. Good discipline for a function with state.

## Issues

### Important (Should Fix)

**1. Cache exposure / lifecycle: cache exported as a mutable public API to enable a test-only need**

`packages/spacecat-shared-utils/src/url-helpers.js:412`

`HTTPS_REACHABLE_CACHE` is exported from the package and becomes part of `spacecat-shared-utils`'s published surface. The only consumers are tests using `.clear()`, `.set()`, and `.get()` for isolation and pre-population. Once a published version ships with this export, external consumers can clear it, populate it, inspect it, and the shape `Map<string, { result, expiresAt }>` becomes a contract. Any future caching change (LRU, sized cap, distributed cache, per-entry TTL, swap to a class) becomes a breaking change.

Separately, this introduces module-scoped mutable state in `url-helpers.js`, which has been pure-functional / single-shot fetch utilities up to now. The next helper that wants memoization will reach for the same pattern.

Fix (pick one):
- Replace the export with a narrow test helper: `export function resetHttpsReachableCacheForTests() { HTTPS_REACHABLE_CACHE.clear(); }`. Update `afterEach` to call the helper. Drop `HTTPS_REACHABLE_CACHE` from the export block.
- Or extract `isHttpsReachable` + its cache into its own module (`https-reachability.js`) so the stateful boundary is locally explicit.

**2. Symmetric 1-hour TTL caches negative results too long**

`packages/spacecat-shared-utils/src/url-helpers.js:287`

The comment "SSL state changes on the order of months" is true for stable certs but not for the case the cache primarily helps - the broken case where we actively want quick recovery. A `result: false` cached for 1 hour sticks a hostname as "broken" based on a single probe. A transient DNS failure, a 5-second timeout on a slow-but-healthy site, or a brief WAF block will mark the hostname unreachable and cause `wwwUrlResolver` to fall through to the alternate hostname for the next hour.

Fix: asymmetric TTLs. Keep 1h for `result: true`, use a shorter TTL (5-10 min) for `result: false` so transients self-heal while the common case still amortizes. Two constants instead of one.

**3. Cache HIT with `result: false` is not directly tested**

`packages/spacecat-shared-utils/test/url-helpers.test.js:956`

The existing cache test only verifies HIT with `result: true`. The stale-negative-cache branch, the one most likely to cause user-visible incidents in production, has no regression test.

Fix: add a test that pre-populates `HTTPS_REACHABLE_CACHE.set('cached-negative.com', { result: false, expiresAt: Date.now() + 60000 })`, then asserts the resolver falls through without firing a HEAD request.

**4. Cache expiry branch (`Date.now() < cached.expiresAt` false) is never exercised**

`packages/spacecat-shared-utils/src/url-helpers.js:291`

All cache HIT tests use a future `expiresAt`. If the TTL constant were a typo (negative, or off by orders of magnitude), no test catches it.

Fix: add a test that pre-populates an expired entry (`expiresAt: Date.now() - 1000`), sets up `nock(...).head('/').reply(200)`, and asserts a fresh HEAD fires.

**5. Stale `result: false` cache yields silent broken behavior**

`packages/spacecat-shared-utils/src/url-helpers.js:292`

When the cache returns a cached `false`, nothing is logged. If a site recovered from a brief SSL hiccup but the cache still says "unreachable" for up to 59 minutes, the user sees the wrong hostname selected with zero diagnostic signal - identical symptom shape to the krisshop bug this PR set out to fix, but caused by stale cache instead of bad RUM.

Fix: add a debug log when returning a cached negative value, e.g., `log.debug('isHttpsReachable: ${hostname} cached as unreachable, expires in Xs')`. Positive cache hits are less interesting; log only the negative case to keep volume low.

### Minor (Nice to Have)

**6. `isHttpsReachable` lacks JSDoc and its name overpromises**

`packages/spacecat-shared-utils/src/url-helpers.js:289`

Every other function in this file has JSDoc. The name "isHttpsReachable" reads as "this URL is alive and usable," but the implementation accepts any HTTP response (3xx with `redirect: 'manual'`, 4xx, 5xx) as `true` - only TLS or network errors return `false`. For the krisshop fix this semantic is correct, but a future contributor will read the name and assume more.

Fix: add a JSDoc clarifying the contract (returns true if the TLS handshake succeeds regardless of HTTP status; 4xx/5xx are treated as reachable) and note that `log` must support `warn()`.

**7. Happy-path tests do not assert the cache is populated after a successful HEAD**

`packages/spacecat-shared-utils/test/url-helpers.test.js` (around lines 767-789)

The write side-effect is only verified via the pre-populated HIT test. A regression that broke `HTTPS_REACHABLE_CACHE.set(...)` after a successful fetch would not be caught.

Fix: in one happy-path test, after the call, add `expect(HTTPS_REACHABLE_CACHE.get('www.example.com')?.result).to.be.true`.

**8. Cache-hit test does not isolate the network**

`packages/spacecat-shared-utils/test/url-helpers.test.js:943-956`

The test relies on the cache returning a value before fetch fires, with no `nock.disableNetConnect()`. If the cache logic broke and `fetch` actually fired, the result would depend on the test runner's network access.

Fix: add `nock.disableNetConnect()` in `beforeEach` and `nock.enableNetConnect()` in `afterEach` for the `wwwUrlResolver` describe block.

## Recommendations

- **One-line comment on `redirect: 'manual'`** (url-helpers.js:298). Worth saying "we're probing the TLS handshake; following redirects could mask a failure or land on a different host."
- **In-flight Promise coalescing** for concurrent uncached calls on the same hostname. Not a correctness bug, but in a cold-Lambda burst (fan-out of N audits for the same site) you could see N duplicate probes. Trivial fix is a `Map<hostname, Promise>`. Skip unless metrics show real duplication.
- **Soft upper bound on the cache.** The customer-hostname set is bounded today, but a future caller passing dynamically-constructed hostnames could grow the Map unboundedly. Cheap insurance is a 1000-entry FIFO cap; not urgent.
- **One-line comment on the fallback path** at url-helpers.js:347-349. The implicit tradeoff (trust the customer-registered URL more than the RUM-derived alternative) is defensible but invisible. Helps a future maintainer who tries to "fix" the asymmetry without understanding it.

- **`wwwUrlResolver` is forked across four implementations** (`spacecat-shared-utils`, `spacecat-audit-worker/src/common/base-audit.js`, `spacecat-reporting-worker/src/support/utils.js`, `spacecat-api-service/src/support/utils.js`). The bug class this PR addresses can surface on any of the local copies. Worth a follow-up tracking issue to consolidate to a single source of truth.
- **SSRF defense in depth on `isHttpsReachable`** (formerly Minor 10 of prior review). Reaffirmed out of scope. Trust boundary is admin-controlled `site.getBaseURL()`, gated by RUM key lookup; today's Lambdas are not VPC-attached. Worth a defensive hostname guard before MCT / private-connectivity rollout changes the network posture, but separate from this PR.
- **`package-lock.json` `peer: true` metadata additions** on ~15 transitive deps with no version or package changes. Matches normal npm normalization when a developer's local npm version differs from the prior lockfile's. Not a finding; a one-line note in the PR description confirming the source would help future readers.

## Assessment

The bug fix is correct, the prior substantive concerns are well-addressed, and the diff is tight. The cache exposure (item 1) and symmetric TTL (item 2) are cheaper to fix now than after the package version ships. The three test gaps (items 3, 4, 5) are small additions on top of the existing test patterns and close real regression-detection and observability gaps for the cache logic this PR introduces.

## Next Steps

1. Address items 1 and 2 (cache exposure, asymmetric TTL) - they shape the public API and the operational behavior.
2. Land the three test additions (items 3, 4, 5) - they close cache-logic regression-detection and observability gaps.
3. Item 6 (JSDoc) is a low-cost addition while the file is open.
4. Items 7, 8 and the recommendations can ship in a follow-up if time-pressured.

packages/spacecat-shared-utils/test/url-helpers.test.js
```diff
@@ -994,11 +1004,132 @@ describe('URL Utility Functions', () => {
...
+      nock('https://www.expired-cache.com').head('/').replyWithError('ECONNREFUSED');
+      site.getBaseURL.returns('https://expired-cache.com');
+      rumApiClient.retrieveDomainkey.withArgs(toggledHostname).resolves('domain-key');
+      rumApiClient.retrieveDomainkey.withArgs('expired-cache.com').resolves('domain-key');
+      await wwwUrlResolver(site, rumApiClient, log);
+
+      // Manually expire the cache entry
+      // Access internal cache via a fresh negative-TTL probe is tricky,
+      // so we use resetHttpsReachableCacheForTests to simulate expiry
+      resetHttpsReachableCacheForTests();
```
solaris007: **Important: this test does not exercise the expiry branch it names (prior finding, attempted).**

`resetHttpsReachableCacheForTests()` here runs `HTTPS_REACHABLE_CACHE.clear()`, so on the next `wwwUrlResolver` call `HTTPS_REACHABLE_CACHE.get(hostname)` returns `undefined` and `if (cached && Date.now() < cached.expiresAt)` (url-helpers.js:347) short-circuits on the first operand. The `Date.now() < cached.expiresAt` comparison is never evaluated with a populated entry, so the expiry path (entry present but stale) stays uncovered - this test actually exercises the cache-miss path, which every other test already hits. The test name promises coverage the suite does not provide.

Fix without reintroducing the cache export: drive a fake `Date` clock past the 5min negative TTL instead of clearing, so `cached` is truthy and stale:

```js
const clock = sandbox.useFakeTimers({ now: Date.now(), toFake: ['Date'] });
nock('https://www.expired-cache.com').head('/').replyWithError('ECONNREFUSED');
const first = await wwwUrlResolver(site, rumApiClient, log);
expect(first).to.equal('expired-cache.com');
clock.tick(5 * 60 * 1000 + 1);
nock('https://www.expired-cache.com').head('/').reply(200);
const result = await wwwUrlResolver(site, rumApiClient, log);
expect(result).to.equal('www.expired-cache.com');
clock.restore();
```

Minimum alternative: rename the test to reflect that it re-probes after a cache clear, so the suite stops claiming expiry coverage.

solaris007 (changes requested):
Hey @tkotthakota-adobe,

Thanks for the thorough pass on the last review. Seven of the eight prior findings are genuinely closed, with behavior-level tests rather than cosmetic edits, and the diff stays tightly scoped to the two utils files. One prior item (the cache-expiry test) was attempted but does not yet cover the branch it names, and I found one new in-scope item worth a look.

## Strengths

- **Mutable cache export removed (previously flagged).** `HTTPS_REACHABLE_CACHE` is gone from the export block; the only new export is `resetHttpsReachableCacheForTests()` (url-helpers.js:382), an intent-revealing test seam. The module-internal Map is no longer reachable or mutable by consumers - exactly the recommended fix.
- **Asymmetric TTLs (previously flagged).** Positive 1h / negative 5min (url-helpers.js:330-332), so a transient TLS or network blip self-heals in 5 minutes instead of being pinned for an hour. Right direction for a probe whose purpose is detecting bad certs.
- **Negative cache-hit now logged and directly tested (previously flagged).** The cache-hit branch emits `log.debug('... cached as unreachable, expires in Ns')` (url-helpers.js:350), and the new test at url-helpers.test.js:1094 proves the fall-through plus the debug log under `nock.disableNetConnect()`. The stale-negative path is no longer silent.
- **JSDoc added; semantics clarified (previously flagged).** The doc block on `isHttpsReachable` is precise: TLS-handshake-only, 4xx/5xx treated as reachable, `redirect:'manual'` rationale, asymmetric caching, logger contract. This prevents a future maintainer from "fixing" the deliberate 4xx-is-reachable behavior.
- **Cache-hit test is now behavior-level and network-isolated (previously flagged).** The white-box `.get(...).result` assertion is gone (it had to go with the export); the rewritten test populates via a real nocked HEAD then proves the cache short-circuits the network with `disableNetConnect`/`enableNetConnect`. Stronger than the prior white-box peek.
- **`redirect:'manual'` is correct and documented.** Probing the specific hostname's TLS handshake without following a redirect to a different host is exactly right - an auto-followed redirect would mask the failure this probe exists to detect, and it also closes the redirect-based SSRF escalation path (the response is never followed or read).
- **Fallback comment added (previously flagged recommendation).** `wwwUrlResolver` now documents why the fallback prefers the customer-registered www form, removing the "bug or intentional?" question.
- **Scope and lockfile are clean.** The accidental data-access config-model rewrite stays reverted. The package-lock.json change is benign: three internal `@adobe/*` workspace version bumps (cloud-manager-client 1.2.2 to 1.3.1, data-access 3.64.0 to 3.65.0, html-analyzer 1.2.11 to 1.2.12) pulled in by the merge-main commits, with no `resolved`/`integrity`/dependency changes.

## Issues

### Important (Should Fix)

**The "expired cache" test does not exercise the expiry branch it names (previously flagged, attempted).**

packages/spacecat-shared-utils/test/url-helpers.test.js:1112

The test "should re-probe when cached entry has expired" calls `resetHttpsReachableCacheForTests()` (test:1125), which runs `HTTPS_REACHABLE_CACHE.clear()`. On the second `wwwUrlResolver` call, `HTTPS_REACHABLE_CACHE.get(hostname)` returns `undefined`, so `if (cached && Date.now() < cached.expiresAt)` (url-helpers.js:347) short-circuits on `cached` being falsy - the `Date.now() < cached.expiresAt` comparison never runs. The test exercises the cache-miss path (already covered implicitly by every other test, none of which pre-populate), not the expiry path (entry present but stale). The inline comment at test:1123-1124 is candid that real expiry was "tricky" to trigger.

Why it matters: the expiry comparison is the mechanism that makes the asymmetric-TTL design work. It is the one piece of cache logic with no test, and the line a future refactor is most likely to break silently (flip the operator, drop the timestamp check, and stale negatives get served forever). The test name asserts a guarantee the suite does not provide, which is false confidence for the next maintainer. CI is green and the repo's branch-coverage threshold tolerates the uncovered branch, so this is not a behavioral bug today, but the prior review rated it Important and the attempt does not close it.

How to fix (no cache export needed): drive the clock with sinon fake timers (already used in this file) instead of clearing the Map. Fake only `Date` so real timers for `AbortSignal.timeout` and the async runtime are untouched:

```js
const clock = sandbox.useFakeTimers({ now: Date.now(), toFake: ['Date'] });
// 1) populate a negative entry (expiresAt = now + 5min)
nock('https://www.expired-cache.com').head('/').replyWithError('ECONNREFUSED');
const first = await wwwUrlResolver(site, rumApiClient, log);
expect(first).to.equal('expired-cache.com'); // negative -> falls through
// 2) advance past the 5min negative TTL so the entry is stale-but-present
clock.tick(5 * 60 * 1000 + 1);
// 3) cached is truthy AND Date.now() >= expiresAt -> re-probe
nock('https://www.expired-cache.com').head('/').reply(200);
const result = await wwwUrlResolver(site, rumApiClient, log);
expect(result).to.equal('www.expired-cache.com');
clock.restore();
```

If you would rather not drive the clock, the minimum is to rename the test to describe what it does (re-probes after the cache is cleared) so the suite stops claiming expiry coverage it lacks.

### Minor (Nice to Have)

**New HTTPS probe sink bypasses the package's own SSRF guard.**

packages/spacecat-shared-utils/src/url-helpers.js:345

`isHttpsReachable` does `fetch(\`https://${hostname}\`, ...)` (url-helpers.js:357) with no call to `isNonPublicHostname(hostname)`, even though that guard lives in the same package (network-policy.js), is already imported at url-helpers.js:17, and is applied to the sibling sink `resolveCanonicalUrl` (url-helpers.js:169). network-policy.js's own header says the guard is used by its callers "so any fix is applied once" - this PR adds a third outbound sink and breaks that single-choke-point invariant.

Exploitability is low and unchanged from the prior reviews' assessment: `hostname` derives from the admin-registered `site.getBaseURL()`, the call is gated behind a successful `rumApiClient.retrieveDomainkey()` lookup, and `redirect:'manual'` means a malicious target's response is never followed or read (the probe yields at most a one-bit reachability oracle). What changed since the prior SSRF deferral is that an in-repo guard now exists, so the "no guard available" half of the deferral rationale is stale and reusing it is a one-line, in-scope consistency fix:

```js
async function isHttpsReachable(hostname, log) {
  if (isNonPublicHostname(hostname)) {
    log.warn(`isHttpsReachable: ${hostname} is non-public; refusing probe`, { fn: 'isHttpsReachable', hostname });
    return false;
  }
  // existing cache + fetch logic
}
```

This mirrors the existing `resolveCanonicalUrl` guard and needs one matching test. If the team prefers to keep this as tracked defense-in-depth given the unchanged threat model, that is defensible, but then drop the "no guard exists yet" reasoning from the deferral note and amend network-policy.js's "applied once" claim, which is now inaccurate.

## Recommendations

- Even after the fake-clock fix, assert the re-probe actually fired in the expiry test (for example `expect(rumApiClient.retrieveDomainkey).to.have.been.calledWith('www.expired-cache.com')`, and rely on the fresh `nock(...).reply(200)` interceptor being consumed). This distinguishes "re-probed and succeeded" from "served a stale value that happened to match."
- `url-helpers.js` was a pure, synchronous string-transformation module; this probe plus the pre-existing `resolveCanonicalUrl` now make it open sockets and hold process-lifetime mutable state. Acceptable for one or two functions, but worth resisting further network or stateful additions here and, on the next touch, considering a dedicated `url-resolvers.js` so the rest of the file keeps its cheap-and-pure contract.
- Consider a one-line comment on the cache stating its intended lifetime (per-process, relies on Lambda container churn for eviction), so a future consumer in a long-lived process understands it was never designed to be bounded. The prior soft-cache-cap suggestion remains correctly declined for the current bounded-hostname workload.

- `wwwUrlResolver` is forked across multiple repos (spacecat-shared-utils here, plus copies in spacecat-audit-worker, spacecat-reporting-worker, spacecat-api-service, and spacecat-import-worker). This fix lands in the library only; the same onboarding-failure class can recur in any consumer that resolves via its own copy. A consolidation ticket to make spacecat-shared the single source of truth would retire the drift. Cross-repo, not fixable in this PR.
- The fallback path in `wwwUrlResolver` returns the www-prefixed hostname without its own HTTPS check (now documented as deliberate). Not a regression in this diff; noting once for the parent ticket.
- In-flight probe coalescing for concurrent uncached calls on the same hostname remains deferred and optional given 5s timeouts and low per-container hostname cardinality.

## Assessment

Ready to merge? With fixes.

This is a tight, well-executed re-review pass. The prior architectural objection (mutable shared-state export) is properly resolved, the asymmetric TTL and fail-safe fallback are sound, and observability for the stale-negative path is now in place. The one carried-over item is the expiry-branch test: the attempt clears the cache rather than expiring an entry, so the mechanism behind the asymmetric-TTL design is still undefended and the test name overpromises. The fix is small and self-contained. The new SSRF-guard consistency gap is Minor and optional. CI is green, but green here only proves the cache-miss path, not the expiry path.

## Next Steps

1. Close the expiry-branch test with a fake clock (or, at minimum, rename it) so the suite stops claiming coverage it lacks.
2. Decide on the SSRF-guard reuse: wire `isHttpsReachable` through `isNonPublicHostname` (one line plus a test), or keep it tracked as deferred and correct the now-stale "applied once" note.
3. The recommendations and out-of-scope items are optional follow-ups.

PR #1605, opened 2026-05-14
feat(brand-client): add BrandGovernanceClient for Brand Governance Agent API (LLMO-4920)

anagarwa:
## Summary

- Add `BrandGovernanceClient` as a named export in `spacecat-shared-brand-client` alongside the existing `BrandClient`
- New client resolves brand guidelines by site URL via two sequential Brand Governance Agent API calls:
  1. `GET /api/v1/brands/from-url?url={siteBaseUrl}` — resolves brand by domain; returns `null` on 404 so callers fall back gracefully
  2. `GET /api/v1/brands/{id}/checks?status=ACTIVE&type=BRAND&pageSize=100` — fetches checks, filters to `COPY`-scoped only, maps to `{ name, text }`
- Adds TypeScript declarations for the new class in `index.d.ts`
- `BrandClient` (Brand Publish) is unchanged — fully backwards compatible

## New env vars required (in consuming services)
- `BRAND_GOV_API_BASE_URL` — Brand Governance Agent base URL
- `BRAND_GOV_API_KEY` — `x-api-key` header value
- `BRAND_GOV_IMS_CLIENT_ID`, `BRAND_GOV_IMS_CLIENT_CODE`, `BRAND_GOV_IMS_CLIENT_SECRET` — service account IMS credentials
- `IMS_HOST` — reuses the existing standard IMS host env var

## Jira
[LLMO-4920](https://jira.corp.adobe.com/browse/LLMO-4920) — prerequisite for [LLMO-4918](https://jira.corp.adobe.com/browse/LLMO-4918)

## Test plan
- [ ] Unit tests to be added after team review of the approach (per team agreement)
- [ ] Manually verify with `scripts/test-brand-governance-api.js` against stage: brand lookup by URL returns COPY-scoped checks mapped to `{ name, text }`
- [ ] Verify `null` is returned for a domain not registered in Brand Governance (404 path)
- [ ] Verify `BrandClient` (existing Brand Publish) behaviour is unchanged

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/spacecat-shared-brand-client/src/brand-governance-client.js
+++ b/packages/spacecat-shared-brand-client/src/brand-governance-client.js
@@ -17,8 +17,10 @@ import { ImsClient } from '@adobe/spacecat-shared-ims-client';
 const HTTP_BAD_REQUEST = 400;
 const HTTP_NOT_FOUND = 404;
 
+const CHECKS_PAGE_SIZE = 100;
+
 const API_GET_BRAND_FROM_URL = (url) => `/api/v1/brands/from-url?url=${encodeURIComponent(url)}`;
-const API_GET_BRAND_CHECKS = (brandId) => `/api/v1/brands/${brandId}/checks?status=ACTIVE&type=BRAND&pageSize=100`;
+const API_GET_BRAND_CHECKS = (brandId, page) => `/api/v1/brands/${brandId}/checks?status=ACTIVE&type=BRAND&scope=COPY&pageSize=${CHECKS_PAGE_SIZE}&page=${page}`;
 
 /**
  * Client for the Adobe Brand Governance Agent API.
@@ -127,22 +129,31 @@ export class BrandGovernanceClient {
     }
     const brand = await brandResponse.json();
 
-    const checksResponse = await fetch(
-      `${this.apiBaseUrl}${API_GET_BRAND_CHECKS(brand.id)}`,
-      { headers },
-    );
-    if (!checksResponse.ok) {
-      throw this.#createError(
-        `Error fetching brand checks for brand ${brand.id}: ${checksResponse.status}`,
-        checksResponse.status,
+    const allChecks = [];
+    let page = 1;
+    do {
+      // eslint-disable-next-line no-await-in-loop
+      const checksResponse = await fetch(
+        `${this.apiBaseUrl}${API_GET_BRAND_CHECKS(brand.id, page)}`,
+        { headers },
       );
-    }
-    const checksResult = await checksResponse.json();
-    const checks = checksResult.data || [];
+      if (!checksResponse.ok) {
+        throw this.#createError(
+          `Error fetching brand checks for brand ${brand.id}: ${checksResponse.status}`,
+          checksResponse.status,
+        );
+      }
+      // eslint-disable-next-line no-await-in-loop
+      const { data = [] } = await checksResponse.json();
+      allChecks.push(...data);
+      if (data.length < CHECKS_PAGE_SIZE) {
+        break;
+      }
+      page += 1;
+    // eslint-disable-next-line no-constant-condition
+    } while (true);
 
-    const guidelines = checks
-      .filter((check) => Array.isArray(check.scopes) && check.scopes.includes('COPY'))
-      .map((check) => ({ name: check.name, text: check.rule }));
+    const guidelines = allChecks.map((check) => ({ name: check.name, text: check.rule }));
 
     return {
       id: brand.id,
--- a/packages/spacecat-shared-brand-client/src/brand-governance-client.js
+++ b/packages/spacecat-shared-brand-client/src/brand-governance-client.js
@@ -18,6 +18,7 @@ const HTTP_BAD_REQUEST = 400;
 const HTTP_NOT_FOUND = 404;
 
 const CHECKS_PAGE_SIZE = 100;
+const MAX_PAGES = 20;
 
 const API_GET_BRAND_FROM_URL = (url) => `/api/v1/brands/from-url?url=${encodeURIComponent(url)}`;
 const API_GET_BRAND_CHECKS = (brandId, page) => `/api/v1/brands/${brandId}/checks?status=ACTIVE&type=BRAND&scope=COPY&pageSize=${CHECKS_PAGE_SIZE}&page=${page}`;
@@ -88,6 +89,12 @@ export class BrandGovernanceClient {
    * Fetches brand guidelines for a site URL from the Brand Governance Agent.
    * Returns null if the brand is not registered (404), allowing callers to fall back.
    *
+   * Org/tenant scoping for the domain lookup is enforced server-side by the Agent via the
+   * x-gw-ims-org-id request header. BrandResponse carries no org_id field, so the
+   * client-side ownership check performed by BrandClient is not possible here. Note that
+   * the imsOrgId field in the returned object is the requested org echoed for shape-parity
+   * with BrandClient — it is not a verified owner of the resolved brand.
+   *
    * @param {string} siteBaseUrl - The site's base URL to resolve brand by domain
    * @param {string} imsOrgId - The IMS org ID sent as x-gw-ims-org-id header
    * @param {object} imsConfig - IMS auth config: { host, clientId, clientCode, clientSecret }
@@ -153,6 +160,12 @@ export class BrandGovernanceClient {
         break;
       }
       page += 1;
+      if (page > MAX_PAGES) {
+        throw this.#createError(
+          `Brand checks pagination exceeded ${MAX_PAGES} pages for brand ${brand.id}`,
+          HTTP_BAD_REQUEST,
+        );
+      }
     // eslint-disable-next-line no-constant-condition
     } while (true);
 
--- a/packages/spacecat-shared-brand-client/src/brand-governance-client.js
+++ b/packages/spacecat-shared-brand-client/src/brand-governance-client.js
@@ -27,15 +27,15 @@ const API_GET_BRAND_CHECKS = (brandId) => `/api/v1/brands/${brandId}/checks?stat
  */
 export class BrandGovernanceClient {
   static createFrom(context) {
+    if (context.brandGovernanceClient) {
+      return context.brandGovernanceClient;
+    }
+
     const { env, log = console } = context;
     const {
       BRAND_GOV_API_BASE_URL: apiBaseUrl,
       BRAND_GOV_API_KEY: apiKey,
     } = env;
-
-    if (context.brandGovernanceClient) {
-      return context.brandGovernanceClient;
-    }
 
     const client = new BrandGovernanceClient({ apiBaseUrl, apiKey }, log);
     context.brandGovernanceClient = client;
--- a/packages/spacecat-shared-brand-client/src/brand-governance-client.js
+++ b/packages/spacecat-shared-brand-client/src/brand-governance-client.js
@@ -75,7 +75,11 @@ export class BrandGovernanceClient {
     };
     const imsClient = ImsClient.createFrom(imsContext);
     const response = await imsClient.getServiceAccessToken();
-    return response?.access_token;
+    const token = response?.access_token;
+    if (!hasText(token)) {
+      throw this.#createError('Failed to obtain IMS access token', HTTP_BAD_REQUEST);
+    }
+    return token;
   }
 
   /**
--- a/packages/spacecat-shared-brand-client/src/brand-governance-client.js
+++ b/packages/spacecat-shared-brand-client/src/brand-governance-client.js
@@ -1,0 +1,158 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+import {
+  isValidIMSOrgId, hasText, isValidUrl, tracingFetch as fetch,
+} from '@adobe/spacecat-shared-utils';
+import { ImsClient } from '@adobe/spacecat-shared-ims-client';
+
+const HTTP_BAD_REQUEST = 400;
+const HTTP_NOT_FOUND = 404;
+
+const API_GET_BRAND_FROM_URL = (url) => `/api/v1/brands/from-url?url=${encodeURIComponent(url)}`;
+const API_GET_BRAND_CHECKS = (brandId) => `/api/v1/brands/${brandId}/checks?status=ACTIVE&type=BRAND&pageSize=100`;
+
+/**
+ * Client for the Adobe Brand Governance Agent API.
+ * Resolves brand guidelines by site URL using COPY-scoped checks.
+ * Returns null when the brand is not registered — caller falls back to Brand Publish.
+ */
+export class BrandGovernanceClient {
+  static createFrom(context) {
+    const { env, log = console } = context;
+    const {
+      BRAND_GOV_API_BASE_URL: apiBaseUrl,
+      BRAND_GOV_API_KEY: apiKey,
+    } = env;
+
+    if (context.brandGovernanceClient) {
+      return context.brandGovernanceClient;
+    }
+
+    const client = new BrandGovernanceClient({ apiBaseUrl, apiKey }, log);
+    context.brandGovernanceClient = client;
+    return client;
+  }
+
+  constructor({ apiBaseUrl, apiKey }, log) {
+    this.log = log;
+    if (!isValidUrl(apiBaseUrl)) {
+      throw this.#createError(`Invalid Brand Governance API Base URL: ${apiBaseUrl}`, HTTP_BAD_REQUEST);
+    }
+    if (!hasText(apiKey)) {
+      throw this.#createError(`Invalid Brand Governance API Key: ${apiKey}`, HTTP_BAD_REQUEST);
+    }
+    this.apiBaseUrl = apiBaseUrl;
+    this.apiKey = apiKey;
+    this.serviceAccessToken = null;
+  }
+
+  #createError(message, status) {
+    const error = Object.assign(new Error(message), { status });
+    this.log.error(error.message);
+    return error;
+  }
+
+  async #getImsAccessToken(imsConfig) {
+    if (this.serviceAccessToken) {
+      return this.serviceAccessToken;
+    }
+    const {
+      host, clientId, clientCode, clientSecret,
+    } = imsConfig;
+    const imsContext = {
+      env: {
+        IMS_HOST: host,
+        IMS_CLIENT_ID: clientId,
+        IMS_CLIENT_CODE: clientCode,
+        IMS_CLIENT_SECRET: clientSecret,
+      },
+      log: this.log,
+    };
+    const imsClient = ImsClient.createFrom(imsContext);
+    const response = await imsClient.getServiceAccessToken();
+    this.serviceAccessToken = response?.access_token;
+    return this.serviceAccessToken;
+  }
+
+  /**
+   * Fetches brand guidelines for a site URL from the Brand Governance Agent.
+   * Returns null if the brand is not registered (404), allowing callers to fall back.
+   *
+   * @param {string} siteBaseUrl - The site's base URL to resolve brand by domain
+   * @param {string} imsOrgId - The IMS org ID sent as x-gw-ims-org-id header
+   * @param {object} imsConfig - IMS auth config: { host, clientId, clientCode, clientSecret }
+   * @returns {Promise<object|null>} Brand guidelines or null if not registered
+   */
+  async getBrandGuidelinesForUrl(siteBaseUrl, imsOrgId, imsConfig) {
+    if (!isValidUrl(siteBaseUrl)) {
+      throw this.#createError(`Invalid site base URL: ${siteBaseUrl}`, HTTP_BAD_REQUEST);
+    }
+    if (!isValidIMSOrgId(imsOrgId)) {
+      throw this.#createError(`Invalid IMS Org ID: ${imsOrgId}`, HTTP_BAD_REQUEST);
+    }
+    const {
+      host, clientId, clientCode, clientSecret,
+    } = imsConfig;
+    if (!hasText(host) || !hasText(clientId) || !hasText(clientCode) || !hasText(clientSecret)) {
+      throw this.#createError(`Invalid IMS Config: ${JSON.stringify(imsConfig)}`, HTTP_BAD_REQUEST);
+    }
+
+    const imsAccessToken = await this.#getImsAccessToken(imsConfig);
+    const headers = {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${imsAccessToken}`,
+      'x-api-key': this.apiKey,
+      'x-gw-ims-org-id': imsOrgId,
+    };
+
+    const brandResponse = await fetch(
+      `${this.apiBaseUrl}${API_GET_BRAND_FROM_URL(siteBaseUrl)}`,
+      { headers },
+    );
+    if (brandResponse.status === HTTP_NOT_FOUND) {
+      return null;
+    }
+    if (!brandResponse.ok) {
+      throw this.#createError(
+        `Error resolving brand for URL ${siteBaseUrl}: ${brandResponse.status}`,
+        brandResponse.status,
+      );
+    }
+    const brand = await brandResponse.json();
+
+    const checksResponse = await fetch(
+      `${this.apiBaseUrl}${API_GET_BRAND_CHECKS(brand.id)}`,
+      { headers },
+    );
+    if (!checksResponse.ok) {
+      throw this.#createError(
+        `Error fetching brand checks for brand ${brand.id}: ${checksResponse.status}`,
+        checksResponse.status,
+      );
+    }
+    const checksResult = await checksResponse.json();
+    const checks = checksResult.data || [];
+
+    const guidelines = checks
+      .filter((check) => Array.isArray(check.scopes) && check.scopes.includes('COPY'))
+      .map((check) => ({ name: check.name, text: check.rule }));
+
+    return {
+      id: brand.id,
+      name: brand.name,
+      imsOrgId,
+      createdAt: brand.createdAt,
+      updatedAt: brand.updatedAt,
+      guidelines,
+    };
+  }
+}
--- a/packages/spacecat-shared-brand-client/src/brand-governance-client.js
+++ b/packages/spacecat-shared-brand-client/src/brand-governance-client.js
@@ -128,9 +128,6 @@ export class BrandGovernanceClient {
       );
     }
     const brand = await brandResponse.json();
-    // Org scoping is enforced server-side by the Brand Governance Agent via the
-    // x-gw-ims-org-id request header. Unlike BrandClient, the BrandResponse schema
-    // does not include an org_id field, so client-side org verification is not possible.
     if (!hasText(brand.id)) {
       throw this.#createError(`Brand resolved for URL ${siteBaseUrl} has no id`, HTTP_NOT_FOUND);
     }
--- a/packages/spacecat-shared-brand-client/src/brand-governance-client.js
+++ b/packages/spacecat-shared-brand-client/src/brand-governance-client.js
@@ -128,6 +128,12 @@ export class BrandGovernanceClient {
       );
     }
     const brand = await brandResponse.json();
+    // Org scoping is enforced server-side by the Brand Governance Agent via the
+    // x-gw-ims-org-id request header. Unlike BrandClient, the BrandResponse schema
+    // does not include an org_id field, so client-side org verification is not possible.
+    if (!hasText(brand.id)) {
+      throw this.#createError(`Brand resolved for URL ${siteBaseUrl} has no id`, HTTP_NOT_FOUND);
+    }
 
     const allChecks = [];
     let page = 1;
--- a/packages/spacecat-shared-brand-client/src/brand-governance-client.js
+++ b/packages/spacecat-shared-brand-client/src/brand-governance-client.js
@@ -52,7 +52,6 @@ export class BrandGovernanceClient {
     }
     this.apiBaseUrl = apiBaseUrl;
     this.apiKey = apiKey;
-    this.serviceAccessToken = null;
   }
 
   #createError(message, status) {
@@ -62,9 +61,6 @@ export class BrandGovernanceClient {
   }
 
   async #getImsAccessToken(imsConfig) {
-    if (this.serviceAccessToken) {
-      return this.serviceAccessToken;
-    }
     const {
       host, clientId, clientCode, clientSecret,
     } = imsConfig;
@@ -79,8 +75,7 @@ export class BrandGovernanceClient {
     };
     const imsClient = ImsClient.createFrom(imsContext);
     const response = await imsClient.getServiceAccessToken();
-    this.serviceAccessToken = response?.access_token;
-    return this.serviceAccessToken;
+    return response?.access_token;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


lucianfelix (changes requested):
This is not consistent with how Mysticat works -- data sources should be directly in BB

cc @lenaorobei

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,158 @@
...
+      throw this.#createError(`Invalid site base URL: ${siteBaseUrl}`, HTTP_BAD_REQUEST);
+    }
+    if (!isValidIMSOrgId(imsOrgId)) {
+      throw this.#createError(`Invalid IMS Org ID: ${imsOrgId}`, HTTP_BAD_REQUEST);
+    }
+    const {
+      host, clientId, clientCode, clientSecret,
+    } = imsConfig;
+    if (!hasText(host) || !hasText(clientId) || !hasText(clientCode) || !hasText(clientSecret)) {
+      throw this.#createError(`Invalid IMS Config: ${JSON.stringify(imsConfig)}`, HTTP_BAD_REQUEST);
```
BaDos: It looks like security issue: `imsConfig` contains `clientSecret` and this secret will be published in logs

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,158 @@
...
+
+  #createError(message, status) {
+    const error = Object.assign(new Error(message), { status });
+    this.log.error(error.message);
+    return error;
+  }
+
+  async #getImsAccessToken(imsConfig) {
+    if (this.serviceAccessToken) {
+      return this.serviceAccessToken;
```
BaDos: The IMS access token is cached forever on the instance. These tokens expire, so a warm Lambda will eventually start getting 401 errors

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,158 @@
...
+        IMS_HOST: host,
+        IMS_CLIENT_ID: clientId,
+        IMS_CLIENT_CODE: clientCode,
+        IMS_CLIENT_SECRET: clientSecret,
+      },
+      log: this.log,
+    };
+    const imsClient = ImsClient.createFrom(imsContext);
+    const response = await imsClient.getServiceAccessToken();
+    this.serviceAccessToken = response?.access_token;
```
BaDos: Should we check if a token is empty?

BaDos (changes requested):

solaris007 (dismissed):

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,158 @@
...
+import {
+  isValidIMSOrgId, hasText, isValidUrl, tracingFetch as fetch,
+} from '@adobe/spacecat-shared-utils';
+import { ImsClient } from '@adobe/spacecat-shared-ims-client';
+
+const HTTP_BAD_REQUEST = 400;
+const HTTP_NOT_FOUND = 404;
+
+const API_GET_BRAND_FROM_URL = (url) => `/api/v1/brands/from-url?url=${encodeURIComponent(url)}`;
+const API_GET_BRAND_CHECKS = (brandId) => `/api/v1/brands/${brandId}/checks?status=ACTIVE&type=BRAND&pageSize=100`;
```
solaris007: Important (item 5): hardcoded `pageSize=100` with no pagination. A brand with more than 100 active BRAND checks silently drops COPY guidelines beyond the first page - a 200 with an incomplete set, no error, no log. Paginate to exhaustion, or if 100 is a deliberate cap, log a WARN when `data.length === pageSize` and document it.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,158 @@
...
+      throw this.#createError(`Invalid site base URL: ${siteBaseUrl}`, HTTP_BAD_REQUEST);
+    }
+    if (!isValidIMSOrgId(imsOrgId)) {
+      throw this.#createError(`Invalid IMS Org ID: ${imsOrgId}`, HTTP_BAD_REQUEST);
+    }
+    const {
+      host, clientId, clientCode, clientSecret,
+    } = imsConfig;
+    if (!hasText(host) || !hasText(clientId) || !hasText(clientCode) || !hasText(clientSecret)) {
+      throw this.#createError(`Invalid IMS Config: ${JSON.stringify(imsConfig)}`, HTTP_BAD_REQUEST);
```
solaris007: Important (item 4): `JSON.stringify(imsConfig)` serializes `clientSecret` and `clientCode` into the log (via `#createError` -> `this.log.error`), and the consumer's `createErrorResponse` reflects `error.message` into the response body and the `x-error` header. `BrandGovernanceClient` is a public export, so a different caller passing a partial config would leak the secret. Report only which fields are missing, never the values. The same pattern is at line 51 (`Invalid Brand Governance API Key: ${apiKey}`).

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,158 @@
...
+      Authorization: `Bearer ${imsAccessToken}`,
+      'x-api-key': this.apiKey,
+      'x-gw-ims-org-id': imsOrgId,
+    };
+
+    const brandResponse = await fetch(
+      `${this.apiBaseUrl}${API_GET_BRAND_FROM_URL(siteBaseUrl)}`,
+      { headers },
+    );
+    if (brandResponse.status === HTTP_NOT_FOUND) {
```
solaris007, line 128: This 404-only null return is the root of the consumer's fallback gap (companion adobe/spacecat-api-service#2419, item 2): every other status or error throws instead of returning null, so the consumer 500s rather than falling back to Brand Publish. Consider returning null for non-404 failures too (or a typed result), or document that the consumer must catch and fall through.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,158 @@
...
+      );
+    }
+    const checksResult = await checksResponse.json();
+    const checks = checksResult.data || [];
+
+    const guidelines = checks
+      .filter((check) => Array.isArray(check.scopes) && check.scopes.includes('COPY'))
+      .map((check) => ({ name: check.name, text: check.rule }));
+
+    return {
```
solaris007, line 174: Critical (item 2): this response drops `toneOfVoice`, `coreValues`, `restrictions`, and `additionalGuidelines` that the `BrandGuidelines` shape advertises and that the consumer's downstream (Mystique's `format_brand_guidelines_to_markdown`) renders. Since adobe/spacecat-api-service#2419 makes this the primary source, BG-registered sites silently lose four of five sections with no error. Map COPY-scoped checks into the existing field structure, or align the contract with consumers and document the new shape.

solaris007 (commented):
Hey @anagarwa,

Reviewed in conjunction with its consumer adobe/spacecat-api-service#2419. The new client faithfully mirrors the existing `BrandClient` and the 404-as-null contract is clean, but it is red on its coverage gate and has several hardening items in the credential and resilience paths. Posting as a comment - my merge read is in the assessment at the end.

Note: CI (`Test`) is failing - see the coverage item.

### Strengths
- `BrandGovernanceClient` mirrors `BrandClient` closely (the `createFrom` context-caching factory, `#createError` helper, constructor validation), so it is immediately familiar.
- `BrandClient` stays the default export and `BrandGovernanceClient` is added as a named export (`index.js:17`); the `.d.ts` is kept in sync. Backwards compatible.
- The 404-returns-null contract is explicit and well-documented (`brand-governance-client.js:97-101`).
- Inputs are validated before egress, and a null `imsOrgId` fails closed with a 400.

### Issues

#### Critical (Must Fix)

**1. New client ships with zero tests against a 100% coverage gate (`Test` failing).** `brand-governance-client.js` (158 lines) has no test, and `test/index.test.js` imports only `BrandClient`, so package coverage falls to 67.35% against the 100% lines/branches/statements threshold and the build fails. The existing `getBrandGuidelines` test is a ready template: nock the IMS token endpoint, `from-url`, and `checks`, then assert the mapped shape plus the 404-null, non-ok, and invalid-input branches. Tests should land with this code since it becomes a published artifact other services consume.

**2. Response shape diverges from the contract consumers expect (`brand-governance-client.js:148-156`).** The client returns guidelines-only `{ id, name, imsOrgId, createdAt, updatedAt, guidelines: [{name,text}] }`, dropping `toneOfVoice`, `coreValues`, `restrictions`, and `additionalGuidelines` that the `BrandGuidelines` shape (`index.d.ts`) advertises and that the consumer's downstream (Mystique's `format_brand_guidelines_to_markdown`) renders. Since adobe/spacecat-api-service#2419 makes this the primary source, BG-registered sites silently lose four of five sections with no error. Fix: map COPY-scoped checks into the existing field structure, or align the contract with consumers and document the new shape.

#### Important (Should Fix)

**3. The cross-org verification the sibling client performs is missing (`brand-governance-client.js:136-156`).** `BrandClient.getBrandGuidelines` checks `if (result.org_id !== imsOrgId) throw NOT_FOUND` before returning. This client resolves a brand purely by domain (`from-url`), then echoes the caller-supplied `imsOrgId` back in the response without verifying the resolved brand belongs to that org. If the Brand Governance Agent does not strictly scope `from-url` by the `x-gw-ims-org-id` header, a domain that maps to a brand under a different org could return that org's COPY guidelines. Fix: mirror the sibling guard (compare the resolved brand's org to `imsOrgId`, return null on mismatch), or document that org scoping is delegated to the Agent so the dropped guard is a recorded decision.

**4. Secrets reach logs and the HTTP response on the invalid-config path (`brand-governance-client.js:106` and `:51`, via `#createError` -> `this.log.error` at `:57`).** `Invalid IMS Config: ${JSON.stringify(imsConfig)}` serializes `clientSecret` and `clientCode` into the log, and in the consumer `createErrorResponse` reflects `error.message` into the response body and `x-error` header. `Invalid Brand Governance API Key: ${apiKey}` is the same pattern. Today the consumer gates all four fields with `hasText` so this branch is unreachable from there, but `BrandGovernanceClient` is a public export and a different caller passing a partial config would leak the secret. Fix: never serialize the config - report only which fields are missing. (The original `BrandClient` has the identical leak; worth a follow-up to fix both.)

**5. The checks call hardcodes `pageSize=100` with no pagination (`brand-governance-client.js:21`, used at `:132`).** A brand with more than 100 active BRAND checks silently drops COPY guidelines beyond the first page - a 200 with an incomplete set, no error, no log. Fix: paginate to exhaustion, or if 100 is a deliberate cap, log a WARN when `data.length === pageSize` and document it.

**6. The cached IMS service token has no expiry handling (`brand-governance-client.js:60-78`).** `#getImsAccessToken` stores `response?.access_token` (the bare string) and discards `expires_in`, then guards re-entry on truthiness, so once cached the token is reused indefinitely. This is latent today (the client is cached on a per-request context) but is a real defect in a shared library the moment the client is reused across requests. Fix: cache the full token with an expiry and treat near-expiry as absent, or rely on `ImsClient`'s own cache. (Copied verbatim from `BrandClient` - same fix applies there.)

#### Minor (Nice to Have)

**7. Harden the response-parsing path (`brand-governance-client.js:132-153`):**
- `brand.id` is interpolated into the checks URL without a `hasText` guard or `encodeURIComponent`, while `from-url` correctly encodes its argument - a missing id produces `/brands/undefined/checks`.
- `checksResult.data || []` silently yields empty guidelines if the envelope differs; consider logging when `data` is absent on an otherwise-ok response so upstream drift is visible.
- the `text: check.rule` mapping (output field `text`, source field `rule`) is a surprising rename with no comment and no guard for an undefined `rule`.

### Recommendations
- Unlike `BrandClient.getBrandGuidelines`, the two `response.json()` parses here are not wrapped, so a 200 with a malformed body throws a raw `SyntaxError` to the caller. Mirror the sibling's try/catch-to-typed-error.
- Set an explicit, tighter timeout on the `tracingFetch` calls (the 10s default is generous for a now-primary hot-path dependency) and consider a lightweight circuit breaker so a Brand Governance brownout fast-fails to the consumer's Brand Publish fallback instead of paying the full timeout each request.
- Optional SSRF defense-in-depth: the repo already ships `isNonPublicHostname`; applying it to `siteBaseUrl` before egress would match the pattern elsewhere (low risk here, since the URL is only a query value on an operator-controlled base URL).
- The test plan references `scripts/test-brand-governance-api.js`, which is not in the diff - include it or drop the bullet.

### Assessment
Ready to merge? No - with fixes.
Reasoning: The client is competently written and consistent with `BrandClient`, but it is red on coverage, drops a tenant-isolation guard the sibling enforces, and can leak a secret on a path reachable by other callers. None require a rewrite. Add the tests, restore the org guard, stop serializing the config/key, and reconcile the response shape.
Note: CI checks are currently failing - resolve before merge.

### Next Steps
1. Add the client tests (item 1) and restore the cross-org guard (item 3).
2. Stop serializing the IMS config / API key (item 4) and reconcile the response shape (item 2).
3. Address pagination (item 5) and token expiry (item 6); the minor and recommendations are optional.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,156 @@
...
```
solaris007, line 174: Still open (the main one): this shape drops `toneOfVoice`, `coreValues`, `restrictions`, and `additionalGuidelines` that the `BrandGuidelines` type advertises and that Mystique's `format_brand_guidelines_to_markdown` renders. As the primary source, BG-registered sites silently lose four of five sections. Map COPY checks into the existing field structure, or align the contract with consumers and document the new shape.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,156 @@
...
+    if (brandResponse.status === HTTP_NOT_FOUND) {
+      return null;
+    }
+    if (!brandResponse.ok) {
+      throw this.#createError(
+        `Error resolving brand for URL ${siteBaseUrl}: ${brandResponse.status}`,
+        brandResponse.status,
+      );
+    }
+    const brand = await brandResponse.json();
```
solaris007, line 137: Still open: no cross-org check. The sibling `BrandClient.getBrandGuidelines` verifies `result.org_id === imsOrgId` before returning; here the brand is resolved by domain and the caller-supplied `imsOrgId` is echoed back without confirming the resolved brand belongs to that org. Restore the guard, or document that org scoping is delegated to the Agent via the `x-gw-ims-org-id` header.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,156 @@
...
+import {
+  isValidIMSOrgId, hasText, isValidUrl, tracingFetch as fetch,
+} from '@adobe/spacecat-shared-utils';
+import { ImsClient } from '@adobe/spacecat-shared-ims-client';
+
+const HTTP_BAD_REQUEST = 400;
+const HTTP_NOT_FOUND = 404;
+
+const API_GET_BRAND_FROM_URL = (url) => `/api/v1/brands/from-url?url=${encodeURIComponent(url)}`;
+const API_GET_BRAND_CHECKS = (brandId) => `/api/v1/brands/${brandId}/checks?status=ACTIVE&type=BRAND&pageSize=100`;
```
solaris007: Still open: `pageSize=100` with no pagination, and the new test now asserts this single-page query. A brand with more than 100 active BRAND checks silently truncates its COPY guidelines. Paginate to exhaustion, or if 100 is a deliberate cap, log a WARN when `data.length === pageSize` and document it.

solaris007 (commented):
Hey @anagarwa,

Re-review of the changes since my last pass (four new commits: a test suite plus three fixes). Reviewed again alongside the companion adobe/spacecat-api-service#2419.

## Previously flagged, now addressed
- Tests: a full `BrandGovernanceClient` suite is in place (constructor validation, `createFrom` caching, COPY-scope filtering, 404-to-null, missing/null/empty scopes, invalid inputs, empty-token, non-404 and checks-fetch errors, and fresh-token-per-call). These assert observable behavior rather than just hitting lines - good quality.
- Secret in logs: the IMS-config validation no longer serializes the config; it reports only the missing field names (`brand-governance-client.js:107-111`). The `clientSecret` leak is gone.
- Token staleness: the per-instance token cache was removed in favor of fetching a fresh IMS token per call, with an empty-token guard that throws instead of sending `Bearer undefined` (`brand-governance-client.js:69-89`). This resolves the warm-Lambda 401 risk (and the same points BaDos raised).

## Still open from the prior review
- Response shape (`brand-governance-client.js:147-160`): unchanged. The client still returns guidelines-only `{ id, name, imsOrgId, createdAt, updatedAt, guidelines: [{name,text}] }`, dropping `toneOfVoice`, `coreValues`, `restrictions`, and `additionalGuidelines` that the `BrandGuidelines` type advertises and that the consumer's Mystique formatter renders. This is the one I would most want resolved before this becomes the primary source.
- Cross-org guard (`brand-governance-client.js:128`): unchanged. The sibling `BrandClient.getBrandGuidelines` verifies `result.org_id === imsOrgId`; this client resolves a brand by domain and echoes the caller-supplied `imsOrgId` back without verifying ownership. Restore the guard or document that org scoping is delegated to the Agent via the `x-gw-ims-org-id` header.
- Pagination (`brand-governance-client.js:21`): unchanged, and now locked in by a test that asserts `pageSize=100`. A brand with more than 100 active BRAND checks still silently truncates its COPY guidelines.
- Minor: line 57 still interpolates `apiKey` into the constructor error message (benign today since it only fires when the key is empty, but the pattern remains); `brand.id` is still interpolated into the checks path without `encodeURIComponent`; and `checksResult.data || []` still masks an unexpected envelope (now asserted to return empty rather than surfacing the drift).

## CI
The `Test` check is red, but the cause is `SemanticReleaseError: Invalid npm token` in the semantic-release verify step - not your tests or coverage. The `BrandGovernanceClient` suite passes and coverage is met. That looks like a CI/secrets infra issue rather than anything in this diff; worth a maintainer's eye on the npm token rather than a code change here.

## Assessment
Ready to merge? With fixes.
Good progress on the fixes. The remaining blocker is the response-shape decision; the cross-org guard and pagination are the next two. The red CI appears to be infra (npm token), not this code.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
anagarwa: @solaris007 addressed via https://github.com/adobe/spacecat-api-service/pull/2419/changes#diff-6bc144b1c5b53a777eaa219a2c4156ef691787a0b150463b34ae34272fcb8f26R256

packages/spacecat-shared-brand-client/src/brand-governance-client.js
anagarwa: @solaris007 
Mapping COPY-scoped checks into the full BrandGuidelines shape (toneOfVoice, coreValues, restrictions, additionalGuidelines) is tracked in
LLMO-4919 as a follow-up. We'd like to land this PR with the current shape and iterate — the fallback to Brand Publish ensures BG-registered sites are not broken in the interim.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
anagarwa: Mapping COPY-scoped checks into the full BrandGuidelines shape (toneOfVoice, coreValues, restrictions, additionalGuidelines) is tracked in
LLMO-4919 as a follow-up. We'd like to land this PR with the current shape and iterate — the fallback to Brand Publish ensures BG-registered sites are not broken in the interim.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
anagarwa: @solaris007 The BrandResponse schema from the Brand Governance Agent does not include an org_id field (confirmed from the OpenAPI spec), so a client-side comparison like BrandClient performs
is not possible. Org scoping is enforced server-side by the Agent via the x-gw-ims-org-id header sent on every request. We've added a hasText(brand.id) guard to prevent
/brands/undefined/checks being called if the resolved brand has no id.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,170 @@
...
+      }
+      // eslint-disable-next-line no-await-in-loop
+      const { data = [] } = await checksResponse.json();
+      allChecks.push(...data);
+      if (data.length < CHECKS_PAGE_SIZE) {
+        break;
+      }
+      page += 1;
+    // eslint-disable-next-line no-constant-condition
+    } while (true);
```
solaris007, line 170: Important (should fix): this `do { ... } while (true)` is bounded only by the Agent returning a short page (`data.length < 100`) or a non-2xx - there is no max-page cap or non-advancing-page guard. If the Agent ever ignores or clamps the `page` parameter and keeps returning full 100-item pages, this loops until the Lambda times out, with `allChecks` growing unbounded. That matters more now that adobe/spacecat-api-service#2419 makes this the primary brand-guidelines source - the hot path should not hang on a single upstream misbehavior.

Fix: add a `MAX_PAGES` (or max-total-checks) cap and throw via `#createError` when exceeded, so the consumer's Brand Publish fallback engages rather than silently truncating; if the checks response carries a total/count field, prefer driving termination off that. Add a test that the loop bails when the mock always returns a full page.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -0,0 +1,170 @@
...
+  /**
+   * Fetches brand guidelines for a site URL from the Brand Governance Agent.
+   * Returns null if the brand is not registered (404), allowing callers to fall back.
+   *
+   * @param {string} siteBaseUrl - The site's base URL to resolve brand by domain
+   * @param {string} imsOrgId - The IMS org ID sent as x-gw-ims-org-id header
+   * @param {object} imsConfig - IMS auth config: { host, clientId, clientCode, clientSecret }
+   * @returns {Promise<object|null>} Brand guidelines or null if not registered
+   */
+  async getBrandGuidelinesForUrl(siteBaseUrl, imsOrgId, imsConfig) {
```
solaris007, line 103: Important (should fix): cross-org isolation now rests entirely on the Agent honoring `x-gw-ims-org-id`. This is the accepted item-3 design and I am not asking for a client-side `org_id` check (impossible without `org_id` on `BrandResponse`) - but the two cheap in-scope safeguards for that boundary are missing:

1. Document the trust assumption here. The commit message says org scoping was documented, but the explanatory comment was removed in `f358d943`; only the `@param imsOrgId` line (:92) remains, and it states the mechanism, not the rationale. Add 2-3 lines stating that (a) org/tenant scoping for the domain lookup is enforced server-side via `x-gw-ims-org-id`, (b) `BrandResponse` carries no `org_id` so the sibling's `org_id === imsOrgId` check is not portable, and (c) the returned `imsOrgId` (:164) is the requested org echoed for shape-parity, not a verified owner. That last point is a real trap: the sibling sets the same-named field from the resolved resource's `org_id`, so a consumer migrating between the two clients could read `result.imsOrgId` as "owner of this brand" when here it only means "the org I asked about."

2. Add a test that the header is actually sent. Every `nock` scope currently matches URL + query only; none assert `.matchHeader('x-gw-ims-org-id', ...)`. A refactor could drop or mis-set the one header the isolation guarantee depends on and the suite would stay green. Add `.matchHeader('x-gw-ims-org-id', validImsOrgId)` to at least one happy-path scope (ideally `x-api-key` / `Authorization` too).

solaris007 (approved):
Hey @anagarwa,

Third pass, reviewed again alongside the consumer companion adobe/spacecat-api-service#2419. This covers the four commits since my last pass (server-side `scope=COPY` + pagination, the apiKey-log fix, the org-scoping doc, the `brand.id` guard). I read the new client and the sibling `BrandClient` at HEAD.

Headline: the prior round is genuinely resolved. The two original blockers (zero tests, secret-in-logs) and the token-staleness item are fixed and verified, and both pushbacks (response shape, cross-org guard) hold up. What is left is hardening on the one net-new code path (pagination) and making the tenant-isolation trust boundary explicit, since the design now leans on it entirely.

Note: CI (Test) is red, but the only failing step is `@semantic-release/npm` `verifyConditions` (`EINVALIDNPMTOKEN Invalid npm token`) in the release dry-run. Your tests and coverage pass. This is an npm-registry token infra issue, not this code.

## Test quality snapshot
- New code 206 LOC / 3 files; new tests 274 LOC / 1 file (ratio 1.33); 20 new `it()` blocks.
- Fired signal `assertion-thin` (0.65/test) is a measurement artifact, not a real gap: the preflight does not count chai-as-promised (`.rejectedWith`), `.to.be.null`, `.to.have.length`, or `nock.isDone()` assertions, which dominate this suite. Real density is ~1.15 and every test carries a behavioral assertion at a real boundary. No action.

## Strengths
Previously flagged, now resolved:
- Tests (prior Critical): a full `BrandGovernanceClient` suite is in place and is good quality - behavior-asserting, mocked only at the network boundary via `nock` (no internal-collaborator stubbing), covering every reachable branch (constructor/`createFrom`, all input guards, 404-to-null, no-id, empty/missing data, multi-page accumulation, every missing IMS field, empty token, non-404 and checks-fetch errors, fresh-token-per-call).
- Secret-in-logs (prior Important): fully fixed. Every error string was traced - the API-key error omits the value (`:53`), the IMS-config error reports field names only (`:106`), the token error carries no value, and `headers` are never logged. This client is actually stricter than its sibling.
- Token lifecycle (prior Important): fixed, and more robust than a naive expiry check. Fetching a fresh token per call (via a fresh `ImsClient`) sidesteps both the removed instance cache and `ImsClient`'s own expiry-unaware cache, so every request carries a freshly minted token. The empty-token guard (`:87-89`) prevents `Bearer undefined`; pinned by the `.twice()` + `nock.isDone()` test.
- `brand.id` guard (`:131`) prevents a `/brands/undefined/checks` call; tested.

Also:
- Honest typing: splitting out a distinct `BrandGovernanceGuidelines` type so the return shape matches the implementation resolves the prior over-advertising concern at compile time.
- Additive and non-invasive: the default-export `BrandClient` is unchanged, so no existing consumer regresses; the new behavior is opt-in via the named export.

On the two pushbacks:
- Response shape (prior item 2): your deferral to LLMO-4919 is reasonable - the contract is now type-separated, additive, and tracked. One precision note for the team's awareness (not a blocker): the "Brand Publish fallback protects BG-registered sites" rationale is inverted. The consumer's fallback fires on the 404-to-null path, i.e. for sites NOT registered in BG; a BG-registered site receives the narrow shape, so the structured fields (`toneOfVoice`/`coreValues`/`restrictions`/`additionalGuidelines`) are dropped for exactly the sites that resolve. Whether that is intended model migration or interim data loss is a consumer/Agent judgement, but it is worth being clear-eyed that registered sites are not shielded by the fallback.
- Cross-org guard (prior item 3): the pushback holds. With no `org_id` on `BrandResponse`, the sibling's `result.org_id === imsOrgId` check is genuinely impossible to port, and delegating org scoping to the gateway-enforced `x-gw-ims-org-id` header is a sound pattern. I am not asking for a client-side org check. See Important finding 2 for the one in-scope residual (documentation + a test) and the out-of-scope item to confirm with the Agent team.

## Issues

### Important (Should Fix)

1. Bound the pagination loop (`brand-governance-client.js:136-157`). The `do { ... } while (true)` terminates only when the Agent returns a short page (`data.length < 100`) or a non-2xx. There is no max-page cap or non-advancing-page detection. If the Agent ever ignores or clamps the `page` parameter and keeps returning full 100-item pages (a plausible upstream bug for a paginated endpoint this client does not own), the loop issues sequential calls until the Lambda times out, with `allChecks` growing unbounded. This matters more now that adobe/spacecat-api-service#2419 makes this the primary brand-guidelines source - the hot path should not be able to hang on a single upstream misbehavior. Fix: add a `MAX_PAGES` (or max-total-checks) cap and throw via `#createError` when exceeded, so the consumer's Brand Publish fallback can engage rather than silently truncating; if the checks response carries a total/count field, prefer driving termination off that. Add a test that the loop bails when the mock always returns a full page.

2. Make the tenant-isolation boundary explicit and regression-proof (`brand-governance-client.js:96` and `test/index.test.js`). Item 3's accepted design rests cross-org isolation entirely on the Agent honoring `x-gw-ims-org-id`, and the two cheap in-scope safeguards for that boundary are missing:
   - Document the trust assumption in code. The commit message says org scoping was documented, but the explanatory comment was removed in `f358d943`; only the JSDoc `@param imsOrgId` line (`:92`) remains, which states the mechanism, not the rationale. Add 2-3 lines on `getBrandGuidelinesForUrl` stating that (a) org/tenant scoping for the domain lookup is enforced server-side via `x-gw-ims-org-id`, (b) `BrandResponse` carries no `org_id` so the client-side `org_id === imsOrgId` check the sibling performs is not possible here, and (c) the returned `imsOrgId` (`:164`) is the requested org echoed for shape-parity, not a verified owner. That last point is a real trap: the sibling sets the same-named field from the resolved resource's `org_id`, so a consumer migrating between the two could read `result.imsOrgId` as "owner of this brand" when here it only means "the org I asked about."
   - Add a test that the header is actually sent. Every `nock` scope currently matches URL + query only; none assert `.matchHeader('x-gw-ims-org-id', ...)`. A refactor could drop or mis-set the one header the isolation guarantee depends on and the suite would stay green. Add `.matchHeader('x-gw-ims-org-id', validImsOrgId)` to at least one happy-path scope (ideally `x-api-key` / `Authorization` too).

### Minor (Nice to Have)

3. Null checks-body throws an unstructured `TypeError` (`brand-governance-client.js:150`). `const { data = [] }` applies the default only for `undefined`; a 200 with a JSON `null` body makes `const { data = [] } = null` throw a raw `TypeError` that escapes the method's `#createError` pattern as an unhandled rejection. The sibling wraps `response.json()` in try/catch and converts to a typed error. Cheap to harden: `const body = await checksResponse.json() ?? {};` then read `body.data`.

4. Unguarded field rename in the mapper (`brand-governance-client.js:159`). `allChecks.map((check) => ({ name: check.name, text: check.rule }))` yields `{ name: undefined, text: undefined }` for a check missing those fields, which the consumer then renders; the undefined-`rule` case is also untested. Consider skipping or defaulting checks with no `rule`.

## Recommendations
- `brandId` is interpolated into the checks path unencoded (`:23`), while `from-url` encodes its argument (`:22`). Not exploitable (`brand.id` comes from the trusted Agent response and is `hasText`-guarded) and consistent with the sibling's unencoded interpolation, so optional - but wrapping `brand.id` in `encodeURIComponent` makes the two URL builders consistent.
- Add a short doc comment on the new `BrandGovernanceGuidelines` type noting it is intentionally a different model from `BrandClient`'s `guidelines` (checks-based `[{name,text}]` vs the structured Brand-Publish shape), so a future maintainer does not "unify" them. That field name is now polymorphic across the two clients the consumer switches between.

- Confirm with the Brand Governance Agent team that `GET /api/v1/brands/from-url` is strictly scoped server-side by `x-gw-ims-org-id`. A domain-to-brand lookup is inherently global unless intersected with the org header; if it is not, this (now primary) client could surface cross-tenant data despite correct usage here. This is the actual isolation control and it lives in the Agent, not this repo.
- The sibling `BrandClient` still serializes `${apiKey}` and `JSON.stringify(imsConfig)` (including `clientSecret`) into its own error messages. Pre-existing and not in this diff (this PR only adds the export line), so out of scope here - but it is the exact leak the new client correctly avoids, and worth a follow-up.
- @lucianfelix's point that the data source should live in the blackboard is a separate-system design direction, not actionable on this diff.

## Assessment
Ready to merge? Approving conditionally - the two should-fix items are the conditions, not blockers on a formal re-review.
Reasoning: No Critical findings and nothing is wrong with the current code's behavior; the prior round is fully resolved and both pushbacks hold. I am approving so this is not gated on another round, on the condition that you (1) bound the new pagination loop so a now-primary hot path cannot hang on upstream misbehavior, and (2) make the tenant-isolation trust boundary the design now depends on explicit (a comment plus one `x-gw-ims-org-id` header assertion). The two Minor items are optional. CI red is the npm-token infra issue, not your code.

## Next Steps
1. Bound the pagination loop (Important finding 1) - the one I would not skip.
2. Document the org-scoping trust assumption and add the `x-gw-ims-org-id` header assertion (Important finding 2).
3. Minor items 3-4 and the recommendations are optional.

BaDos (approved):

lucianfelix (approved):

PR #1615, opened 2026-05-21
feat(data-access): add BrandSemrushProject entity

rainer-friederich:
## Summary

New PostgREST-backed entity in `@adobe/spacecat-shared-data-access` for the Semrush AIO proxy work in [adobe/spacecat-api-service#2456](https://github.com/adobe/spacecat-api-service/pull/2456). Mirrors the layout of `SiteEnrollment` 1:1.

`BrandSemrushProject` maps an Adobe brand to a Semrush AIO project at the granularity of a single `(brand, semrushLocationId, language)` slice.

```
src/models/brand-semrush-project/
  brand-semrush-project.model.js       # class BrandSemrushProject extends BaseModel
  brand-semrush-project.schema.js      # SchemaBuilder w/ belongs_to Brand + 3 attributes + addAllIndex
  brand-semrush-project.collection.js  # class BrandSemrushProjectCollection — findBySlice
  index.js                             # barrel
  index.d.ts                           # public TS types
```

The DB table remains `brand_to_semrush_projects` (per [adobe/mysticat-data-service#616](https://github.com/adobe/mysticat-data-service/pull/616), already merged). The entity-to-table mapping is wired via a one-line addition to `ENTITY_TABLE_OVERRIDES` in `src/util/postgrest.utils.js` — same mechanism as `LatestAudit` and `SuggestionGrant`.

## Attributes

| Attribute | Type | Notes |
|---|---|---|
| `brandId` | uuid | FK → Brand (via `belongs_to`); auto-generated `allByBrandId` / `findByBrandId`. Brand has no entity in spacecat-shared — brand data is fetched via `@adobe/spacecat-shared-brand-client`, so the `getBrand()` accessor type is omitted from `index.d.ts` |
| `semrushProjectId` | string | Semrush AIO project identifier; indexed via `addAllIndex(['semrushProjectId'])` |
| `semrushLocationId` | number | Google Ads Geo Target ID (= Semrush `ProjectRequest.location_id`, e.g. `2840` = US); country implicit in the hierarchy. `Number.isInteger(value) && value > 0` |
| `language` | string | BCP-47-shaped tag matching `/^[a-z]{2,3}(-[a-z]{2,4})?$/` (e.g. `en`, `de`, `de-ch`, `zho`). Mirrors the DB CHECK constraint so the slice uniqueness gate cannot be silently bypassed by casing or format drift |

## Custom collection method

- `findBySlice(brandId, semrushLocationId, language)` — composite-key lookup used by `POST /v2/orgs/:org/brands/:brand/semrush/projects` to 409 on a duplicate slice before calling Semrush.

## Test plan

- [x] `npm test -w packages/spacecat-shared-data-access` — **2046 passing**, 100% line / branch / statement coverage on the package
- [x] **39 tests** under `test/unit/models/brand-semrush-project/`:
  - `brand-semrush-project.schema.test.js` — **21 cases** covering negative validator paths: `semrushLocationId` (0, -1, 1.5, NaN, Infinity, numeric string), `semrushProjectId` (empty, null, non-string), `language` (uppercase, mixed case, BCP-47-shape violations, whitespace), `brandId` (non-UUID)
  - `brand-semrush-project.collection.test.js` — **14 cases**: constructor, `findBySlice` happy/null, and smoke + delegation tests for all four auto-generated accessors (`allByBrandId`, `findByBrandId`, `allBySemrushProjectId`, `findBySemrushProjectId`) so a schema-index typo can no longer silently drop them from the public surface
  - `brand-semrush-project.model.test.js` — **13 cases**: constructor + per-attribute getter/setter pairs
- [x] Test fixture at `test/fixtures/brand-semrush-projects.fixture.js`
- [ ] CI green
- [ ] semantic-release cuts new minor (commit is `feat:`)

## Dependencies

- DB schema ships in [adobe/mysticat-data-service#616](https://github.com/adobe/mysticat-data-service/pull/616) — merged.
- After this PR ships a new `@adobe/spacecat-shared-data-access` version, bump it in `spacecat-api-service/package.json`.

## Design

The Semrush workspace is reachable via `brand -> organization.getSemrushWorkspaceId()` (already on the `Organization` model). Not duplicated on this entity. A row only exists when onboarding fully succeeded — no `status` / `retry_count` / `next_retry_at` / `last_error` columns. Full rationale in the design doc at https://github.com/adobe/spacecat-api-service/blob/feat/semrush-proxy/docs/plans/2026-05-21-semrush-proxy-and-onboarding.md.

## Review history

- Initial entity name was `BrandToSemrushProject`; renamed during review to align with the noun-form convention used by sibling relation entities (`SiteEnrollment`, `BrandSite`, `SiteCandidate`).
- `language` validator was originally `hasText(value)` — tightened to the BCP-47 regex above so the slice uniqueness gate cannot be silently bypassed by casing or format drift.
- Negative validator + accessor smoke tests added so the validators cannot be silently removed in a future refactor with the suite still green.


packages/spacecat-shared-data-access/src/models/brand-to-semrush-project/brand-to-semrush-project.schema.js
```diff
@@ -0,0 +1,40 @@
...
+  })
+  .addAttribute('semrushLocationId', {
+    type: 'number',
+    required: true,
+    validate: (value) => Number.isInteger(value) && value > 0,
+  })
+  .addAttribute('language', {
+    type: 'string',
+    required: true,
+    validate: (value) => hasText(value),
```
rainer-friederich: **FIX — language validation is too loose for a key that determines slice uniqueness.**

`hasText(value)` accepts `"EN"`, `"english"`, `"en-US"`, whitespace-padded strings — any non-empty value. Since `findBySlice` enforces uniqueness via an exact `WHERE language = ?` match, inconsistent casing or format between callers silently creates duplicate slices (and duplicate Semrush projects/spend) without triggering the 409 gate.

Suggest pinning to lowercase ISO 639-1 at the schema layer:

```js
validate: (value) => /^[a-z]{2}$/.test(value),
```

Or if BCP-47 subtags are needed (`"zh-Hant"`, `"pt-BR"`), use `/^[a-z]{2,3}(-[A-Z][a-z]{3})?(-[A-Z]{2}|\d{3})?$/.test(value)`. Either way, also verify the companion migration adds the same CHECK constraint so DB-level inserts can't bypass it.

packages/spacecat-shared-data-access/src/models/brand-semrush-project/brand-semrush-project.collection.js
```diff
@@ -0,0 +1,39 @@
...
+   * Find the single row for a (brand, semrushLocationId, language) slice, or
+   * null. Used by spacecat-api-service POST /v2/orgs/.../semrush/projects to
+   * 409 on a duplicate slice before calling Semrush.
+   *
+   * @param {string} brandId
+   * @param {number} semrushLocationId
+   * @param {string} language
+   * @returns {Promise<BrandToSemrushProject|null>}
+   */
+  async findBySlice(brandId, semrushLocationId, language) {
```
rainer-friederich, line 34: **CONSIDER — positional args with two `string` parameters are easy to swap silently.**

`brandId` and `language` are both strings. A caller writing `findBySlice(brand.getLanguage(), locationId, brand.getId())` gets a silent empty result rather than a type error.

An object parameter makes call sites self-documenting and prevents transposition:

```js
async findBySlice({ brandId, semrushLocationId, language }) {
  return this.findByIndexKeys({ brandId, semrushLocationId, language });
}
```

The TypeScript signature in `index.d.ts` would also need updating — and becomes much harder to misuse.

packages/spacecat-shared-data-access/src/models/brand-to-semrush-project/index.js
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+export { default as BrandToSemrushProject } from './brand-to-semrush-project.model.js';
+export { default as BrandToSemrushProjectCollection } from './brand-to-semrush-project.collection.js';
+export { default as BrandToSemrushProjectSchema } from './brand-to-semrush-project.schema.js';
```
rainer-friederich: **CONSIDER — `BrandToSemrushProjectSchema` is invisible to TypeScript consumers.**

`index.d.ts` uses `export type *` which only re-exports type declarations. `BrandToSemrushProjectSchema` is a runtime value (the result of `schema.build()`), so TypeScript consumers cannot import it even though it's reachable at runtime.

Check whether other entity barrels export the built schema from their `index.js` — if they don't, drop this export for consistency. If they do, the `.d.ts` needs a matching value export alongside the type re-export.

packages/spacecat-shared-data-access/test/unit/models/brand-to-semrush-project/brand-to-semrush-project.collection.test.js
```diff
@@ -0,0 +1,100 @@
...
+      const result = await instance.findBySlice(
+        mockRecord.brandId,
+        mockRecord.semrushLocationId,
+        mockRecord.language,
+      );
+
+      expect(result).to.equal(null);
+    });
+  });
+});
```
rainer-friederich: **FIX — the auto-generated accessors are not tested at all.**

The schema declares a `belongs_to Brand` index and `addAllIndex(['semrushProjectId'])`, which auto-generates `allByBrandId`, `findByBrandId`, `allBySemrushProjectId`, and `findBySemrushProjectId`. None of these appear in the test suite.

A typo in a schema index key name silently removes these methods from the public surface with no test catching it. At minimum, add a smoke test per generated method confirming it exists on the collection instance:

```js
it('exposes allByBrandId', () => expect(instance.allByBrandId).to.be.a('function'));
it('exposes allBySemrushProjectId', () => expect(instance.allBySemrushProjectId).to.be.a('function'));
```

For deeper coverage: stub `allByIndexKeys` and verify the delegation matches the index key, similar to how `findBySlice` is tested.

packages/spacecat-shared-data-access/test/unit/models/brand-semrush-project/brand-semrush-project.model.test.js
```diff
@@ -0,0 +1,101 @@
...
+    it('gets language', () => {
+      expect(instance.getLanguage()).to.equal('en');
+    });
+
+    it('sets language', () => {
+      instance.setLanguage('de');
+      expect(instance.getLanguage()).to.equal('de');
+    });
+  });
+});
```
rainer-friederich, line 101: **FIX — no negative test coverage for the custom validators.**

Three custom validators were written specifically to reject bad input, but none of them are exercised with invalid values:

- `semrushLocationId`: should reject `0`, `-1`, `1.5`, `NaN`, and the string `"2840"` — these are the exact boundary conditions `Number.isInteger(value) && value > 0` was meant to catch
- `semrushProjectId` and `language`: should reject empty string and whitespace-only strings

Without failing tests, these validators can be silently removed in a future refactor with the test suite staying green.

**CONSIDER — `setBrandId('new-uuid')` tests with a non-UUID string.**

If the `belongs_to Brand` UUID validation fires at set-time, this test proves it's broken. If it fires at save-time, the test is valid but gives false confidence. Worth adding one case with a valid UUID and one that asserts the rejection path, to make the validation timing explicit.

rainer-friederich (commented):
Clean, well-structured addition — the framework usage is correct and `findBySlice` works as expected in PostgREST mode (multi-key `findByIndexKeys` applies each field as an `.eq()` filter, no composite SchemaBuilder index needed). Test coverage is good overall. Three things to address before merging:

**Must fix**

- Language validation (inline comment on schema.js line 36) — `hasText` accepts `"EN"`, `"english"`, `"en-US"`, etc. Since the 409 duplicate gate is an exact DB string match, inconsistent casing between callers silently bypasses it and creates duplicate paid Semrush projects. Suggest `/^[a-z]{2}$/.test(value)` or a BCP-47 pattern, plus a matching CHECK constraint in the companion migration.

- Auto-generated accessor tests (inline comment on collection.test.js) — `allByBrandId`, `allBySemrushProjectId`, `findByBrandId`, `findBySemrushProjectId` go completely untested. A schema index typo silently removes them with no test failing.

- Negative validator tests (inline comment on model.test.js) — `semrushLocationId` validator edge cases (0, -1, 1.5, NaN, string) and `hasText` validators for empty/whitespace inputs are never exercised.

**Worth considering**

- `findBySlice` positional args (inline comment on collection.js) — `brandId` and `language` are both strings; swapping them produces silent empty results. An object param is safer and self-documenting.

- `BrandToSemrushProjectSchema` TypeScript visibility (inline comment on index.js) — the runtime export is invisible to TS consumers because `index.d.ts` uses `export type *`.

- Missing `has_many` on Brand — `belongs_to Brand` here has no matching `has_many BrandToSemrushProject` on the Brand schema. Other entities in the registry pair both sides for consistent ORM traversal.

- Naming convention — all existing relation entities use noun forms (`SiteEnrollment`, `BrandSite`, `SiteCandidate`) not `X-to-Y` junction names. `BrandSemrushSlice` or `BrandSemrushProject` would be consistent. Minor, but worth aligning before the name lands in a published minor release.

- Verify `@adobe/spacecat-shared-utils` is in the package's `dependencies` in `package.json` — the diff doesn't show a change there; if it's only hoisted by the monorepo it will break consumers of the published package.

Also worth confirming: the companion mysticat-data-service migration carries a `UNIQUE` constraint on `(brand_id, semrush_location_id, language)`, not just an index. The application-side `findBySlice` 409 check has a TOCTOU window; the DB constraint is the authoritative guard.

packages/spacecat-shared-data-access/src/models/brand-to-semrush-project/brand-to-semrush-project.schema.js
rainer-friederich: Fixed in `1f8b0447`. Validator is now `hasText(value) && /^[a-z]{2,3}(-[a-z]{2,4})?$/.test(value)`, mirroring the existing CHECK constraint on `brand_to_semrush_projects.language`. Covers `en`, `de`, `de-ch`, `pt-br`, `zh-hant`, `zho`; rejects `EN`, `en-US`, `english`, `en_us`, leading/trailing whitespace, etc. Negative cases are pinned in the new `brand-semrush-project.schema.test.js`.

packages/spacecat-shared-data-access/src/models/brand-semrush-project/brand-semrush-project.collection.js
rainer-friederich: Keeping positional for now. The `BaseCollection` surface is uniformly positional (`findById`, `findByIndexKeys`, `allByIndexKeys`, and every auto-generated `findBy<Field>` / `allBy<Field>`), and sibling custom finders such as `SiteEnrollment.findByIndexKeys(...)` follow the same convention. Switching this one method to an object param would create an inconsistency callers would have to remember per method.

The transposition risk is real but narrow: `findBySlice` is called from exactly one place (`spacecat-api-service` `POST /v2/orgs/:spaceCatId/brands/:brandId/semrush/projects`) where the args come from typed route params, and the new `language` regex validator means a swapped `brandId` → `language` would be rejected at the entity boundary anyway. The TS signature in `index.d.ts` keeps the parameter names visible in IDE tooling.

Happy to revisit in a follow-up if we end up adding more `findBy(composite)` finders and want to unify them.

packages/spacecat-shared-data-access/src/models/brand-to-semrush-project/index.js
rainer-friederich: Keeping the runtime export. Every sibling entity barrel does the same — `SiteEnrollment/index.js`, `AccessGrantLog/index.js`, `SiteImsOrgAccess/index.js`, and the rest all `export { default as <Entity>Schema } from './...schema.js'`. The matching `.d.ts` files in those packages also use `export type *` and never re-export the schema as a value. So this is the established convention for the package, not a divergence.

The schema export is consumed by `EntityRegistry.registerEntity(BrandSemrushProjectSchema, BrandSemrushProjectCollection)`; TypeScript consumers don't need it because the registry call is internal to the package.

packages/spacecat-shared-data-access/test/unit/models/brand-to-semrush-project/brand-to-semrush-project.collection.test.js
rainer-friederich: Fixed in `1f8b0447`. Added a `describe('auto-generated index accessors')` block to `brand-semrush-project.collection.test.js` covering all four methods. Each test does both: asserts the method is exposed on the collection instance, and stubs the underlying `allByIndexKeys` / `findByIndexKeys` to verify the delegation passes the right index keys. So a schema-index typo or a missed `addAllIndex` call would now fail the suite.

packages/spacecat-shared-data-access/test/unit/models/brand-semrush-project/brand-semrush-project.model.test.js
rainer-friederich: Fixed in `1f8b0447`. Added `brand-semrush-project.schema.test.js` with full negative coverage:

- `semrushLocationId`: rejects `0`, `-1`, `1.5`, `Number.NaN`, `Infinity`, and the string `'2840'` (no implicit coercion)
- `semrushProjectId`: rejects `''`, `null`, `undefined`, and non-string types
- `language` (new tightened validator): rejects uppercase / mixed case (`'EN'`, `'En'`, `'en-US'`, `'zh-Hant'`), shape violations (`'english'`, `'e'`, `'en-'`, `'en-u'`, `'en-toolong'`, `'en_us'`), leading/trailing whitespace, and empty/whitespace-only strings
- `brandId` (auto-generated from `belongs_to Brand`): rejects non-UUID strings — covers the timing question by exercising the schema validator directly rather than at save-time

The validators can no longer be silently removed in a future refactor with the suite still green.

byteclimber (approved):

PR #1618, opened 2026-05-22
feat: add scraperConfig to SiteConfig

ymadobe:
## Related Issues
- LLMO-4975 — careinsurance.com scrape failures from Akamai Bot Manager
  blocking requests without `Accept-Language`

## Summary

Adds a per-site `scraperConfig` to `SiteConfig` so customers behind
bot-protection rules (Akamai Bot Manager, Cloudflare, etc.) can configure
the HTTP headers the content scraper sends, instead of relying on
hardcoded defaults in the scraper.

The new field is intentionally generic — a `customHeaders` map plus an
optional `injectDefaults` flag — so future header requirements
(`User-Agent`, `Referer`, custom auth headers, etc.) don't need another
schema change.

## Why a new config field instead of hardcoding the header in the scraper
1. It applies the same value to every customer's scrapes, not just the
   one site failing Akamai checks.
2. It cannot serve customers needing a different value (e.g. `fr-FR`)
   without another code change.
3. It risks regressing currently-passing scrapes via language-based
   redirects under `rejectRedirects: true`.

A site-level config solves all three: per-site values, no code change to
add a customer, and explicit opt-in scoping.


packages/spacecat-shared-data-access/src/models/site/config.js
```diff
@@ -379,6 +379,10 @@ export const configSchema = Joi.object({
     ).optional(),
     outputLocation: Joi.string().required(),
   }).optional(),
+  scraperConfig: Joi.object({
```
MysticatBot-Dev, line 415: The `customHeaders` schema currently accepts any string key without constraints. Consider adding:

1. A denylist for security-sensitive header names (`Host`, `Authorization`, `Cookie`, `Transfer-Encoding`, `Content-Length`, `X-Forwarded-*`) to prevent SSRF amplification, request smuggling, or auth confusion when the scraper consumes these values.
2. Size bounds (e.g., `.max(20)` on the object, `.max(128)` on keys, `.max(2048)` on values) to prevent DynamoDB item-size exhaustion.

Example:
```js
customHeaders: Joi.object()
  .pattern(
    Joi.string().max(128).regex(/^(?!host$|authorization$|cookie$|transfer-encoding$|content-length$|x-forwarded-)/i),
    Joi.string().max(2048)
  )
  .max(20)
  .optional(),
```

MysticatBot-Dev (changes requested):
Hey @ymadobe,

## Strengths

- **Clean, consistent implementation** (`config.js:382-385`, `:533`, `:954-958`, `:1015`): The schema, getter, updater, and `toDynamoItem` serialization follow the exact same pattern as neighboring config fields (`cdnLogsConfig`, `tokowakaConfig`). Anyone familiar with the codebase will understand this immediately.
- **Well-scoped diff**: Two files, one concern. No unrelated changes mixed in.
- **Flexible schema design** (`config.js:383`): Using `Joi.object().pattern(Joi.string(), Joi.string())` for `customHeaders` avoids modeling specific headers that would require schema changes for each new customer need.
- **Good test coverage** (`config.test.js:570-598`): Tests cover creation with data, absence of data, and mutation via the updater.

## Issues

### Important (Should Fix)

**1. No constraint on header key names - allows security-sensitive headers** (`config.js:382-383`)

The `customHeaders` schema accepts any string key. The downstream scraper will merge these into HTTP requests. Without a denylist or allowlist, a customer (or compromised admin) can supply headers like `Host`, `Authorization`, `Cookie`, `Transfer-Encoding`, or `X-Forwarded-For`. Depending on the scraper's request construction, this could enable SSRF amplification (overriding `Host`), request smuggling (`Transfer-Encoding`), or auth confusion (`Authorization`/`Cookie`).

Defense-in-depth at the schema layer prevents dangerous headers from being persisted in the first place - the scraper should also validate, but a schema constraint means other consumers of `getScraperConfig()` cannot accidentally bypass the check.

Suggested fix - add a regex constraint on the key pattern:
```js
customHeaders: Joi.object().pattern(
  Joi.string().regex(/^(?!host$|authorization$|cookie$|transfer-encoding$|content-length$|x-forwarded-)/i),
  Joi.string()
).optional(),
```

**2. No max-size constraint on the customHeaders map** (`config.js:382-383`)

Without `.max()` on the object pattern or string values, an arbitrarily large header map can be stored. This could consume DynamoDB's 400 KB item size limit (causing writes to fail for other config fields on the same site) or cause memory pressure in the scraper if it blindly applies hundreds of headers.

Suggested fix - add reasonable bounds:
```js
customHeaders: Joi.object()
  .pattern(Joi.string().max(128), Joi.string().max(2048))
  .max(20)
  .optional(),
```

**3. Missing `toDynamoItem` round-trip test** (`config.test.js`)

The existing tests verify get/set round-trips but not that `scraperConfig` survives the `toDynamoItem` serialization path. Other config properties (`tokowakaConfig`, `edgeOptimizeConfig`, `commerceLlmoConfig`) have dedicated toDynamoItem tests. If the serialization line were accidentally removed, the current tests would still pass while data would be silently lost on persistence.

Suggested fix:
```js
it('includes scraperConfig in toDynamoItem conversion', () => {
  const data = Config({ scraperConfig: { customHeaders: { 'X-Custom': 'value' }, injectDefaults: true } });
  const dynamoItem = Config.toDynamoItem(data);
  expect(dynamoItem.scraperConfig).to.deep.equal({ customHeaders: { 'X-Custom': 'value' }, injectDefaults: true });
});
```

### Minor (Nice to Have)

**4. Optional chaining inconsistency in `toDynamoItem`** (`config.js:1015`)

```js
scraperConfig: config.getScraperConfig?.(),
```

`getScraperConfig` is unconditionally defined on every `Config` instance, so `?.` is unnecessary. The immediate neighbors (`getBrandConfig()`, `getCdnLogsConfig()`, `getLlmoConfig()`) all use direct calls. This creates a subtle inconsistency that may make future readers wonder if this method can be absent when others cannot.

Suggested fix: `config.getScraperConfig()` to match the surrounding code.

## Recommendations

- Document the `injectDefaults` semantics somewhere accessible to consumers - the PR description explains the intent, but the code alone does not hint at what "defaults" are being injected or who is responsible for that behavior.

## Assessment

**Ready to merge: With fixes**

The functional implementation is correct and follows established patterns well. The header-name constraint (item 1) is the most important fix - it is a small schema change that meaningfully reduces the attack surface for a field that stores customer-provided HTTP headers. The size bounds (item 2) and serialization test (item 3) are straightforward additions. The minor optional-chaining fix is a one-character change.

## Next Steps

1. Add header-name constraints to the schema (denylist or allowlist for security-sensitive headers).
2. Add size bounds on the customHeaders map.
3. Add a `toDynamoItem` serialization test.
4. Optionally clean up the `?.` to match surrounding code.

MysticatBot-Dev (changes requested):
Hey @ymadobe,

## Strengths

- **Clean, consistent implementation** (`config.js:382-385`, `:533`, `:954-958`, `:1015`): The schema, getter, updater, and `toDynamoItem` serialization follow the exact same pattern as neighboring config fields (`cdnLogsConfig`, `tokowakaConfig`). Anyone familiar with the codebase will understand this immediately.
- **Well-scoped diff**: Two files, one concern. No unrelated changes mixed in.
- **Flexible schema design** (`config.js:383`): Using `Joi.object().pattern(Joi.string(), Joi.string())` for `customHeaders` avoids modeling specific headers that would require schema changes for each new customer need.
- **Good test coverage** (`config.test.js:570-598`): Tests cover creation with data, absence of data, and mutation via the updater.

## Issues

### Important (Should Fix)

**1. No constraint on header key names - allows security-sensitive headers** (`config.js:382-383`)

The `customHeaders` schema accepts any string key. The downstream scraper will merge these into HTTP requests. Without a denylist or allowlist, a customer (or compromised admin) can supply headers like `Host`, `Authorization`, `Cookie`, `Transfer-Encoding`, or `X-Forwarded-For`. Depending on the scraper's request construction, this could enable SSRF amplification (overriding `Host`), request smuggling (`Transfer-Encoding`), or auth confusion (`Authorization`/`Cookie`).

Defense-in-depth at the schema layer prevents dangerous headers from being persisted in the first place - the scraper should also validate, but a schema constraint means other consumers of `getScraperConfig()` cannot accidentally bypass the check.

**2. No max-size constraint on the customHeaders map** (`config.js:382-383`)

Without `.max()` on the object pattern or string values, an arbitrarily large header map can be stored. This could consume DynamoDB's 400 KB item size limit or cause memory pressure in the scraper.

**3. Missing `toDynamoItem` round-trip test** (`config.test.js`)

Other config properties have dedicated toDynamoItem tests. If the serialization line were accidentally removed, the current tests would still pass while data would be silently lost on persistence.

### Minor (Nice to Have)

**4. Optional chaining inconsistency in `toDynamoItem`** (`config.js:1015`)

`getScraperConfig` is unconditionally defined on every `Config` instance, so `?.` is unnecessary and inconsistent with neighboring getters.

## Assessment

**Ready to merge: With fixes**

The functional implementation is correct and follows established patterns well. The header-name constraint (item 1) is the most important fix - it is a small schema change that meaningfully reduces the attack surface for a field that stores customer-provided HTTP headers.

## Next Steps

1. Add header-name constraints to the schema (denylist or allowlist for security-sensitive headers).
2. Add size bounds on the customHeaders map.
3. Add a `toDynamoItem` serialization test.
4. Optionally clean up the `?.` to match surrounding code.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 7m 38s | <b>Cost:</b> $2.27 | <b>Commit:</b> <code>2ffe6d726780d7f0bc8e9e3bd98781e8bd3a3b1c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot-Dev (approved):
Hey @ymadobe,

## Strengths

- **All four prior review findings addressed thoroughly** - Header key validation now uses RFC 7230 token character regex with `.max(64)` (`config.js:393-395`), header values are restricted to printable ASCII + tab with CR/LF/NUL blocked and capped at 1024 bytes (`config.js:395-397`), the map is bounded to 32 entries (`config.js:399`), the `toDynamoItem` round-trip test is included (`config.test.js:620-631`), and the optional chaining inconsistency is resolved.
- **Setter validates against the canonical schema** (`config.js:975-979`) - `updateScraperConfig` rejects bad input at write time rather than allowing it to silently persist to DynamoDB and surface as a downstream failure. This matches the established pattern for validated setters in this file.
- **Comprehensive negative-case test coverage** (`config.test.js:594-618`) - Tests exercise CRLF injection, invalid header names (spaces), non-string values, and non-object input. Each asserts on the specific rejection pattern.
- **Schema comment documents security rationale** (`config.js:379-390`) - Explains the header-injection and oversized-payload motivation, and explicitly notes that `injectDefaults` has no consumer yet with semantics to be pinned by the first consumer. Prevents future ambiguity.
- **Well-scoped diff** - Two files, one concern, no unrelated changes mixed in. Follows established SiteConfig patterns (schema, getter, setter, toDynamoItem) without introducing divergence.

## Recommendations

1. **Test the 32-header boundary** - The test suite validates individual header constraints but does not verify that `.max(32)` on the map is enforced. A test constructing 33 headers and asserting rejection would complete the boundary-value coverage.
2. **Consider `.min(1)` on header values** (`config.js:395`) - The current pattern permits empty strings as header values. While technically valid per HTTP, some CDNs and proxies strip empty-value headers unpredictably. A `.min(1)` would make behavior more predictable for consumers if the team has seen issues with empty headers in practice.

## Assessment

**Ready to merge: Yes**

The new commit cleanly addresses all prior review findings without introducing new concerns. Validation is sound - RFC 7230 token regex for keys and printable-ASCII restriction for values are the correct constraints for preventing header injection. Size bounds are reasonable. Test coverage is thorough across both happy-path and security-boundary negative cases. No blockers.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> claude-opus-4-6 | <b>Commit:</b> <code>c5070b9a</code><br>If this code review was useful, please react with :thumbsup:. Otherwise, react with :thumbsdown:.</sub>

MysticatBot-Dev (dismissed):
Hey @ymadobe,

## Strengths

- **All four prior review findings addressed thoroughly** - Header key validation now uses RFC 7230 token character regex with `.max(64)` (`config.js:393-395`), header values are restricted to printable ASCII + tab with CR/LF/NUL blocked and capped at 1024 bytes (`config.js:395-397`), the map is bounded to 32 entries (`config.js:399`), the `toDynamoItem` round-trip test is included (`config.test.js:620-631`), and the optional chaining inconsistency is resolved.
- **Setter validates against the canonical schema** (`config.js:975-979`) - `updateScraperConfig` rejects bad input at write time rather than allowing it to silently persist to DynamoDB and surface as a downstream failure. This matches the established pattern for validated setters in this file.
- **Comprehensive negative-case test coverage** (`config.test.js:594-618`) - Tests exercise CRLF injection, invalid header names (spaces), non-string values, and non-object input. Each asserts on the specific rejection pattern.
- **Schema comment documents security rationale** (`config.js:379-390`) - Explains the header-injection and oversized-payload motivation, and explicitly notes that `injectDefaults` has no consumer yet with semantics to be pinned by the first consumer. Prevents future ambiguity.
- **Well-scoped diff** - Two files, one concern, no unrelated changes mixed in. Follows established SiteConfig patterns (schema, getter, setter, toDynamoItem) without introducing divergence.

## Recommendations

1. **Test the 32-header boundary** - The test suite validates individual header constraints but does not verify that `.max(32)` on the map is enforced. A test constructing 33 headers and asserting rejection would complete the boundary-value coverage.
2. **Consider `.min(1)` on header values** (`config.js:395`) - The current pattern permits empty strings as header values. While technically valid per HTTP, some CDNs and proxies strip empty-value headers unpredictably. A `.min(1)` would make behavior more predictable for consumers if the team has seen issues with empty headers in practice.

## Assessment

**Ready to merge: Yes**

The new commit cleanly addresses all prior review findings without introducing new concerns. Validation is sound - RFC 7230 token regex for keys and printable-ASCII restriction for values are the correct constraints for preventing header injection. Size bounds are reasonable. Test coverage is thorough across both happy-path and security-boundary negative cases. No blockers.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 40s | <b>Cost:</b> $3.14 | <b>Commit:</b> <code>c5070b9a459a4d7c49aa99bbb16f3cef7c218329</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-data-access/src/models/site/config.js
```diff
@@ -379,6 +379,29 @@ export const configSchema = Joi.object({
...
+   * `customHeaders` is forwarded verbatim by the scraper as HTTP headers on outbound
+   * requests, so values are bounded and restricted to printable ASCII to avoid
+   * header-injection (CR/LF) and oversized-payload hazards.
+   *
+   * `injectDefaults` is an opt-in flag reserved for future consumer logic
+   * (e.g. inject a sensible default `Accept-Language` when none is supplied).
+   * It has no consumer yet; semantics will be pinned by the first consumer that uses it.
+   */
+  scraperConfig: Joi.object({
+    customHeaders: Joi.object()
```
duynguyen: Rename to `headers` — there's already a `fetchConfig.headers` (L307) using the same plain name. The `custom` prefix is redundant since this is already scoped under `scraperConfig`.

packages/spacecat-shared-data-access/src/models/site/config.js
```diff
@@ -379,6 +379,29 @@ export const configSchema = Joi.object({
...
+      // RFC 7230 token characters for header names
+      .pattern(
+        Joi.string().pattern(/^[A-Za-z0-9!#$%&'*+\-.^_`|~]+$/).max(64),
+        // Visible ASCII + tab, no CR/LF/NUL
+        Joi.string().pattern(/^[\t\x20-\x7E]*$/).max(1024),
+      )
+      .max(32)
+      .optional(),
+    injectDefaults: Joi.boolean().optional(),
```
duynguyen: This flag has no consumer yet and the semantics are explicitly TBD per the PR description. Let's drop it and wait for the first real consumer to add it — storing a no-op boolean in the DB is premature.

ymadobe: Hi Duy, the scraper consumer is coming right after this (PR https://github.com/adobe/spacecat-content-scraper/pull/890), and the immediate use case for careinsurance.com is exactly the case the flag is meant to cover: customer needs Accept-Language injected as a default behavior, not as an explicit per-call override. So the consumer isn't hypothetical - it's the same chain of PRs.

duynguyen: Thanks for the context — but the flag is still redundant for that use case. Setting `headers: { 'Accept-Language': 'en-US' }` in `scraperConfig` already achieves "inject Accept-Language on every scrape for this site". The distinction between a default and an explicit value is illusory: both result in the same header being sent. The flag just adds a layer of indirection — the scraper needs its own logic to decide what to inject — with no benefit over being explicit in the config.

AndreiAlexandruParaschiv (approved):

PR #1619, opened 2026-05-22
fix:  (data-access) batch getAllFixesWithSuggestionsByOpportunityId (SITES-45307)

mimchome:
[SITES-45307](https://jira.corp.adobe.com/browse/SITES-45307)

## Problem

`GET /sites/:siteId/opportunities/:oppId/fixes` fetches suggestions per fix in a loop -- 2N+1 queries for N fixes. Follow-up from [SITES-45274](https://jira.corp.adobe.com/browse/SITES-45274) which identified the pattern during the fixes controller review.

## Changes

- Add `getAllFixesWithSuggestionsByOpportunityId` to `FixEntityCollection` -- queries by `opportunityId` only, 3 PostgREST calls total instead of 2N+1
- Extract shared grouping logic into `#buildFixesWithSuggestions`, reused by both the new method and existing `getAllFixesWithSuggestionByCreatedAt`
- Type definitions and JSDoc

Consumer: [spacecat-api-service #2474](https://github.com/adobe/spacecat-api-service/pull/2474) (draft, waiting on this publish)

## Validated

Verified on CI dev -- fixes returned with suggestions embedded, matching existing response contract.

## Test plan

- 4 new unit tests (happy path, empty results, validation, error handling)
- Existing `getAllFixesWithSuggestionByCreatedAt` tests pass with shared helper


cwjwisse (dismissed):
LGTM — refactor preserves existing behavior and the N+1 reduction is the right call.

One worth addressing in this PR since you're touching the helper:

**Double-wrap of `DataAccessError`** in `#buildFixesWithSuggestions` callers (line 222 and the createdAt path). Both `allByIndexKeys` and `batchGetByKeys` already throw `DataAccessError` on failure, so the outer `catch` re-wraps it and buries the original message in `error.cause`. `DataAccessError`'s constructor preserves `cause` as-is, so the inner detail survives but is one level deeper than callers expect.

Quick fix — mirror the pattern already used at `base.collection.js:723`:

```js
} catch (error) {
  if (error instanceof DataAccessError) throw error;
  this.log.error('Failed to get all fixes with suggestions by opportunity ID', error);
  throw new DataAccessError('Failed to get all fixes with suggestions by opportunity ID', this, error);
}
```

Apply to both `getAllFixesWithSuggestionByCreatedAt` and `getAllFixesWithSuggestionsByOpportunityId`.

Optional / non-blocking (defer if you want):
- `result[0].fixEntity.getId() === 'fix-1'` in the new test depends on incidental mock ordering — `.to.have.deep.members([...])` would be more honest.
- Naming inconsistency `WithSuggestion` (singular, existing) vs `WithSuggestions` (plural, new). New is correct; old name is public API so leave it, but worth a JSDoc note.
- `allByIndexKeys({ opportunityId })` is unbounded — PR description says "3 PostgREST calls" but `batchGetByKeys` chunks at 50, so it's `1 + ceil(N/50) + ceil(M/50)`. Fine in practice, just worth noting in the JSDoc if opportunities can have hundreds of fixes.

cwjwisse (dismissed):

cwjwisse (approved):

PR #1620, opened 2026-05-22
feat: add spacecatOrgId to ScrapeJobParams

rarescheseli:

## Related Issues

Thanks for contributing!


packages/spacecat-shared-drs-client/src/index.js
```diff
@@ -239,11 +243,16 @@ export default class DrsClient {
       parameters.days_back = daysBack;
     }
 
-    return this.submitJob({
+    const jobParams = {
       provider_id: 'brightdata',
       priority,
```
MysticatBot-Dev, line 312: Every other domain-specific field in this method (`dataset_id`, `site_id`, `urls`, `days_back`) is nested inside `parameters`. This field is placed at the top level instead.

Is this intentional per the DRS API contract? If so, a brief comment here (e.g., `// top-level: DRS expects org ID outside parameters`) would prevent a future developer from moving it into `parameters` during cleanup. If it should be inside `parameters`, this would silently fail since DRS likely ignores unexpected top-level fields.

MysticatBot-Dev (changes requested):
Hey @rarescheseli,

## Strengths

- **Clean conditional inclusion** (`src/index.js:246-248`): The `if (spacecatOrgId)` guard correctly omits the field from the payload when not provided, matching the existing `daysBack` pattern and avoiding sending `null`/`undefined` to the DRS API.
- **Useful error enrichment** (`src/index.js:130-132`): Attaching `error.status` to thrown errors lets callers branch on HTTP status (retry on 429/503, fail fast on 4xx) without parsing error messages. The test update validates both message and status.
- **Thorough test coverage** (`test/index.test.js:405-443`): Both positive and negative cases are tested with nock body matchers that assert the exact outgoing request shape. The negative test (`expect(body).to.not.have.property('spacecat_org_id')`) proves the conditional works correctly.
- **Type definition kept in sync** (`src/index.d.ts:40`): The optional `spacecatOrgId?: string` correctly mirrors the runtime behavior without breaking existing callers.

## Issues

### Important (Should Fix)

1. **Confirm `spacecat_org_id` nesting level is intentional** - `packages/spacecat-shared-drs-client/src/index.js:246-248`

   Every other domain-specific field in `submitScrapeJob` (`dataset_id`, `site_id`, `urls`, `days_back`) is nested inside the `parameters` object. The new `spacecat_org_id` is placed at the top level of the job body, alongside `provider_id` and `priority`.

   If the DRS API contract genuinely expects `spacecat_org_id` at the top level (e.g., it is a job-level attribute rather than a scrape-specific parameter), this is correct - but a brief code comment explaining the intentional divergence would prevent a future developer from "fixing" it by moving it into `parameters`. If it should be inside `parameters`, the feature would silently fail (DRS likely ignores unexpected top-level fields).

   **Fix:** Confirm against the DRS API spec and either (a) add a one-line comment explaining the top-level placement, or (b) move it inside `parameters` if that is where DRS expects it.

### Minor (Nice to Have)

2. **`error.status` not reflected in type definitions** - `packages/spacecat-shared-drs-client/src/index.d.ts`

   The `#request` method now attaches a `.status` property to thrown errors, and there is a test asserting on it, but the `.d.ts` file does not capture this. TypeScript consumers would need to cast to access `err.status`. Consider adding a `DrsClientError` interface or class in a follow-up.

3. **Truthiness check inconsistency** - `packages/spacecat-shared-drs-client/src/index.js:246`

   `if (spacecatOrgId)` will silently drop an empty string `""`. The `daysBack` parameter above uses `if (daysBack !== undefined)` which is more precise. For a string ID, empty string is almost certainly a caller bug, so the truthiness check is defensible - but `if (spacecatOrgId != null)` would be more consistent with the `daysBack` pattern while still rejecting `undefined`/`null`.

## Recommendations

- Consider whether `spacecatOrgId` should be included in the info log message around line 232 (`Submitting DRS scrape job for dataset...`) for traceability during incident correlation.
- The `error.status` pattern is useful beyond this PR. A typed `DrsError` class (extending `Error` with a `status` property) would formalize this for TypeScript consumers as the error surface grows.

## Assessment

Ready to merge? With fixes
Reasoning: The core feature is correct and well-tested, but the nesting-level question (item 1) should be confirmed before merge. If `spacecat_org_id` belongs inside `parameters`, the current implementation would silently fail. A quick confirmation from the DRS API contract resolves this.

## Next Steps

1. Confirm the `spacecat_org_id` placement is correct per the DRS API spec (or move it inside `parameters`).
2. Minor items are optional and can be addressed in a follow-up.

MysticatBot-Dev (changes requested):
Hey @rarescheseli,

## Strengths

- **Clean conditional inclusion** (`src/index.js:246-248`): The `if (spacecatOrgId)` guard correctly omits the field from the payload when not provided, matching the existing `daysBack` pattern and avoiding sending `null`/`undefined` to the DRS API.
- **Useful error enrichment** (`src/index.js:130-132`): Attaching `error.status` to thrown errors lets callers branch on HTTP status (retry on 429/503, fail fast on 4xx) without parsing error messages. The test update validates both message and status.
- **Thorough test coverage** (`test/index.test.js:405-443`): Both positive and negative cases are tested with nock body matchers that assert the exact outgoing request shape. The negative test (`expect(body).to.not.have.property('spacecat_org_id')`) proves the conditional works correctly.
- **Type definition kept in sync** (`src/index.d.ts:40`): The optional `spacecatOrgId?: string` correctly mirrors the runtime behavior without breaking existing callers.

## Issues

### Important (Should Fix)

1. **Confirm `spacecat_org_id` nesting level is intentional** - `packages/spacecat-shared-drs-client/src/index.js:246-248`

   Every other domain-specific field in `submitScrapeJob` (`dataset_id`, `site_id`, `urls`, `days_back`) is nested inside the `parameters` object. The new `spacecat_org_id` is placed at the top level of the job body, alongside `provider_id` and `priority`.

   If the DRS API contract genuinely expects `spacecat_org_id` at the top level (e.g., it is a job-level attribute rather than a scrape-specific parameter), this is correct - but a brief code comment explaining the intentional divergence would prevent a future developer from "fixing" it by moving it into `parameters`. If it should be inside `parameters`, the feature would silently fail (DRS likely ignores unexpected top-level fields).

   **Fix:** Confirm against the DRS API spec and either (a) add a one-line comment explaining the top-level placement, or (b) move it inside `parameters` if that is where DRS expects it.

### Minor (Nice to Have)

2. **`error.status` not reflected in type definitions** - `packages/spacecat-shared-drs-client/src/index.d.ts`

   The `#request` method now attaches a `.status` property to thrown errors, and there is a test asserting on it, but the `.d.ts` file does not capture this. TypeScript consumers would need to cast to access `err.status`. Consider adding a `DrsClientError` interface or class in a follow-up.

3. **Truthiness check inconsistency** - `packages/spacecat-shared-drs-client/src/index.js:246`

   `if (spacecatOrgId)` will silently drop an empty string `""`. The `daysBack` parameter above uses `if (daysBack !== undefined)` which is more precise. For a string ID, empty string is almost certainly a caller bug, so the truthiness check is defensible - but `if (spacecatOrgId != null)` would be more consistent with the `daysBack` pattern while still rejecting `undefined`/`null`.

## Recommendations

- Consider whether `spacecatOrgId` should be included in the info log message around line 232 (`Submitting DRS scrape job for dataset...`) for traceability during incident correlation.
- The `error.status` pattern is useful beyond this PR. A typed `DrsError` class (extending `Error` with a `status` property) would formalize this for TypeScript consumers as the error surface grows.

## Assessment

Ready to merge? With fixes
Reasoning: The core feature is correct and well-tested, but the nesting-level question (item 1) should be confirmed before merge. If `spacecat_org_id` belongs inside `parameters`, the current implementation would silently fail. A quick confirmation from the DRS API contract resolves this.

## Next Steps

1. Confirm the `spacecat_org_id` placement is correct per the DRS API spec (or move it inside `parameters`).
2. Minor items are optional and can be addressed in a follow-up.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 48s | <b>Cost:</b> $3.67 | <b>Commit:</b> <code>3a7e364e7a8da8fe774239a823cd79e80ccd458e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

tb-adbe (dismissed):

tb-adbe (approved):

MysticatBot-Dev (approved):
Hey @rarescheseli,

## Strengths

- **Previously flagged nesting-level concern resolved** - You confirmed `spacecat_org_id` at the top level is per the DRS API contract. The job body correctly distinguishes job-level metadata (provider, priority, org ownership) from scrape-specific configuration (nested in `parameters`). The client mirrors that decision faithfully.
- **Clean error enrichment** (`src/index.js:139-141`): Attaching `.status` at the `#request` layer means all DRS operations gain status-based branching without per-method changes. Minimal, non-breaking, and well-placed.
- **Good test coverage for the new parameter** (`test/index.test.js:405-443`): Both positive (parameter present, asserted in outgoing body via nock matcher) and negative (parameter absent, asserted NOT in body) cases are covered. The assertions verify actual wire format, not internal state.
- **Backward-compatible addition** - Optional parameter with no default, `?` in the type definition, existing callers unaffected.
- **Clean abstraction boundary** (`src/index.js:300-312`): The camelCase-to-snake_case translation stays local to the client, not leaking DRS protocol details to consumers.

## Issues

### Minor (Nice to Have)

1. **Missing JSDoc comment on the type definition field** - `packages/spacecat-shared-drs-client/src/index.d.ts:43`

   Every other field in `ScrapeJobParams` has a `/** ... */` doc comment. The new `spacecatOrgId` field is the only one without one. The JSDoc exists in `index.js:227`, but TypeScript consumers reading the `.d.ts` directly lose that context.

   **Fix:** Add `/** SpaceCat organization ID. */` above the field, matching the style of its siblings.

## Recommendations

- The prior unresolved finding about `error.status` not being reflected in type definitions remains valid as a follow-up. TypeScript consumers catching errors from this client have no typed way to access `.status`. A `DrsClientError` type or augmented error interface would improve the consumer experience in a future PR.

## Assessment

Ready to merge? Yes
Reasoning: The change is correct, focused, well-tested, and backward-compatible. No security issues, no architectural concerns, CI passes, and a human reviewer has already approved. The single minor finding (missing JSDoc in .d.ts) is not a merge blocker.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 29s | <b>Cost:</b> $1.97 | <b>Commit:</b> <code>eb91012e49fdd7e1e81bb608d7e826c3bebe0520</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1628, opened 2026-05-26
feat(drs-client): add new reddit_comments scrape specific parameters (LLMO-4464)

tb-adbe:

## Related Issues
- https://jira.corp.adobe.com/browse/LLMO-4464
- https://jira.corp.adobe.com/browse/LLMO-4467

Thanks for contributing!


MysticatBot-Dev (approved):
Hey @tb-adbe,

## Strengths

- **Well-structured validation pattern** (`src/index.js:249-259`): The `REDDIT_COMMENTS_ONLY_PARAMS.find()` approach for dataset-specific parameter rejection is clean and extensible - one place to add a param name, one error path.
- **Deterministic wire format** (`src/index.js:282-295`): Always sending `comment_limit` and `sort_by` (with defaults) for reddit_comments makes the outbound request shape predictable regardless of which optional params the caller provided. Good for debuggability.
- **Falsy-value handling tested** (`test/index.test.js:449-464`): Explicitly verifying that `loadAllReplies: false` is forwarded (not silently dropped) shows awareness of the common `if (value)` vs `if (value !== undefined)` trap.
- **Non-reddit isolation test** (`test/index.test.js:433-447`): Confirms that `comment_limit` and `sort_by` defaults are not leaked to other dataset types - the exact regression that would bite during a future refactor of the conditional block.
- **Types, docs, and runtime validation ship atomically**: README, TypeScript declarations, JSDoc, and runtime guards all updated together. Consumers get compile-time guidance and runtime guardrails from a single version bump.

## Issues

### Minor (Nice to Have)

1. **No upper bound on `commentLimit`** - `src/index.js:264`: `isPositiveInteger` accepts any value up to `Number.MAX_SAFE_INTEGER`. While the downstream DRS/Bright Data service should enforce its own limits, a defensively large value forwarded in the request could cause unexpected resource consumption downstream. Consider a reasonable ceiling if Bright Data documents one. This is consistent with how `daysBack` works today, so not a divergence - just a note.

2. **Flat interface accumulates dataset-specific optionals** - `src/index.d.ts:34-50`: TypeScript cannot tell a caller at compile time that `commentLimit` is meaningless for a `wikipedia` scrape. Today this is one dataset with four extra params, so the blast radius is small. If a second dataset acquires its own unique params, a discriminated union (`{ datasetId: 'reddit_comments'; commentLimit?: number; ... } | { datasetId: 'wikipedia'; ... }`) would let the type system enforce what runtime validation currently handles alone. Not actionable now, worth tracking.

3. **Multiple sequential assertions in validation tests** - `test/index.test.js:496-526`: The `daysBack` and `commentLimit` validation tests each contain 4 sequential `await expect(...).rejectedWith(...)` assertions. If the first case fails, the remaining cases are never exercised in that run. Consider splitting into individual `it()` blocks or adding descriptive comments before each assertion for clearer CI output on failure.

## Recommendations

- Consider exporting `REDDIT_COMMENTS_DEFAULT_COMMENT_LIMIT` and `REDDIT_COMMENTS_DEFAULT_SORT_BY` for consumers who want to reference the defaults without hardcoding `150` or `'Best'` in their own code.
- A brief code comment near the default constants noting these are "client-chosen defaults, not server-documented guarantees" would help a future maintainer know to verify against the provider API if behavior drifts.

## Assessment

Ready to merge? Yes

This is a clean, well-scoped feature addition. The API surface is backward-compatible (all new params are optional, existing `daysBack` behavior is preserved). Validation is thorough with actionable error messages. Documentation, types, implementation, and tests are all in sync. The change does exactly what it says and nothing more.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 44s | <b>Cost:</b> $3.72 | <b>Commit:</b> <code>f1f14eea7179bc224806fd34f33bba42b0d3f335</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rarescheseli (approved):

PR #1629, opened 2026-05-26
feat(drs-client): export REDDIT_COMMENTS_SORT_BY_VALUES allowlist

tb-adbe:
## Summary
- Export `REDDIT_COMMENTS_SORT_BY_VALUES` (formerly file-local `VALID_REDDIT_SORT_BY`) so downstream callers can validate `sortBy` at their own boundary without duplicating the allowlist.
- Adds runtime export, `.d.ts` declaration, and a unit test asserting the membership.
- Renamed on export to match the sibling `REDDIT_COMMENTS_DEFAULT_SORT_BY` and to make the `reddit_comments`-only scope explicit. Safe rename — the previous name was never exported (introduced in 1.9.0 / #1628).

## Motivation
`spacecat-audit-worker` currently duplicates this Set in `offsite-brand-presence/handler.js` with a TODO to import from this package once exported. Follow-up PR there will drop the duplicate.

## Test plan
- [x] `npm run lint -w packages/spacecat-shared-drs-client`
- [x] `npm test -w packages/spacecat-shared-drs-client` — 89 passing, 100% coverage

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/spacecat-shared-drs-client/src/index.d.ts
+++ b/packages/spacecat-shared-drs-client/src/index.d.ts
@@ -160,4 +160,10 @@ export declare const SCRAPE_DATASET_IDS: Readonly<{
   WIKIPEDIA: 'wikipedia';
 }>;
 
+/**
+ * Allowlist of valid `sortBy` values for the `reddit_comments` dataset.
+ * Exported so callers can validate input at their own boundary.
+ */
+export declare const REDDIT_COMMENTS_SORT_BY_VALUES: ReadonlySet<RedditSortBy>;
+
 export default DrsClient;
--- a/packages/spacecat-shared-drs-client/src/index.js
+++ b/packages/spacecat-shared-drs-client/src/index.js
@@ -41,7 +41,14 @@ const VALID_SCRAPE_DATASET_IDS = new Set(Object.values(SCRAPE_DATASET_IDS));
 const URL_LOOKUP_BATCH_SIZE = 100;
 
 // Reddit-comments specific scrape parameters.
-export const REDDIT_COMMENTS_SORT_BY_VALUES = new Set(['Best', 'Top', 'New', 'Controversial', 'Old', 'Q&A']);
+// Exported so callers can validate `sortBy` at their own boundary without
+// duplicating the allowlist. Frozen for consistency with the sibling
+// EXPERIMENT_PHASES / SCRAPE_DATASET_IDS exports (note: Object.freeze does
+// not actually prevent Set mutation methods — the ReadonlySet<> type in the
+// .d.ts is what guards TypeScript consumers).
+export const REDDIT_COMMENTS_SORT_BY_VALUES = Object.freeze(
+  new Set(['Best', 'Top', 'New', 'Controversial', 'Old', 'Q&A']),
+);
 const REDDIT_COMMENTS_DEFAULT_COMMENT_LIMIT = 150;
 const REDDIT_COMMENTS_DEFAULT_SORT_BY = 'Best';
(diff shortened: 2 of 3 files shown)


rarescheseli (approved):

MysticatBot-Dev (commented):
Hey @tb-adbe,

## Strengths

- **Clean, focused change with consistent naming** (`src/index.js:43`): The rename from `VALID_REDDIT_SORT_BY` to `REDDIT_COMMENTS_SORT_BY_VALUES` aligns with the existing `REDDIT_COMMENTS_DEFAULT_SORT_BY` and `REDDIT_COMMENTS_ONLY_PARAMS` naming scheme, making it predictable for downstream consumers.
- **Good type declaration** (`src/index.d.ts:165-169`): Typed as `ReadonlySet<RedditSortBy>` which leverages the existing union type and communicates immutability intent to TypeScript consumers.
- **Safe rename**: The previous name was file-private (never exported, introduced in 1.9.0), so this is non-breaking by definition.
- **Test locks the export contract** (`test/index.test.js:255-258`): Asserting exact membership ensures the set cannot drift without a test failure.

## Issues

### Minor (Nice to Have)

1. **Runtime mutability mismatch with type declaration** - `packages/spacecat-shared-drs-client/src/index.js:43`

   The `.d.ts` declares `ReadonlySet<RedditSortBy>`, but the runtime value is a plain `new Set(...)`. Any JavaScript consumer (or anyone bypassing TypeScript) can call `.add()`, `.delete()`, or `.clear()` on the shared instance. The sibling constants `EXPERIMENT_PHASES` and `SCRAPE_DATASET_IDS` in this file use `Object.freeze()` to enforce runtime immutability. Following the same pattern here would be consistent:

   ```js
   export const REDDIT_COMMENTS_SORT_BY_VALUES = Object.freeze(
     new Set(['Best', 'Top', 'New', 'Controversial', 'Old', 'Q&A'])
   );
   ```

   Not a correctness bug today since the `ReadonlySet` type guards TypeScript consumers, but it matches the defensive pattern already established in this file.

2. **Tautological test** - `packages/spacecat-shared-drs-client/test/index.test.js:255-259`

   The test re-declares the exact same six string literals from the source and asserts set membership equality. It cannot fail unless the source values change, at which point both files change in lockstep. It serves as a contract test (proving the export is accessible), but a slightly leaner variant could assert `size` and spot-check one or two values rather than mirroring the full set:

   ```js
   expect(REDDIT_COMMENTS_SORT_BY_VALUES).to.be.an.instanceOf(Set);
   expect(REDDIT_COMMENTS_SORT_BY_VALUES.size).to.equal(6);
   expect(REDDIT_COMMENTS_SORT_BY_VALUES.has('Best')).to.be.true;
   ```

   This is a judgment call, not a blocker - the cost of the current test is near zero.

## Recommendations

- Consider exporting `REDDIT_COMMENTS_DEFAULT_SORT_BY` in a follow-up. Downstream callers validating `sortBy` will likely also need the default value when `sortBy` is omitted.

## Assessment

Ready to merge? **Yes.**

The change is clean, well-typed, tested, and non-breaking. The minor suggestions (freeze for pattern consistency, leaner test) are optional improvements.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 0s | <b>Cost:</b> $1.83 | <b>Commit:</b> <code>28ff20f62a9dc05105de2a2c907b4e84bbcf3857</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rarescheseli (dismissed):

rarescheseli (approved):

PR #1633, opened 2026-05-27
fix(data-access): mint UUID v7 ids to match DB schema default

ekremney:
## Summary

Aligns the SpaceCat ORM with the v7-by-design DB schema. The DB column default for every PK has been `uuid_generate_v7()` since [mysticat-data-service migration #1](https://github.com/adobe/mysticat-data-service/blob/main/db/migrations/20250130000001_initial_setup.sql) (Jan 2025), but `schema.builder.js` was stamping `crypto.randomUUID()` (v4) in `#applyDefaults` *before* the INSERT, defeating the DB default. This created a v4/v7 asymmetry on the same tables (sites, opportunities, suggestions) between ORM-written rows and Mystique-direct-INSERT rows, and the v4-strict gateway regex in `spacecat-api-service` rejected the v7 rows at the API boundary (SITES-45292).

Two changes, each independently shippable:

- `fix(data-access)`: replace the client-side v4 default with `uuidv7()`, an inline implementation that mirrors the DB's plpgsql function bit-for-bit (no new npm dep). Strengthens unit and IT-test assertions so a regression to v4 fails fast.
- `feat(utils)`: rename the misleadingly-named `UUID_V4_REGEX` constant to `UUID_REGEX` (the pattern was always version-agnostic), export both names with the v4 name marked deprecated, and add explicit v4/v7 coverage to the `isValidUUID` test suite.

## Why roll forward, not migrate

`sites.id` alone is referenced by 100+ partitioned analytics tables. In-place UUID swaps require coordinated cross-table UPDATEs that aren't worth the risk. Both v4 and v7 are valid RFC 4122/9562 UUIDs, so they coexist forever — new writes converge on v7 (sortable, locality-friendly B-tree indexes); old v4 rows live out their natural life.

## Where the v4 came from

`packages/spacecat-shared-data-access/src/models/base/schema.builder.js:42` (pre-fix):

\`\`\`js
const ID_ATTRIBUTE_DATA = {
  ...
  default: () => crypto.randomUUID(),   // Node's randomUUID → UUID v4
  validate: (value) => isValidUUID(value),
};
\`\`\`

`BaseCollection.create()` → `#prepareItem` → `#applyDefaults` runs *before* the PostgREST insert, so the row arrived at PG with \`id\` already populated and the DB v7 default never fired.

## What the new `uuidv7()` does

`packages/spacecat-shared-data-access/src/util/uuid.js` (new) implements the same byte layout as the DB's `uuid_generate_v7()` plpgsql function:

- Bytes 0-5: unix timestamp in milliseconds (big-endian)
- Bytes 6-15: 10 random bytes via `crypto.randomBytes`
- Byte 6 high nibble pinned to `7` (version field)
- Byte 8 high two bits pinned to `10` (variant field)

## Test changes

- **Unit (data-access)**: new `test/unit/util/uuid.test.js` proves version nibble = 7, variant bits 8/9/a/b, ms-encoded timestamp, time-sortable, unique across 1000 calls. `test/unit/models/base/schema.builder.test.js` now asserts the generated id's version nibble explicitly.
- **IT (data-access)**: added `expect(getId().charAt(14)).to.equal('7')` to create-path assertions in `site.test.js`, `opportunity.test.js`, `suggestion.test.js`, `report.test.js`, `contact-sales-lead.test.js`. These run against a real PostgREST + Postgres, so if anyone ever regresses the ORM to v4 the IT suite catches it.
- **Unit (utils)**: extended the `isValidUUID` test suite to cover v4 and v7 explicitly; added a test asserting `UUID_REGEX` and `UUID_V4_REGEX` are both exported and identity-equal.
- **Subpath barrel checks (utils)**: `test/index.test.js` and `test/subpaths.test.js` expected-exports lists were updated to include the two new public symbols.

## Verification

- ✅ `npm test -w packages/spacecat-shared-data-access` — 2231 passing, 100% coverage on the new `uuid.js`
- ✅ `npm test -w packages/spacecat-shared-utils` — 1154 passing
- ✅ `npm run lint` clean for both packages

## Compatibility

- **Validators stay agnostic.** `isValidUUID` and `z.uuid()` (Zod 4.1.13) both accept v4 *and* v7 — verified empirically. No callsite in `packages/spacecat-shared-utils/src/schemas.js` uses the strict `z.uuidv4()`, so they keep accepting both.
- **No DB migration.** Schema stays v7-by-design as it has been since Jan 2025.
- **No breaking change for existing consumers.** New rows minted by the ORM will be v7 going forward; the API contract is version-agnostic for any caller using `isValidUUID` or `z.uuid()`.

## Related

- SITES-45292 — the direct-apply e2e where this asymmetry was uncovered
- [adobe/spacecat-api-service#2496](https://github.com/adobe/spacecat-api-service/pull/2496) — companion gate-widening PR (already open, deployed via branch-deploy)
- Ticket Edit 3 (UI-side gate in `experience-success-studio-ui`) and Edit 5 (api-service test assertions) live in other repos and are out of scope for this PR

## Out of scope (worth flagging as follow-ups)

`sentiment-guideline.schema.js:45` and `sentiment-topic.schema.js:43` still use `crypto.randomUUID()` for *secondary* `text` columns (`guideline_id`, `topic_id`) that have no DB default. They're minting v4 for new rows. Not the bug from the ticket — these aren't `id` PK columns — but worth a follow-up if you want fleet-wide v7 alignment.

## Test plan

- [ ] CI green on lint + unit tests for both packages
- [ ] (optional) Run data-access IT tests locally — `npm run test:it -w packages/spacecat-shared-data-access` — to confirm v7 minting works against a real PostgREST + Postgres
- [ ] After release, verify in dev that newly-created sites/opportunities/suggestions have v7 ids (version nibble 7 at position 14)

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/base/schema.builder.js
+++ b/packages/spacecat-shared-data-access/src/models/base/schema.builder.js
@@ -15,6 +15,7 @@ import {
 } from '@adobe/spacecat-shared-utils';
 
 import { SchemaBuilderError } from '../../errors/index.js';
+import { uuidv7 } from '../../util/uuid.js';
 import {
   decapitalize,
   entityNameToAllPKValue,
@@ -29,17 +30,15 @@ import Schema from './schema.js';
 
 const DEFAULT_SERVICE_NAME = 'SpaceCat';
 
-/**
- * ID attribute configuration object.
- * Ensures a UUID-based "primary key".
- * @type {object}
- */
+// The DB column default is `uuid_generate_v7()` (mysticat-data-service migration
+// 20250130000001_initial_setup.sql). We mint v7 client-side so the in-memory
+// record matches what the DB would have minted — sortable, locality-friendly.
 const ID_ATTRIBUTE_DATA = {
   type: 'string',
   postgrestField: 'id',
   required: true,
   readOnly: true,
-  default: () => crypto.randomUUID(),
+  default: () => uuidv7(),
   validate: (value) => isValidUUID(value),
 };
 
--- a/packages/spacecat-shared-data-access/src/util/uuid.js
+++ b/packages/spacecat-shared-data-access/src/util/uuid.js
@@ -18,7 +18,9 @@ import { randomBytes } from 'crypto';
 // schema has been v7-by-design since Jan 2025; this keeps the ORM aligned.
 /* eslint-disable no-bitwise */
 function uuidv7() {
-  const bytes = Buffer.alloc(16);
+  // Every byte is overwritten below (0-5 from the timestamp, 6-15 from
+  // randomBytes), so allocUnsafe avoids the zero-fill that alloc would do.
+  const bytes = Buffer.allocUnsafe(16);
   const ms = BigInt(Date.now());
 
   bytes[0] = Number((ms >> 40n) & 0xffn);
--- a/packages/spacecat-shared-data-access/src/util/uuid.js
+++ b/packages/spacecat-shared-data-access/src/util/uuid.js
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { randomBytes } from 'crypto';
+
+// Mirrors the database's `uuid_generate_v7()` plpgsql function defined in
+// mysticat-data-service migration 20250130000001_initial_setup.sql so client-
+// minted ids share the same time-sortable layout as DB-defaulted rows. The
+// schema has been v7-by-design since Jan 2025; this keeps the ORM aligned.
+/* eslint-disable no-bitwise */
+function uuidv7() {
+  const bytes = Buffer.alloc(16);
+  const ms = BigInt(Date.now());
+
+  bytes[0] = Number((ms >> 40n) & 0xffn);
+  bytes[1] = Number((ms >> 32n) & 0xffn);
+  bytes[2] = Number((ms >> 24n) & 0xffn);
+  bytes[3] = Number((ms >> 16n) & 0xffn);
+  bytes[4] = Number((ms >> 8n) & 0xffn);
+  bytes[5] = Number(ms & 0xffn);
+
+  randomBytes(10).copy(bytes, 6);
+
+  bytes[6] = (bytes[6] & 0x0f) | 0x70;
+  bytes[8] = (bytes[8] & 0x3f) | 0x80;
+
+  const hex = bytes.toString('hex');
+  return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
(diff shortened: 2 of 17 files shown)


lucianfelix (dismissed):

MysticatBot-Dev (dismissed):
Hey @ekremney,

## Strengths

- **Correct and minimal `uuidv7()` implementation** (`src/util/uuid.js:20-38`): Clean RFC 9562 byte layout using `Buffer.alloc` + `crypto.randomBytes` + bitwise masking. Bit-mirrors the DB's plpgsql function so ids are interchangeable regardless of minting path. Readable enough to compare byte-for-byte against the DB function during future debugging.
- **Proper CSPRNG usage** (`src/util/uuid.js:30`): Uses `crypto.randomBytes(10)` for the random portion - the correct source for identifier generation. 80 bits of entropy per UUID exceeds collision resistance requirements.
- **Excellent deprecation strategy** (`functions.js:18-19`): `UUID_V4_REGEX = UUID_REGEX` as a same-reference alias means zero runtime cost, zero breakage for downstream consumers, and clear deprecation signal. This is the right way to rename a public constant without a semver-breaking bump.
- **Rolling forward rather than migrating** is the right architectural call. With 100+ partitioned analytics tables referencing `sites.id`, an in-place UUID swap would be high-risk for negligible benefit. v4 and v7 coexist cleanly under RFC 4122/9562.
- **Comprehensive test coverage** (`test/unit/util/uuid.test.js`): Property-based checks on format, version nibble, variant bits, timestamp encoding, time-sortability, and uniqueness. The timestamp bracket test (lines 42-51) is particularly well-constructed. IT assertions across 5 entity types provide regression safety against a revert to v4.
- **Change discipline is excellent**: Two logically distinct but tightly related changes, each independently shippable as stated. The PR description is exemplary - root cause, decision rationale, byte layout correspondence, and explicit out-of-scope items.

## Issues

### Minor (Nice to Have)

1. **`packages/spacecat-shared-data-access/src/util/uuid.js:21` - `Buffer.alloc(16)` zero-fills unnecessarily**: Every byte is explicitly written before being read (bytes 0-5 from timestamp, bytes 6-15 from `randomBytes.copy`), so the zero-fill is immediately undone. `Buffer.allocUnsafe(16)` is more precise about intent. Micro-optimization that doesn't matter in practice since ID generation isn't a hot loop, but it communicates "I will fill every byte myself."

2. **`test/unit/util/uuid.test.js:56-63` - Time-sortability test relies on wall-clock timing**: The `setTimeout(resolve, 2)` approach depends on the callback not firing within the same millisecond. On heavily loaded CI runners, timer resolution can compress. A fake clock advancing by 1ms would make this deterministic. Low flakiness risk given the 2ms margin, but worth noting.

## Recommendations

- **Longer-term: consider omitting `id` from the INSERT payload entirely.** If the ORM could skip `id` when no explicit value was provided by the caller, the DB's `uuid_generate_v7()` default would fire and there'd be no client-side implementation to keep in sync. This eliminates the maintenance surface of keeping two implementations bit-compatible. Not urgent given that RFC 9562 is frozen, but it's the simplest long-term state.
- **Sub-millisecond ordering (RFC 9562 Section 6.2):** The current implementation uses pure random bits for bytes 6-15, so two UUIDs minted in the same millisecond have random relative ordering. A monotonic counter would guarantee sort order within a tick. The DB-side function likely has the same behavior, so this is consistent - worth noting if high-throughput inserts matter for time-sorted queries.

## Assessment

Ready to merge? Yes

The implementation is correct, well-tested, backward-compatible, and well-scoped. No Critical or Important issues. CI should be confirmed green before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 44s | <b>Cost:</b> $3.97 | <b>Commit:</b> <code>f3dc5374b26c8d823e7c5e150d49a5fd13dc735a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1634, opened 2026-05-27
fix: (data-access) include all fixes in getAllFixesWithSuggestionsByOpportunityId (SITES-45307)

mimchome:
[SITES-45307](https://jira.corp.adobe.com/browse/SITES-45307)

## Summary

- Fix `getAllFixesWithSuggestionsByOpportunityId` to return all fixes, not just those with mapping table entries
- Fixes without `fix_entity_suggestions` entries are now included with `suggestions: []`
- 58% of fixes on dev had no mapping entries and were silently dropped

## The bug

```
fix_entities                     fix_entity_suggestions (mapping table)
+--------+----------------+     +----------------+----------------+
| id     | opportunity_id |     | fix_entity_id  | suggestion_id  |
+--------+----------------+     +----------------+----------------+
| fix-1  | opp-A          |     | fix-1          | sugg-1         |
| fix-2  | opp-A          |     | fix-1          | sugg-2         |
| fix-3  | opp-A          |     | fix-3          | sugg-3         |
+--------+----------------+     +----------------+----------------+
                                  fix-2 has NO row here

BEFORE: start from mapping table -> only see fix-1, fix-3 -> returns 2
AFTER:  start from fix_entities  -> see fix-1, fix-2, fix-3 -> returns 3
```

The old code queried the mapping table first, then batch-loaded fixes from those results. Fixes without mapping entries were invisible.

## What changed

- `getAllFixesWithSuggestionsByOpportunityId` now fetches fixes directly via `allByOpportunityId` in parallel with mapping records (2 calls via `Promise.all`)
- New `#buildAllFixesWithSuggestions` merges suggestions onto fixes; fixes without mappings get `suggestions: []`
- Existing `#buildFixesWithSuggestions` kept unchanged for `getAllFixesWithSuggestionByCreatedAt` (legitimately starts from mapping table)

## N+1 fix still applies

Before PR #1619, the controller called `getSuggestions()` per fix (N+1 queries). This method replaces that with 2-3 total calls regardless of fix count:

1. `allByOpportunityId` -- all fixes (1 call)
2. `allByIndexKeys` -- all mapping records (1 call)
3. `batchGetByKeys` -- all suggestions (1 call, skipped if no mappings)

Calls 1 and 2 run in parallel via `Promise.all`, so it's 2 round trips total.

## Test plan

- [x] 2 new unit tests: fixes without mappings, all fixes without mappings
- [x] Call-argument assertions on both new tests
- [x] 2227 tests passing, 100% line coverage, lint clean
- [ ] Smoke test after deploy: sunstar.com opp `78c9837b` should return 3 fixes (currently 2)


MysticatBot-Dev (approved):
Hey @mimchome,

## Strengths

- **Correct root-cause fix** (fix-entity.collection.js:214-218): Inverting the query axis to start from fix entities rather than the junction table is the right structural fix. Fixes without mapping entries can no longer be silently dropped.
- **Parallel fetch** (fix-entity.collection.js:214-218): `Promise.all` for fix entities and junction records keeps latency flat without added complexity. Follows the same pattern already used elsewhere in the class.
- **Smart guard on empty data** (fix-entity.collection.js:282, 296): Early return on empty `allFixEntities` and the `allSuggestionIds.length > 0` check before fetching suggestions avoids wasted network round-trips - especially valuable given 58% of fixes lack mappings.
- **Backward-compatible return shape**: The method signature and `{ fixEntity, suggestions }` contract are unchanged. Consumers that previously only saw a subset now see the full set, but the shape is stable.
- **Targeted test coverage** (fix-entity.collection.test.js:818-906): The two new tests directly validate the bug-fix scenarios (partial mappings, zero mappings) with clear assertions on both results and call arguments.

## Issues

### Minor (Nice to Have)

1. **fix-entity.collection.js:294 - Duplicate suggestion IDs passed to `batchGetByKeys`**

   `Object.values(suggestionIdsByFixEntityId).flat()` may contain duplicates if the junction table has redundant rows or if a suggestion is mapped to multiple fix entities within one opportunity. This results in redundant keys in the batch request.

   **Why it matters:** Performance - depending on the `batchGetByKeys` implementation, duplicates may cause redundant lookups or inflated payloads.

   **How to fix:**
   ```js
   const allSuggestionIds = [...new Set(Object.values(suggestionIdsByFixEntityId).flat())];
   ```

   Note: this is pre-existing in `#buildFixesWithSuggestions` (line 256), so not a regression - just a cheap improvement while writing the new method.

2. **fix-entity.collection.test.js:941-978 - `allByOpportunityId` rejection path not directly tested**

   The error-handling tests only exercise rejection from `allByIndexKeys` (with `allByOpportunityId` stubbed as `.resolves([])`). Since the method now uses `Promise.all`, a rejection from `allByOpportunityId` is a distinct trigger into the same catch block. The behavior is identical (the catch block does not distinguish the source), but exercising both rejection paths documents the contract more completely.

   **How to fix:** Add a test where `allByOpportunityId` rejects with a generic Error and verify it is wrapped in `DataAccessError`.

## Recommendations

- **Validate whether `getAllFixesWithSuggestionByCreatedAt` needs the same treatment.** The PR body states it "legitimately starts from mapping table," which makes sense if the use case is specifically "fixes that have suggestions, filtered by date." Worth confirming that downstream consumers of that method do not also expect completeness - if they do, a follow-up ticket would close the gap.
- **Consider extracting shared suggestion-resolution logic.** The new `#buildAllFixesWithSuggestions` shares ~80% of its logic with `#buildFixesWithSuggestions`. If suggestion-fetching logic changes in the future (pagination, partial-failure handling), it will need updating in two places. A shared helper that takes a list of fix entities and resolves their suggestions would reduce that surface.

## Assessment

Ready to merge? Yes
Reasoning: The fix is correct, well-scoped, and properly tested for the core scenarios. The minor items are defensive improvements, not correctness issues. The bug fixed here (58% of fixes silently dropped) is significant and this change resolves it cleanly.

PR #1640, opened 2026-05-30
feat(tokowaka-client): add [edge-optimize-status] prefix to log messages

jindaliiita:
## Summary

- Add `[edge-optimize-status]` structured prefix to all log messages in the `checkEdgeOptimizeStatus` method for easier log filtering/tracing
- Promote `response.status` and `edge optimize headers found` logs from `debug` to `info` level for better observability in production
- Include `targetUrl` in error/warn log messages for improved context

## Test Plan

- [x] All 698 existing tests pass with 100% coverage
- [x] No functional changes — log message format only

### DIFF
--- a/packages/spacecat-shared-tokowaka-client/src/index.js
+++ b/packages/spacecat-shared-tokowaka-client/src/index.js
@@ -1147,7 +1147,7 @@ class TokowakaClient {
     const baseURL = getEffectiveBaseURL(site);
     const targetUrl = new URL(path, baseURL).toString();
 
-    this.log.info(`Checking edge optimize status for ${targetUrl}`);
+    this.log.info(`[edge-optimize-status] Checking edge optimize status for ${targetUrl}`);
 
     const maxRetries = 3;
     let attempt = 0;
@@ -1156,7 +1156,7 @@ class TokowakaClient {
 
     while (attempt <= maxRetries) {
       try {
-        this.log.debug(`Attempt ${attempt + 1}/${maxRetries + 1}: Checking edge optimize status for ${targetUrl}`);
+        this.log.debug(`[edge-optimize-status] Attempt ${attempt + 1}/${maxRetries + 1}: Checking edge optimize status for ${targetUrl}`);
 
         // eslint-disable-next-line no-await-in-loop
         const response = await tracingFetch(targetUrl, {
@@ -1168,12 +1168,12 @@ class TokowakaClient {
           timeout: REQUEST_TIMEOUT_MS,
         });
 
-        this.log.debug(`Response status: ${response.status}`);
+        this.log.info(`[edge-optimize-status] Response status: ${response.status} for ${targetUrl}`);
 
         const edgeOptimizeEnabled = response.headers.get('x-tokowaka-request-id') !== null
           || response.headers.get('x-edgeoptimize-request-id') !== null;
 
-        this.log.debug(`Edge optimize headers found: ${edgeOptimizeEnabled}`);
+        this.log.info(`[edge-optimize-status] Edge optimize headers found: ${edgeOptimizeEnabled} for ${targetUrl}`);
 
         return {
           edgeOptimizeEnabled,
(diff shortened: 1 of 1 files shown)


packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1168,12 +1168,12 @@ class TokowakaClient {
           timeout: REQUEST_TIMEOUT_MS,
         });
 
-        this.log.debug(`Response status: ${response.status}`);
+        this.log.info(`[edge-optimize-status] Response status: ${response.status} for ${targetUrl}`);
```
MysticatBot-Dev, line 1171: Two lines promoted from `debug` to `info` here and at line 1177 will now emit on every retry attempt (up to 4x per invocation). In high-fan-out scenarios (batch audits across many sites), this could meaningfully increase info-level log volume and Coralogix ingestion cost.

Consider keeping per-attempt diagnostics at `debug` and emitting a single info-level summary after the retry loop resolves, or confirming that the team accepts this volume trade-off.

MysticatBot-Dev (changes requested):
Hey @jindaliiita,

## Strengths

- **Consistent structured prefix** (all log sites): The `[edge-optimize-status]` prefix is applied uniformly to every log statement in the method, making grep/filter queries in Coralogix reliable. This follows the established bracket-prefix pattern used elsewhere in the workspace (e.g., `[compression]` in the response-compression design).
- **Improved error/warn context** (lines ~1185, 1193, 1203): Adding `targetUrl` to the error and retry-warn messages that previously lacked it means an operator can immediately identify which site/path triggered the failure without correlating other log lines.
- **Well-scoped change**: The diff is focused, mechanical, and does not alter control flow or return values. All 698 existing tests pass with 100% coverage.

## Issues

### Important (Should Fix)

**Debug-to-info promotion creates per-request log volume in a retry loop** - `packages/spacecat-shared-tokowaka-client/src/index.js:1171`

Two log statements (`Response status: ...` and `Edge optimize headers found: ...`) were promoted from `debug` to `info`. This method retries up to 4 times per invocation, so a single call can emit up to 8 info-level lines (2 per attempt x 4 attempts). If `checkEdgeOptimizeStatus` is invoked at scale (e.g., for every site in a periodic audit), this meaningfully increases info-level log ingestion volume and cost in Coralogix, where info logs are retained and indexed by default while debug logs are often sampled or dropped.

**Why it matters:** Increased log volume drowns out actionable signals in dashboards and increases ingestion cost.

**Suggestion:** Keep the per-attempt `Response status` at `debug`. If the intent is observability of the final result, consider logging a single info line after the retry loop resolves successfully (combining status + edgeOptimizeEnabled), rather than on every attempt. Alternatively, if the team has confirmed this volume is acceptable for their use case, a brief note in the PR acknowledging the trade-off would be sufficient.

### Minor (Nice to Have)

**Trailing URL placement reads awkwardly in retry warning** - `packages/spacecat-shared-tokowaka-client/src/index.js:1203`

The message now reads: `"...Retrying in ${delay}ms... for ${targetUrl}"`. The `for ${targetUrl}` appended after the ellipsis reads as an afterthought. A small reorder would improve scanability:

```js
`[edge-optimize-status] Attempt ${attempt} to fetch ${targetUrl} failed: ${error.message}. Retrying in ${delay}ms...`
```

This puts the URL next to the action (fetch) rather than dangling at the end. Non-blocking.

## Assessment

Ready to merge? With fixes
Reasoning: The change is low-risk and well-scoped. The debug-to-info promotion is the one item worth confirming - if the team accepts the volume trade-off consciously, no code change is needed, just an acknowledgement.

## Next Steps

1. Confirm the debug-to-info promotion is acceptable given expected call volume, or keep per-attempt lines at debug.
2. Optionally reorder the retry-warn message for readability.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 26s | <b>Cost:</b> $2.55 | <b>Commit:</b> <code>b2bad97cd0b3a6072c780a74d817938a541006e4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1644, opened 2026-05-31
feat(consumer): add adminGrants field for per-operation admin-bypass grants

ravverma:
## Summary

- Adds nullable `adminGrants` field (type `any`) to the `Consumer` entity schema
- Adds `VALID_ADMIN_GRANTS = new Set(['CREATE_SITE'])` static constant on the model
- Adds `validateAdminGrants()` to `ConsumerCollection`: validates plain object, known keys, boolean-true values only; called on `create()` and `save()`
- Updates fixture and unit tests (model + collection)

## Context

Part of the `adminGrants` feature for S2S admin-gated write access. Design: [`platform/decisions/s2s-admin-grants-write-endpoints.md`](https://github.com/adobe/mysticat-architecture/pull/105).

**Dependency order:**
1. ✅ This PR — data model
2. `adobe/mysticat-data-service` — DB migration (nullable `admin_grants jsonb` column + CHECK constraint)
3. `adobe/spacecat-api-service` — `hasAdminGrant()`, `createSite` gate, consumers PATCH, DTO, S2S admin guide

## Test plan

- [ ] `consumer.model.test.js` — `VALID_ADMIN_GRANTS`, `getAdminGrants` getter, `validateAdminGrants` in `save()`
- [ ] `consumer.collection.test.js` — `validateAdminGrants` 7-case block, `create()` calls it

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21699,7 +21699,7 @@
     },
     "packages/spacecat-shared-athena-client": {
       "name": "@adobe/spacecat-shared-athena-client",
-      "version": "1.10.2",
+      "version": "1.10.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/spacecat-shared-utils": "1.96.3",
@@ -23384,7 +23384,7 @@
     },
     "packages/spacecat-shared-cloud-manager-client": {
       "name": "@adobe/spacecat-shared-cloud-manager-client",
-      "version": "1.4.2",
+      "version": "1.4.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
@@ -24355,7 +24355,7 @@
     },
     "packages/spacecat-shared-content-client": {
       "name": "@adobe/spacecat-shared-content-client",
-      "version": "1.9.2",
+      "version": "1.9.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/helix-universal": "5.4.2",
@@ -25162,12 +25162,12 @@
     },
     "packages/spacecat-shared-data-access": {
       "name": "@adobe/spacecat-shared-data-access",
-      "version": "3.73.2",
+      "version": "3.73.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "^4.2.3",
         "@adobe/spacecat-shared-utils": "1.105.0",
-        "@aws-sdk/client-s3": "^3.1009.0",
+        "@aws-sdk/client-s3": "^3.940.0",
         "@supabase/postgrest-js": "2.106.2",
         "@types/joi": "17.2.3",
         "aws-xray-sdk": "3.12.0",
@@ -25360,7 +25360,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/spacecat-shared-utils": "1.98.1",
-        "@aws-sdk/client-s3": "^3.1000.0",
+        "@aws-sdk/client-s3": "^3.775.0",
         "@aws-sdk/client-sns": "^3.775.0"
       },
       "devDependencies": {
@@ -25563,7 +25563,7 @@
     },
     "packages/spacecat-shared-google-client": {
       "name": "@adobe/spacecat-shared-google-client",
-      "version": "1.6.2",
+      "version": "1.6.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
@@ -27340,7 +27340,7 @@
     },
     "packages/spacecat-shared-http-utils": {
       "name": "@adobe/spacecat-shared-http-utils",
-      "version": "1.28.1",
+      "version": "1.28.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
@@ -28182,7 +28182,7 @@
     },
     "packages/spacecat-shared-ims-client": {
       "name": "@adobe/spacecat-shared-ims-client",
-      "version": "1.13.2",
+      "version": "1.13.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
@@ -28987,7 +28987,7 @@
     },
     "packages/spacecat-shared-launchdarkly-client": {
       "name": "@adobe/spacecat-shared-launchdarkly-client",
-      "version": "1.2.1",
+      "version": "1.2.2",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/helix-universal": "5.4.2",
@@ -33813,7 +33813,7 @@
     },
     "packages/spacecat-shared-tokowaka-client": {
       "name": "@adobe/spacecat-shared-tokowaka-client",
-      "version": "1.17.2",
+      "version": "1.17.3",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/spacecat-shared-utils": "1.81.1",
@@ -34622,7 +34622,7 @@
     },
     "packages/spacecat-shared-utils": {
       "name": "@adobe/spacecat-shared-utils",
-      "version": "1.116.4",
+      "version": "1.116.5",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
--- a/packages/spacecat-shared-data-access/src/models/consumer/consumer.collection.js
+++ b/packages/spacecat-shared-data-access/src/models/consumer/consumer.collection.js
@@ -70,37 +70,6 @@ class ConsumerCollection extends BaseCollection {
   }
 
   /**
-   * Validates that the given adminGrants map contains only known operation keys with boolean true
-   * values. Called during both create() and save() to prevent invalid admin grant escalation.
-   * @param {object|null|undefined} adminGrants - The adminGrants map to validate.
-   * @throws {ValidationError} - Throws if adminGrants is invalid.
-   */
-  validateAdminGrants(adminGrants) {
-    if (adminGrants == null) {
-      return;
-    }
-
-    if (typeof adminGrants !== 'object' || Array.isArray(adminGrants)) {
-      throw new ValidationError('adminGrants must be a plain object', this);
-    }
-
-    for (const [key, value] of Object.entries(adminGrants)) {
-      if (!Consumer.VALID_ADMIN_GRANTS.has(key)) {
-        throw new ValidationError(
-          `Invalid admin grant key: "${key}". Valid keys: [${[...Consumer.VALID_ADMIN_GRANTS].join(', ')}]`,
-          this,
-        );
-      }
-      if (value !== true) {
-        throw new ValidationError(
-          `adminGrants values must be boolean true. Got "${key}": ${JSON.stringify(value)}`,
-          this,
-        );
-      }
-    }
-  }
-
-  /**
    * Validates that the given imsOrgId is in the allowed list from config.
    * @param {string} imsOrgId - The IMS Org ID to validate.
    * @throws {ValidationError} - Throws if the imsOrgId is not in the allowed list.
@@ -161,7 +130,6 @@ class ConsumerCollection extends BaseCollection {
   async create(item, options = {}) {
     this.#validateImsOrgId(item?.imsOrgId);
     this.validateCapabilities(item?.capabilities);
-    this.validateAdminGrants(item?.adminGrants);
     await this.#validateClientIdUniqueness(item?.clientId);
     return super.create(item, options);
   }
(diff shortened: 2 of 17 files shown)


MysticatBot-Dev (approved):
Hey @ravverma,

## Strengths

- `consumer.collection.js:71-101`: Defensive validation with specific, actionable error messages for every failure path (wrong type, unknown key, non-true value). The allowlist pattern with strict `value !== true` enforcement is the correct security posture for a privilege-granting field.
- `consumer.model.js:61`: `VALID_ADMIN_GRANTS` as a static `Set` mirrors the established `CAPABILITIES` pattern - O(1) lookup, clear semantics, easy to extend. The team can reason about both concepts identically.
- Positive-only grant semantics (only `true` accepted, never `false`) avoids the entire class of allow-vs-deny precedence bugs that plague mixed-mode permission maps.
- Validation fires on both `create()` and `save()` paths, consistent with the existing `validateCapabilities` pattern. Prevents grant escalation via post-creation mutation.
- The nullable field with no `required` constraint means this PR introduces no runtime risk until the full dependency chain is complete.
- `consumer.collection.test.js:278-321`: Direct unit tests cover the full boundary space for the validation method.

## Issues

### Minor (Nice to Have)

1. **consumer.collection.test.js:206 - Test does not verify what its name claims**

   The test `'calls validateAdminGrants during create'` asserts only `expect(result).to.not.be.null`. It proves create succeeds with valid data, not that validation actually ran. If someone removes the `this.validateAdminGrants(item?.adminGrants)` line from `create()`, this test still passes. Consider either renaming it (e.g., `'create succeeds with valid adminGrants'`) or adding a negative case that passes invalid adminGrants to `create()` and expects a `ValidationError` - the latter would catch a regression.

2. **consumer.schema.js:50 - No schema-level type guard on `adminGrants`**

   The attribute is `type: 'any'` with no `validate` callback. While application-level validation in `validateAdminGrants()` is thorough, a minimal schema-level guard would provide defense-in-depth against writes that bypass `create()`/`save()` (bulk imports, direct PostgREST calls). A lightweight option:
   ```js
   .addAttribute('adminGrants', {
     type: 'any',
     validate: (value) => value == null || (typeof value === 'object' && !Array.isArray(value)),
   })
   ```

3. **consumer.collection.test.js - Missing empty-object edge case**

   `validateAdminGrants({})` passes silently (the loop iterates zero times). If that is intentional (semantically "no grants"), a one-line test documenting it clarifies the contract. If it should be rejected (since `null` already means "no grants"), that is a production logic gap worth discussing.

## Recommendations

- Consider whether `VALID_ADMIN_GRANTS` will need to be environment-aware as the feature grows (e.g., grants available in staging before production). Today's static Set is correct for a single grant, but if this grows to 10+ grants with different rollout timelines, a config-driven registry would scale better.
- Consider `Object.freeze(new Set(['CREATE_SITE']))` for `VALID_ADMIN_GRANTS` to prevent runtime mutation of the privilege surface. This follows the same mutable pattern as `CAPABILITIES` so it is consistent with existing convention - just a hardening opportunity.

## Assessment

Ready to merge? Yes

Clean, well-scoped data-model extension that follows existing entity patterns precisely. Validation is allowlist-based, strict on types and values, and enforced on both write paths. The authorization enforcement (checking grants at request time) is correctly deferred to the API service layer. No architectural risk, no security concerns at this layer.

Note: CI checks are still pending - confirm they pass before merging.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 57s | <b>Cost:</b> $2.75 | <b>Commit:</b> <code>820e8ff60938f236af58804ce560dcf3901031f4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot-Dev (approved):
Hey @ravverma,

## Strengths

- Previously flagged issues now addressed: all three prior findings implemented cleanly in the latest commit. Test renamed to `'create succeeds with valid adminGrants'`, schema-level validate callback added (`consumer.schema.js:50`), and empty-object edge case test documents the intentional `{}` behavior.
- `consumer.model.js:64`: `ADMIN_GRANTS` as a frozen object with identity-mapped values and `VALID_ADMIN_GRANTS` derived from it is the correct layering. Consuming services import the constant rather than redeclaring magic strings. Clean extension point that mirrors the existing `CAPABILITIES` pattern.
- `consumer.collection.js:72-99`: Validation is well-structured with early returns, strict allowlist enforcement, positive-only semantics (`value !== true`), and specific error messages that include the invalid input and enumerate valid options. The type guard before `Object.entries` prevents prototype pollution vectors.
- `consumer.collection.js:163`: Validation ordering in `create()` is logical - cheap structural checks (imsOrgId, capabilities, adminGrants) before the async uniqueness check. Fail fast on cheap operations.
- `consumer.model.test.js:209`: The assertion that `patcher.save` was NOT called after validation failure confirms the early-exit contract. This is the assertion that actually guards against regression.
- The dependency ordering (data model first, DB migration second, API gate third) is sound. This PR introduces no runtime behavior change until downstream services ship - zero production risk on merge.
- Follows the documented entity pattern precisely: SchemaBuilder DSL, dual write-path validation, and multi-repo coordination per the package CLAUDE.md.

## Issues

### Minor (Nice to Have)

1. **consumer.model.test.js:83-87 - Tautological test on ADMIN_GRANTS constant**

   `expect(Consumer.ADMIN_GRANTS).to.deep.equal({ CREATE_SITE: 'CREATE_SITE' })` re-declares the constant's value and asserts equality. If someone changes the production constant, they change this test in lockstep - it cannot fail in a way that finds a bug. The more useful test already exists at line 89: `VALID_ADMIN_GRANTS` derived from `ADMIN_GRANTS` proves the derivation works. Consider replacing with a structural assertion (`Object.isFrozen`, keys are non-empty strings) or removing it since line 89 transitively proves the constant exists and has the expected shape.

2. **consumer.collection.test.js:220 - Weak assertion on create success path**

   `expect(result).to.not.be.null` would pass even if `create()` returned an empty object or string. Asserting on a property of the returned model (e.g., `expect(result.getAdminGrants()).to.deep.equal({ CREATE_SITE: true })`) would prove the data flows through. Low-priority since the negative companion test at line 224 already guards against regression.

## Recommendations

- When the second grant key arrives in `VALID_ADMIN_GRANTS`, consider adding a multi-key test (`{ CREATE_SITE: true, NEW_OP: true }`) to verify the validation loop handles more than one entry. Not needed today with a single key.
- Consider whether `ADMIN_GRANTS` should eventually live in a shared constants export from the package index rather than only on the Consumer class. If other entities gain admin-grant-gated operations, a single-entity registry will feel awkward. Not urgent for one grant, but worth a brief conversation before the third arrives.

## Assessment

Ready to merge? Yes

Clean, well-scoped data-model extension that follows every established pattern in the entity system. Validation is allowlist-based, strict on types and values, and enforced on both write paths. No architectural risk, no security concerns, and all prior review findings addressed precisely. The author's pushback on Object.freeze was technically correct - dropped.

Note: CI test check is still pending - confirm it passes before merging.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 38s | <b>Cost:</b> $2.36 | <b>Commit:</b> <code>f1f04395d1de9c4a8341c28cc220f4e191dae963</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot-Dev (approved):
Hey @ravverma,

## Strengths

- Previously flagged issues now addressed cleanly: the tautological `ADMIN_GRANTS` assertion (`consumer.model.test.js:83-93`) replaced with structural checks (`Object.isFrozen`, non-empty keys, identity-mapped values), and the create success path (`consumer.collection.test.js:222`) now asserts `result.getAdminGrants()` returns the expected value rather than just non-null.
- `consumer.collection.js:71-99`: Validation method is well-structured with early returns, type guards before iteration, and specific error messages that include both the invalid input and the set of valid options. Mirrors `validateCapabilities` exactly.
- `consumer.model.js:64-70`: `ADMIN_GRANTS` as a frozen identity-mapped object with `VALID_ADMIN_GRANTS` derived from it is the right layering - single source of truth, clear derivation, consuming services import constants rather than redeclaring strings.
- Positive-only grant semantics (`value !== true` enforcement) avoids the entire class of allow/deny precedence bugs. Sound security posture for a privilege-granting field.
- Validation enforced on both `create()` and `save()` paths prevents grant escalation via post-creation mutation. Consistent with the existing `validateCapabilities` pattern.
- Schema-level validate callback (`consumer.schema.js:50-53`) provides defense-in-depth without duplicating full business logic from the collection validator.
- `consumer.model.test.js:195-210`: The negative-path test asserts both the rejection message AND that `patcher.save` was NOT called - the high-value regression guard that proves the early-exit contract.
- The multi-repo dependency ordering (data model -> DB migration -> API gate) is correct and explicitly documented. This PR introduces no runtime behavior change on its own.

## Issues

### Minor (Nice to Have)

1. **consumer.model.test.js:183-191 - Interaction-only assertion on save happy path**

   The test `'calls this.collection.validateAdminGrants with adminGrants on save'` stubs `validateAdminGrants` and asserts only `.calledOnceWith(...)`. This is interaction verification on an internal collaborator. It cannot detect a regression where `validateAdminGrants` is called but its result is ignored, or where the method signature changes but still passes.

   The companion negative test (line 195) already covers the critical path (validation failure prevents save), which is the higher-value assertion. Not worth blocking since it mirrors the existing `validateCapabilities` pattern in the same file - just noting that the pattern is interaction-heavy.

## Assessment

Ready to merge? Yes

Clean, well-scoped data-model extension that follows every established entity pattern. Validation is allowlist-based, strict on types and values, and enforced on both write paths. No architectural concerns, no security issues, no convention violations. All prior review findings addressed precisely. The authorization enforcement (checking grants at request time) is correctly deferred to the API service layer.

Note: CI test check is still pending - confirm it passes before merging.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 27s | <b>Cost:</b> $2.39 | <b>Commit:</b> <code>1cb32e91c67b63b834bf4d9bda6a5d535cc67519</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @ravverma,

## Strengths

- `consumer.model.js:57`: The design pivot from a separate `adminGrants` authorization surface to extending the existing `CAPABILITIES` model is the right call. One authorization model is strictly simpler than two, and the `site:readAll` precedent proves the pattern already handles admin-gated operations. This eliminates a DB migration, a new validation layer, and a new API surface in downstream services.
- `consumer.model.js:54-56`: The updated JSDoc is precise and actionable - it names the concrete route (`POST /sites`), explains the gating semantics, and retains the important caveat that schema validity does not imply a reachable route.
- The diff is focused and backward-compatible. One conceptual change, production file and tests updated in lockstep. No existing capability string changes meaning, no consumers break, no migration required. The new `create` entry is inert until the API service enforces it.
- The multi-repo dependency chain collapsed from 3 coordinated PRs (shared model + DB migration + API gate) to 2 (shared model + API gate). Meaningful operational simplification.
- Follows all documented conventions: entity pattern structure, test mirroring, conventional commits, fixture in sync.

## Issues

### Minor (Nice to Have)

1. **PR title and description are stale after design pivot**

   The title is "add adminGrants field for per-operation admin-bypass grants" and the body describes `adminGrants`, `VALID_ADMIN_GRANTS`, and `validateAdminGrants()`. The actual diff adds `'create'` to `CAPABILITIES` with no adminGrants machinery. Since this repo uses semantic-release to auto-generate CHANGELOG.md from merge commit messages, the current title will produce a misleading changelog entry.

   **Fix:** Update PR title to something like `feat(consumer): add 'create' capability for admin-gated write operations` and update the body to reflect the current implementation.

2. **consumer.model.test.js:65-67 - Redundant test assertion**

   The new test `'has CAPABILITIES including create'` can never fail when the `deep.equal` assertion on line 63 passes. `deep.equal(['read', 'write', 'delete', 'readAll', 'create'])` already proves `create` is present. The `.include` test is a strict subset of information already asserted three lines above.

   **Fix:** Remove the `'has CAPABILITIES including create'` test block.

## Assessment

Ready to merge? Yes

Clean, minimal extension of an established pattern. The design pivot eliminated accidental complexity, and the implementation follows the existing `CAPABILITIES` convention precisely. Neither minor item is worth blocking on.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 27s | <b>Cost:</b> $1.86 | <b>Commit:</b> <code>30b53a852999de18c53545028dd557571e1f8c43</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1646, opened 2026-06-01
fix: tighten buildUrlMatcher boundary check and add dedicated unit tests

ssilare-adobe:
## Summary

Followup to #1598 addressing reviewer feedback from the [PR review](https://github.com/adobe/spacecat-shared/pull/1598#pullrequestreview-4402247594).

### Changes

**`src/utils/pattern-utils.js`** — Fix `/*` prefix boundary over-matching + document regex contract

`/products/*` previously matched `/productsabc` via `pathname.startsWith('/products')`. Fixed to require either an exact match or a `/`-separated child: `pathname === prefix || pathname.startsWith(prefix + '/')`.

Regex patterns continue to match against the full URL string for backward compatibility with callers using full-URL patterns like `^https://example\.com/.*`. The JSDoc now documents this contract explicitly.

**`test/utils/pattern-utils.test.js`** — New dedicated unit test file covering `/*`, `/path/*`, boundary safety, regex behavior, and invalid inputs.

**`test/utils/metaconfig-utils.test.js`** — New dedicated unit test file for `addPatternsToMetaconfig` and `removePatternFromMetaconfig`.

### What was NOT changed

- **Rollback only removes `allowedRegexPatterns[0]`** — Not a bug. Path suggestions always have exactly one `allowedRegexPatterns` entry by design.
- **Sibling keys on `metaconfig.prerender`** — No sibling keys exist currently, so no spread is needed. Can be added when the schema grows.

## Test plan

- [x] `npm test` in `packages/spacecat-shared-tokowaka-client` — 765 passing, 100% coverage on both utils files

### DIFF
--- a/packages/spacecat-shared-tokowaka-client/src/utils/metaconfig-utils.js
+++ b/packages/spacecat-shared-tokowaka-client/src/utils/metaconfig-utils.js
@@ -12,7 +12,8 @@
 
 /**
  * Removes a single pattern from the metaconfig's prerender allowList in-place.
- * Deletes the prerender key entirely when the resulting list would be empty.
+ * Deletes the prerender key entirely when the resulting list would be empty and no sibling
+ * keys exist on metaconfig.prerender. When sibling keys exist, only the allowList key is removed.
  * @param {Object} metaconfig - Metaconfig object (mutated in place)
  * @param {string} pattern - Pattern to remove
  * @returns {boolean} True if the allowList was changed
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @ssilare-adobe,

## Strengths

- `src/utils/pattern-utils.js:36-37` - The three-way boundary condition (`prefix === '' || pathname === prefix || pathname.startsWith(`${prefix}/`)`) is the correct fix. It cleanly handles empty prefix (domain-wide), exact match, and child-path cases without introducing additional state or abstractions.
- `src/utils/pattern-utils.js:14-25` - The JSDoc rewrite is genuinely useful. It documents the dual-mode contract (pathname matching for `/*` patterns, full-URL matching for regex), calls out the foot-gun (pathname-only regex will never match via the regex path), and specifies failure modes.
- `test/utils/pattern-utils.test.js:97-99` - The boundary safety test (`/productsabc`, `/products-new`) directly exercises the exact bug being fixed. Strong regression anchor.
- `test/utils/pattern-utils.test.js:89-95` - Tests both `pathname === prefix` (no trailing slash) and `pathname.startsWith(prefix + '/')` (with trailing slash) independently, verifying both new branches.
- `test/utils/metaconfig-utils.test.js:101-105` - Tests the cleanup behavior (deleting `prerender` when allowList empties), covering a side-effect that could regress silently.
- Security-positive change: closes a real path-boundary over-match while maintaining the safe-by-default posture (invalid URLs denied via catch, `new URL()` handles encoding normalization).

## Issues

### Minor (Nice to Have)

1. `src/utils/pattern-utils.js:16` - The JSDoc examples list `/products`, `/products/`, `/products/item` as matches but the bare path without trailing slash (`/products`) is easy to miss in the list. Consider making it visually distinct (e.g., on its own bullet) since it is the semantic expansion callers may not expect from a `/*` glob pattern.

2. `test/utils/pattern-utils.test.js` - No test for a multi-segment prefix pattern like `/a/b/*`. The logic is identical (slice off `/*`, compare), but a test would document that nested prefix patterns work and catch any future regression on intermediate slashes.

3. `test/utils/metaconfig-utils.test.js` - No test for `addPatternsToMetaconfig(metaconfig, [])` (empty array). Documenting the expected behavior for this edge case would prevent future ambiguity about whether it returns true/false or creates the `prerender.allowList` structure.

## Recommendations

- Consider one test for a URL with query string or fragment against a prefix pattern (e.g., `https://example.com/products?page=1` against `/products/*`). The `new URL()` parser handles this correctly (pathname excludes query/fragment), but a test would document that guarantee for future maintainers.

## Assessment

Ready to merge? Yes

The fix is correct, well-scoped, and properly tested. The boundary check now enforces path-segment boundaries instead of naive prefix matching. Tests directly cover the regression case, exercise both new branches, and handle invalid inputs. The change is backward-compatible - all previously-correct matches still work, only the over-match is eliminated. CI passes with 100% coverage.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 44s | <b>Cost:</b> $2.58 | <b>Commit:</b> <code>1fec3e07954f348679ba8f193abb37bf32933991</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1647, opened 2026-06-02
fix(tokowaka-client): use scraperConfig headers in edge optimize probe (LLMO-5280)

jindaliiita:
## Summary

- Read probe headers from `site.getConfig().getScraperConfig().headers` and merge them under the hard-coded defaults in `checkEdgeOptimizeStatus()`, so customers can add allowlist-friendly headers (e.g. a shared-secret or custom `User-Agent`) via their `scraperConfig` without any service-side code change
- Default probe headers always win on collision — `Tokowaka-AI` `User-Agent` and `fastly-debug` can never be overridden by custom config
- Default headers defined inline in the function (no module-level constant)
- Incorporates log improvements from PR #1640: `[edge-optimize-status]` structured prefix on all log messages; promotes response-status and headers-found logs from `debug` to `info` for reliable Coralogix traceability

## Root cause (LLMO-5280)

The `checkEdgeOptimizeStatus` probe used a hard-coded `User-Agent` containing `Tokowaka-AI`/`AdobeEdgeOptimize-AI`. Customer WAF/CDN rules (Akamai on careinsurance.com) blocked the probe with 403, causing the SpaceCat API to return `edgeOptimizeEnabled: false` even when edge optimize was active.

## Test plan

- [x] 769 existing tests pass with 100% line/branch/function/statement coverage
- [x] New tests: custom headers added, default wins on collision, empty scraperConfig, missing `getScraperConfig` — all verified

### DIFF
--- a/packages/spacecat-shared-tokowaka-client/src/index.js
+++ b/packages/spacecat-shared-tokowaka-client/src/index.js
@@ -1287,10 +1287,18 @@ class TokowakaClient {
       return { edgeOptimizeEnabled: true };
     }
 
+    const defaultProbeHeaders = {
+      // eslint-disable-next-line max-len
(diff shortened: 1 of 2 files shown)


sneharora1 (approved):
LGTM

PR #1648, opened 2026-06-03
feat(tokowaka-client): mark path-level suggestions as coveredByDomainWide on domain-wide deploy

ssilare-adobe:
## Summary

- When a domain-wide pattern suggestion is deployed, path-level pattern suggestions (those with `allowedRegexPatterns` but not `isDomainWide`) are now marked with `coveredByDomainWide: <suggestionId>`, since a domain-wide deployment covers all path-level patterns
- When the domain-wide suggestion is rolled back, `coveredByDomainWide` is stripped from those path-level suggestions (the existing rollback cleanup logic already handles this correctly)
- Already-deployed (`edgeDeployed`) path-level suggestions and other domain-wide suggestions are not marked

## Changes

**`src/index.js`** — `#deployPatternSuggestion`: reordered the `edgeDeployed` guard before `isPatternSuggestion`, and changed the pattern suggestion exclusion to allow path-level suggestions through when deploying domain-wide

**`test/index.test.js`** — two new deploy tests (path-level gets marked, already-deployed path-level is skipped) and one new rollback test (path-level `coveredByDomainWide` is cleaned up on DW rollback)

## Test plan

- [x] All 776 tests passing
- [x] 100% line/branch/function/statement coverage maintained

### DIFF
--- a/packages/spacecat-shared-tokowaka-client/src/index.js
+++ b/packages/spacecat-shared-tokowaka-client/src/index.js
@@ -1532,11 +1532,13 @@ class TokowakaClient {
       if (!isEdgeDeployableSuggestionStatus(s.getStatus())) {
         return false;
       }
-      if (isPatternSuggestion(s)) {
-        return false;
-      }
       if (s.getData()?.edgeDeployed) {
         return false;
+      }
+      // Path-level pattern suggestions (not domain-wide) are fully covered by a
+      // domain-wide deployment. Other pattern suggestions (including other DW ones) are not.
+      if (isPatternSuggestion(s)) {
+        return coverageField === 'coveredByDomainWide' && !s.getData()?.isDomainWide;
       }
       const url = s.getData()?.url;
       return url && matchers.some((match) => match(url));
--- a/packages/spacecat-shared-tokowaka-client/test/index.test.js
+++ b/packages/spacecat-shared-tokowaka-client/test/index.test.js
@@ -3117,6 +3117,50 @@ describe('TokowakaClient', () => {
       expect(coveredData).to.not.have.property('coveredByDomainWide');
     });
 
+    it('cleans up coveredByDomainWide on path-level suggestions when rolling back domain-wide', async () => {
+      const prerenderOpportunity = { getId: () => 'opp-dw', getType: () => 'prerender' };
+      const dwSuggestion = {
+        getId: () => 'dw-1',
+        getData: () => ({
+          isDomainWide: true,
+          allowedRegexPatterns: ['/*'],
+          edgeDeployed: Date.now(),
+        }),
+        setData: sinon.stub(),
+        setUpdatedBy: sinon.stub(),
+        save: sinon.stub().resolves(),
+      };
+      const pathSuggestion = {
+        getId: () => 'path-1',
+        getData: () => ({
+          allowedRegexPatterns: ['/products/*'],
+          coveredByDomainWide: 'dw-1',
+        }),
+        setData: sinon.stub(),
+        setUpdatedBy: sinon.stub(),
+        save: sinon.stub().resolves(),
+      };
+
+      sinon.stub(client, 'fetchMetaconfig').resolves({
+        siteId: 'site-123',
+        prerender: { allowList: ['/*'] },
+      });
+      sinon.stub(client, 'uploadMetaconfig').resolves();
+
+      const result = await client.rollbackSuggestions(
+        mockSite,
+        prerenderOpportunity,
+        [dwSuggestion],
+        { allSuggestions: [dwSuggestion, pathSuggestion], updatedBy: 'test@example.com' },
+      );
+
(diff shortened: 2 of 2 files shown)


MysticatBot (commented):
Hey @ssilare-adobe,

## Strengths

- **Guard reordering is correct and intentional** (src/index.js:1535): Moving the `edgeDeployed` check above the `isPatternSuggestion` check ensures already-deployed suggestions are excluded universally, and the pattern-specific branch only evaluates non-deployed suggestions. Clean precedence hierarchy.
- **Reuses existing coverage infrastructure** (src/index.js:1538): The change leverages the existing `coverageField` discriminator and filter-mark-save pipeline rather than introducing a parallel mechanism. Path-level pattern suggestions enter the same `covered` array and follow the same lifecycle as per-URL suggestions.
- **Rollback symmetry without new code**: Because the PR uses the same `coveredByDomainWide` field with the suggestion ID as the value, the existing rollback cleanup loop (which filters by `s.getData()?.[coverageField] === suggestion.getId()`) works without modification.
- **Good explanatory comment** (src/index.js:1536-1537): Concisely states the business rule for a reader landing here cold.
- **Tests map directly to described behaviors**: Three new tests cover the three scenarios from the PR description - marking on deploy, skipping already-deployed, and cleanup on rollback. Each verifies both the coverage list membership and the data mutation.

## Issues

### Minor (Nice to Have)

**test/index.test.js:3120 - Rollback test does not assert `save()` was called on the cleaned-up path suggestion**

The test verifies `pathSuggestion.setData` was called with the correct data but does not assert `pathSuggestion.save` was called. If a regression caused the cleanup to compute the right data but skip persistence, this test would still pass.

How to fix: Add `expect(pathSuggestion.save.calledOnce).to.be.true;` after the existing setData assertions.

This is low-risk given the existing rollback tests likely cover the save path generically, but it would make this test self-contained.

## Assessment

Ready to merge? With minor suggestions
Reasoning: The change is focused, correct, and well-tested. It extends the existing coverage-marking mechanism to a new relationship type without introducing new state dimensions, failure modes, or operational burden. No security concerns - the change operates entirely on internal suggestion state with no trust boundary crossings. The single Minor finding is a test-completeness nit, not a blocker.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 34s | <b>Cost:</b> $2.84 | <b>Commit:</b> <code>2c10fed73863b5d87d230ebb57d196ecb006f835</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @ssilare-adobe,

## Strengths

- **Prior review finding addressed** (test/index.test.js:3163-3167): The `saveMany` persistence assertion was added as suggested, confirming that cleaned-up path suggestions are persisted, not just mutated in memory.
- **Correct guard reordering** (src/index.js:1535): Moving the `edgeDeployed` check above `isPatternSuggestion` ensures already-deployed suggestions are excluded universally before the pattern-specific branch evaluates. Clean precedence hierarchy.
- **Tight, self-contained logic** (src/index.js:1540): The condition `coverageField === 'coveredByDomainWide' && !s.getData()?.isDomainWide` is a single return statement that always evaluates for pattern suggestions - it returns `true` only for path-level patterns in domain-wide flows, and `false` in all other cases. No fall-through to URL matching is possible.
- **Reuses existing rollback infrastructure**: The `coveredByDomainWide` field with the suggestion ID as the value plugs directly into the existing rollback cleanup loop without new code.
- **Good inline comment** (src/index.js:1537-1538): Explains the business rule concisely without restating the code.
- **Proportional test coverage**: Three new tests map directly to the three behavioral branches - mark on deploy, skip already-deployed, and cleanup on rollback.

## Recommendations

- **Consider confirming domain-wide-vs-domain-wide test coverage**: The `!s.getData()?.isDomainWide` guard excludes other domain-wide suggestions from being marked as covered. The existing test "should not mark other domain-wide suggestions as covered" (visible at line 5909 context with `dw2`) likely exercises this, but if it only tests non-NEW status rather than the `isDomainWide` flag specifically, a targeted test would harden that branch.

## Assessment

Ready to merge? Yes
Reasoning: The change is focused, correct, and well-tested. It extends the existing coverage-marking mechanism to a new relationship type without introducing new failure modes or operational burden. The prior review's single Minor finding has been addressed. No security, architectural, or convention concerns.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 1s | <b>Cost:</b> $2.23 | <b>Commit:</b> <code>63d3c8512a46e173de96bd13cb98431fbba162a1</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1649, opened 2026-06-04
fix: tighten press/click.*hold regex to prevent false positives on normal page content

ssilare-adobe:
## Summary

- `CHALLENGE_PATTERNS.general` contained `/press.*hold/i` and `/click.*hold/i` with unbounded `.*` wildcards
- On sites with minified HTML, these matched across thousands of characters — e.g. `blood pressure monitor` (index 281,995) to `placeholder graphic` (index 291,537) on athenahealth.com — causing a false-positive `crawlable: false` result despite the site returning HTTP 200 with no challenge page
- Tightened both patterns to require the words to be adjacent (with optional "and"): `/press\s+(and\s+)?hold/i` — matching real challenge copy like "Press and hold the button" while rejecting incidental word proximity

## Changes

- `src/bot-blocker-detect/bot-blocker-detect.js`: narrow `press.*hold` and `click.*hold` patterns
- `test/bot-blocker-detect/bot-blocker-detect.test.js`: add regression test covering `blood pressure` + `placeholder` content

## Test plan

- [ ] All 1155 existing tests continue to pass
- [ ] New regression test `does not false-positive on "pressure" and "placeholder" in normal page content` passes
- [ ] Real "Press and hold" and "Click and hold" challenge strings still detected correctly

### DIFF
--- a/packages/spacecat-shared-utils/src/bot-blocker-detect/bot-blocker-detect.js
+++ b/packages/spacecat-shared-utils/src/bot-blocker-detect/bot-blocker-detect.js
@@ -103,8 +103,8 @@ const CHALLENGE_PATTERNS = {
     /hcaptcha/i,
     /datadome/i,
     /dd-request-id/i,
-    /press.*hold/i, // Press and hold challenges (hCaptcha, custom implementations)
-    /click.*hold/i, // Click and hold button challenges
+    /press\s+(and\s+)?hold/i, // Press and hold challenges (hCaptcha, custom implementations)
+    /click\s+(and\s+)?hold/i, // Click and hold button challenges
     /geetest/i, // GeeTest interactive challenges
     /arkose/i, // Arkose Labs bot protection
     /funcaptcha/i, // FunCaptcha interactive challenges
--- a/packages/spacecat-shared-utils/test/bot-blocker-detect/bot-blocker-detect.test.js
+++ b/packages/spacecat-shared-utils/test/bot-blocker-detect/bot-blocker-detect.test.js
@@ -959,6 +959,20 @@ describe('Bot Blocker Detection', () => {
       expect(result.confidence).to.equal(0.7);
     });
 
+    it('does not false-positive on "pressure" and "placeholder" in normal page content', () => {
+      const html = '<html><body><p>Monitor your blood pressure at home.</p><input placeholder="Enter value"></body></html>';
+      const headers = {};
+
+      const result = analyzeBotProtection({
+        status: 200,
+        headers,
+        html,
+      });
+
+      expect(result.crawlable).to.be.true;
+      expect(result.type).to.equal('none');
+    });
+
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @ssilare-adobe,

## Strengths

- `bot-blocker-detect.js:106-107`: The fix directly addresses the root cause. Replacing unbounded `.*` with `\s+(and\s+)?` eliminates catastrophic matching across thousands of characters of minified HTML while preserving detection of real challenge copy. The new patterns are also free of ReDoS risk - no nested quantifiers or overlapping alternations, and worst-case backtracking is linear.
- `bot-blocker-detect.test.js:962-975`: The regression test reproduces the exact failure scenario from the bug report ("pressure" + "placeholder" proximity) with realistic HTML content and good behavior-level assertions checking both `crawlable` and `type`.
- Change discipline is excellent. Two patterns changed, one test added, nothing else touched. The PR body explains the mechanism of failure thoroughly with concrete index positions from a real site.
- Follows all documented project conventions: test mirrors src structure, uses describe/it with expect assertions, and the commit message follows conventional commits format.

## Issues

### Minor (Nice to Have)

1. `bot-blocker-detect.js:106-107`: The patterns lack word boundaries, so they can match inside compound words. For example, `/press\s+(and\s+)?hold/i` would match in content like "Express Holdings" - "press" at the tail of "Express", whitespace, then "hold" at the head of "Holdings". Adding `\b` at both edges (`/\bpress\s+(and\s+)?hold\b/i` and `/\bclick\s+(and\s+)?hold\b/i`) would close this. This is Minor because the false-positive surface is dramatically smaller than before, and the practical likelihood is low.

2. `bot-blocker-detect.test.js`: The new regression test only exercises the `press...hold` false-positive path. A symmetric negative test for `click...hold` (e.g., HTML containing "clickbait" near "household") would make future regressions on either pattern independently detectable.

## Recommendations

- The broader pattern (a flat array of regexes run against full HTML documents) has a structural fragility: any regex with unbounded wildcards is one minified page away from the same class of false positive. Worth considering in a future effort: a lightweight validation harness that runs each pattern against a corpus of known-good pages as a CI gate, catching overly broad patterns before they ship.

## Assessment

Ready to merge? Yes
Reasoning: This is a well-scoped, minimal fix to a real production false-positive scenario. The tightened patterns correctly model the intended detection semantics, CI passes with all 1155 tests green, and no security or architectural concerns exist.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 33s | <b>Cost:</b> $2.26 | <b>Commit:</b> <code>261f994f1820ac1291696b2f1f6a023dcb8c994e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1650, opened 2026-06-04
feat(utils): accept optional headers in detectBotBlocker

ymadobe:
## Related Issues
- LLMO-4975 — careinsurance.com bot-blocker probe reports false-positive block even though scraping succeeds via `scraperConfig.headers`

## Summary
Plumb an optional `headers` map into `detectBotBlocker` so callers can mirror the headers the real scraper sends. Today the probe hits the origin with only a hardcoded `User-Agent`, so any site whose WAF allowlists scraper traffic via a required header (e.g. Akamai Bot Manager checking `Accept-Language`) probes as `crawlable: false` even though the scraper itself gets through fine.

`headers` defaults to `{}`, so existing callers are unaffected. `User-Agent` is spread last in the fetch headers and always overridden with `SPACECAT_USER_AGENT` — the `scraperConfig` schema in `@adobe/spacecat-shared-data-access` already blocks reserved header names at write time; the ordering is belt-and-braces.

## Test plan
- [x] 2 new unit tests: headers passthrough + UA override defense
- [x] Full `spacecat-shared-utils` suite passes (1154 tests)
- [x] ESLint clean

### DIFF
--- a/packages/spacecat-shared-utils/src/bot-blocker-detect/bot-blocker-detect.js
+++ b/packages/spacecat-shared-utils/src/bot-blocker-detect/bot-blocker-detect.js
@@ -326,6 +326,9 @@ function analyzeError(error) {
  *
  * @param {Object} config - Configuration object
  * @param {string} config.baseUrl - The base URL to check
+ * @param {Object} [config.headers={}] - Extra request headers to send alongside the
+ *   probe (e.g. per-site `scraperConfig.headers` so the probe mirrors what the
+ *   real scraper sends). `User-Agent` is always overridden with SPACECAT_USER_AGENT.
  * @param {number} [config.timeout=5000] - Request timeout in milliseconds
  * @param {Object} [config.log=console] - Logger with warn/debug methods
  * @returns {Promise<Object>} Detection result with:
@@ -337,7 +340,12 @@ function analyzeError(error) {
  *   - confidence {number}: Confidence level (0.0-1.0, see confidence level constants)
  * @throws {Error} If baseUrl is not a valid URL
  */
-export async function detectBotBlocker({ baseUrl, timeout = DEFAULT_TIMEOUT, log = console }) {
+export async function detectBotBlocker({
+  baseUrl,
+  headers = {},
+  timeout = DEFAULT_TIMEOUT,
+  log = console,
+}) {
   if (!baseUrl || !isValidUrl(baseUrl)) {
     throw new Error('Invalid baseUrl');
   }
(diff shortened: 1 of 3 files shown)


MysticatBot (approved):
Hey @ymadobe,

## Strengths

- **Backward-compatible extension** (`bot-blocker-detect.js:342`): `headers = {}` default means every existing caller continues working unchanged. Clean way to extend a public API.
- **UA override ordering is correct** (`bot-blocker-detect.js:374`): Spreading caller headers first and `User-Agent` last ensures the scraper identity cannot be overridden. Defense-in-depth with the upstream schema validation in `spacecat-shared-data-access`.
- **TypeScript declaration kept in sync** (`index.d.ts:15`): `headers?: Record<string, string>` correctly models HTTP headers and matches the runtime default.
- **Tests cover both critical behaviors** (`bot-blocker-detect.test.js:235-270`): Header passthrough AND UA override verified independently using nock's `matchHeader` at the wire boundary - behavioral tests, not implementation-detail tests.
- **Well-scoped change**: Three files, one concern, no unrelated cleanup. Follows project conventions (conventional commit, test mirrors src structure, nock for HTTP mocking).

## Issues

### Minor (Nice to Have)

1. **`bot-blocker-detect.js:328` - JSDoc type could be more precise**: The `@param` tag declares `{Object}` for `config.headers`. The actual type is `Record<string, string>` (as the `.d.ts` correctly states). Using `{Object.<string, string>}` in the JSDoc would make the contract self-documenting for callers not using TypeScript.

   How to fix: `@param {Object.<string, string>} [config.headers={}] - ...`

2. **`bot-blocker-detect.test.js:257` - UA override test does not positively assert the expected value**: The `matchHeader` predicate confirms the sent UA is *not* `'attacker-ua'` and is truthy, but does not confirm it equals `SPACECAT_USER_AGENT`. If the spread logic were accidentally reversed, this test could still pass.

   How to fix: Match the expected constant directly:
   ```js
   .matchHeader('User-Agent', SPACECAT_USER_AGENT)
   ```

## Recommendations

- Consider a test for `headers` containing an empty-string value (e.g., `{ 'X-Empty': '' }`). Some WAFs key on header *presence* regardless of value, and this validates the spread does not strip falsy values.
- The function performs no validation on header names beyond the UA override. If a future caller bypasses the data-access schema (e.g., a CLI tool calling `detectBotBlocker` directly), security-sensitive headers (`Host`, `Authorization`, `Cookie`) would pass through. Worth tracking if the function gains callers outside the scraperConfig pipeline.

## Assessment

Ready to merge? Yes
Reasoning: Focused, backward-compatible, and well-tested. The UA override defense is sound. Minor items are cosmetic tightening opportunities, not blockers. CI passes with 1154 tests green.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 31s | <b>Cost:</b> $2.47 | <b>Commit:</b> <code>b81eac8179c320b86665c06d1621542c80fee8de</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anagarwa (approved):

PR #1655, opened 2026-06-09
feat: add `allByExternalOwnerIdAndExternalSiteId` type to SiteCollection

radhikagpt1208:
## Summary

Adds the missing TypeScript type declaration for `allByExternalOwnerIdAndExternalSiteId` on the `SiteCollection` interface in `spacecat-shared-data-access`.

The runtime method already exists — it is auto-generated from the site schema's external owner/site index (the same index that backs `findByExternalOwnerIdAndExternalSiteId`). This change only declares the corresponding type so it mirrors the existing `findBy...` declaration directly above it.

## What changed

- `src/models/site/index.d.ts` — declare `allByExternalOwnerIdAndExternalSiteId(externalOwnerId: string, externalSiteId: string): Promise<Site[]>` on `SiteCollection`.

## Impact

- **Runtime:** none — purely a type-declaration change. No behavior change.
- **Type-checking:** TS consumers can now call the method without casting, and editors provide proper autocomplete.

## Test plan

- `npm run lint` passes.
- Existing unit tests pass (no runtime change).
- Verified the singular `findByExternalOwnerIdAndExternalSiteId` is already declared and that the plural `allBy...` accessor is generated from the same index.


packages/spacecat-shared-data-access/src/models/site/index.d.ts
```diff
@@ -419,4 +419,7 @@ export interface SiteCollection extends BaseCollection<Site> {
   findByExternalOwnerIdAndExternalSiteId(
     externalOwnerId: string, externalSiteId: string
   ): Promise<Site | null>;
+  allByExternalOwnerIdAndExternalSiteId(
```
atopper, line 422: A blank line before this would help make it stand out.
If exported, shouldn't it have a javadoc explaining it?

atopper (approved):
One suggestion.

packages/spacecat-shared-data-access/src/models/site/index.d.ts
radhikagpt1208: `externalOwnerId` and `externalSiteId` are existing database columns, and this method follows the same naming/query pattern as the other `findBy*/allBy*` methods in this interface. Since it's a straightforward lookup on persisted fields, I didn't add a JSDoc because it would largely restate what is already expressed by the method name and parameter names.

PR #1659, opened 2026-06-10
fix(data-access): add retry logic for S3 DNS failures in configuration retrieval

danielbatica:
## Summary

Adds exponential backoff retry mechanism to handle transient DNS and network errors when retrieving configuration from S3.

## Problem

The SpaceCat platform has been experiencing chronic DNS resolution failures when accessing S3 configuration, causing audits to fail immediately with:

```
getaddrinfo EBUSY spacecat-prod-importer.s3.us-east-1.amazonaws.com
```

### Historical Context

- **First observed:** December 2025
- **Escalation:** May 2026 (10x increase in frequency)
- **Impact:** 3,000+ errors across 24+ days over 6 months
- **Worst incident:** May 11, 2026 - 496 audit failures in one day
- **Most recent:** June 8, 2026 - Vonage accessibility audit failure

### Root Cause

1. Configuration check happens **before any audit work begins**
2. DNS resolver gets exhausted during high-concurrency periods (scheduled audit batches, Lambda cold starts)
3. **No retry logic** - single DNS failure = immediate abort
4. Affects all services: audit-worker, api-service, import-worker

### Call Stack

```
StepAudit.run()
  → isAuditDisabledForSite()
    → isAuditEnabledForSite()
      → Configuration.findLatest()  ← DNS FAILURE HERE
        → s3Client.send(GetObjectCommand)  
          → getaddrinfo EBUSY
            → [AUDIT ABORTS]
```

## Solution

Added `retryWithBackoff()` helper function with exponential backoff:
- **Initial delay:** 100ms
- **Max retries:** 3 attempts
- **Delays:** 100ms → 200ms → 400ms
- **Total max time:** ~700ms

### Retryable Errors

- `EBUSY` - DNS resolver busy (primary issue)
- `ECONNRESET` - Connection reset
- `ETIMEDOUT` - Connection timeout
- `ENOTFOUND` - DNS not found
- `NetworkingError` - AWS SDK networking error
- Any error with `getaddrinfo` in message

### Non-Retryable Errors

- `NoSuchKey` - Configuration doesn't exist (first-time setup)
- `DataAccessError` - Application-level errors
- All other errors after 3 retries

## Changes

**File:** `packages/spacecat-shared-data-access/src/models/configuration/configuration.collection.js`

1. Added `retryWithBackoff()` helper function (lines 25-62)
2. Updated `findLatest()` to wrap S3 call with retry logic
3. Updated `findByVersion()` to wrap S3 call with retry logic
4. Added JSDoc comments documenting retry behavior

## Testing

- [x] JavaScript syntax validation passed
- [ ] Unit tests need to be added
- [ ] Integration tests with simulated DNS failures

### Expected Improvement

- **Reduces audit failures** from transient DNS issues by ~95%
- **No change to happy path** - retries only happen on errors
- **Minimal latency impact** - max 700ms total retry time (only on failures)
- **No infrastructure changes required**

### Deployment Strategy

1. Deploy to staging first
2. Monitor for 48 hours
3. Verify retry metrics and success rate
4. Deploy to production
5. Monitor high-concurrency periods (weekly audit batches)

## Monitoring

After deployment, watch for:
- Decrease in `getaddrinfo EBUSY` errors in Coralogix
- Increase in successful configuration retrievals after retries
- Warning logs: `"S3 operation failed with EBUSY, retrying in Xms"`
- No increase in overall S3 API call rate (retries should be rare)

## Future Improvements

Additional measures to consider (separate PRs):
- [ ] Add in-memory configuration cache (5-min TTL)
- [ ] Add VPC endpoint for S3 (bypass DNS entirely)
- [ ] Pre-warm configuration cache on Lambda cold start
- [ ] Add CloudWatch metrics for retry rates
- [ ] Add circuit breaker pattern

## References

- Coralogix logs: search for `getaddrinfo EBUSY spacecat-prod-importer.s3`
- Incident dates: May 11, May 12-14, June 4-5, June 8
- Related services: audit-worker, api-service, import-worker


packages/spacecat-shared-data-access/src/models/configuration/configuration.collection.js
```diff
@@ -22,6 +22,46 @@ import { checkConfiguration } from './configuration.schema.js';
...
+ * @param {Object} log - Logger instance
+ * @returns {Promise} - Result of the operation
+ */
+async function retryWithBackoff(operation, maxRetries = 3, initialDelay = 100, log) {
+  let lastError;
+
+  for (let attempt = 0; attempt <= maxRetries; attempt++) {
+    try {
+      return await operation();
+    } catch (error) {
```
MysticatBot: issue (blocking): The `log` parameter has no default value and no null guard. If `this.log` is ever undefined when these methods are called, the function throws `TypeError: Cannot read properties of undefined (reading 'warn')` on the first retry attempt - masking the original transient error with an unrelated TypeError.

This is the worst place for a null-dereference: inside an error-handling path meant to add resilience. A crash here turns a retryable DNS failure into an undiagnosable one.

**Fix:** Guard the log call:
```js
if (log) log.warn(`S3 operation failed with ${error.code || error.name}, retrying in ${delay}ms (attempt ${attempt + 1}/${maxRetries})`);
```
Or default the parameter: `log = { warn: () => {} }`

packages/spacecat-shared-data-access/src/models/configuration/configuration.collection.js
```diff
@@ -22,6 +22,46 @@ import { checkConfiguration } from './configuration.schema.js';
...
+ * @param {number} initialDelay - Initial delay in ms (default: 100)
+ * @param {Object} log - Logger instance
+ * @returns {Promise} - Result of the operation
+ */
+async function retryWithBackoff(operation, maxRetries = 3, initialDelay = 100, log) {
+  let lastError;
+
+  for (let attempt = 0; attempt <= maxRetries; attempt++) {
+    try {
+      return await operation();
```
MysticatBot: issue (blocking): Pure exponential backoff without jitter causes thundering herd when multiple Lambda invocations hit DNS exhaustion simultaneously - exactly the scenario the PR describes ("DNS resolver gets exhausted during high-concurrency periods").

All instances retry at the same wall-clock offsets (100ms, 200ms, 400ms), concentrating retries on the already-saturated resolver at the same instants. This can worsen the problem rather than mitigate it.

**Fix:** Add jitter to spread retry load:
```js
const jitter = 0.5 + Math.random() * 0.5;
const delay = Math.floor(initialDelay * Math.pow(2, attempt) * jitter);
```
Or use full jitter: `Math.floor(Math.random() * initialDelay * Math.pow(2, attempt))`

MysticatBot (changes requested):
Hey @danielbatica,

**Verdict: Request changes** - two robustness issues in the retry helper that matter most under the exact failure conditions this fix targets.
_Changes: Adds exponential backoff retry mechanism to `ConfigurationCollection` S3 operations to handle transient DNS and network errors (2 files)._

## Must fix before merge
1. [Important] Missing null guard on `log` parameter - crash during error handling masks the original transient error - `configuration.collection.js:40` (details inline)
2. [Important] Deterministic backoff without jitter causes thundering herd under correlated failures - `configuration.collection.js:39` (details inline)

<details>

- nit: `throw lastError` after the loop (line 46) is dead code - the catch block already re-throws when `attempt === maxRetries`, so the only path to line 46 is a negative `maxRetries` - `configuration.collection.js:46`
- suggestion: Investigate whether the AWS SDK v3 built-in retry (`@aws-sdk/middleware-retry`) already retries these DNS errors or if the two layers compound (up to 16 total attempts in worst case) - `configuration.collection.js:33`
- suggestion: Consider extracting `retryWithBackoff` to a shared utility so other S3 consumers (audit-worker, api-service, import-worker) can inherit the same protection without duplicating the pattern - out of scope for this PR but worth tracking given the PR description notes the problem affects all services.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 34s | <b>Cost:</b> $2.82 | <b>Commit:</b> <code>8886075738796e097a6578640b3a63e4f34b0626</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ekremney (approved):
Thanks for tackling this — the `EBUSY` failures have been a chronic source of pain and it's good to see them being addressed.

One thought before the inline comments: when I saw the branch name `fix/s3-dns-retry-and-cache` I assumed this PR would also introduce a small in-memory cache for the configuration. A 1-minute TTL cache would arguably do more for the underlying issue than the retries themselves — `findLatest()` is called on essentially every audit invocation, but the config barely changes between deploys, so 99%+ of those S3 calls (and therefore 99%+ of the DNS lookups causing the exhaustion) are avoidable. Would be great to see that follow up soon, even if separately.

A few smaller things on the retry implementation itself:

**1. Add jitter to the backoff.** The delays are currently deterministic: 100ms → 200ms → 400ms. The scenario the PR describes is many Lambdas failing concurrently against an exhausted DNS resolver — but without jitter, every one of those Lambdas will retry at exactly the same offsets from their original failure. That's a textbook thundering herd: the retry storm hits the resolver in three synchronized spikes, which is the same pattern that caused the original exhaustion. A small random factor (e.g. `±20%`, as in `ims-base-client.js`) spreads the retries out and lets the resolver recover between waves.

**2. Parameter ordering of `retryWithBackoff`.** `function retryWithBackoff(operation, maxRetries = 3, initialDelay = 100, log)` has `log` positionally last but without a default, while the two params before it do have defaults. It works, but reads awkwardly and forces callers to pass the defaults explicitly to reach `log` (which is what the call sites do today). An options object would read better and make future tuning (custom `isRetryable` predicates, jitter config, etc.) drop-in:

```js
retryWithBackoff(operation, log, { maxRetries, initialDelay });
```

**3. Unreachable `throw lastError;` at the end of the function.** The loop runs `maxRetries + 1` times, and on the final iteration the `attempt === maxRetries` branch always throws. The post-loop `throw lastError;` can never execute — safe to drop, or restructure the loop so it isn't needed.

Thanks again!

packages/spacecat-shared-data-access/test/unit/models/configuration/configuration.collection.test.js
```diff
@@ -214,4 +214,269 @@ describe('ConfigurationCollection', () => {
...
+            VersionId: 'retry-success-version',
+          });
+
+        const result = await instance.findLatest();
+
+        expect(result).to.be.an('object');
+        expect(result.getId()).to.equal('retry-success-version');
+        expect(mockS3Client.send).to.have.been.calledTwice;
+        expect(mockLogger.warn).to.have.been.calledWith('S3 operation failed with EBUSY, retrying in 100ms (attempt 1/3)');
+      });
```
MysticatBot: issue (blocking): The assertion `calledWith("...retrying in 100ms...")` expects an exact delay, but the production code applies +-20% jitter via `Math.random()`. The actual logged delay for attempt 0 ranges from 80ms to 120ms (`Math.floor(100 * (0.8 + Math.random() * 0.4))`), so this assertion passes only when `Math.random()` lands exactly at 0.5.

The same issue affects the ECONNRESET test (line 179: "retrying in 100ms", "retrying in 200ms") and any other test asserting exact delay values.

The jitter timing test at line 375 correctly uses `within(80, 120)` ranges - the earlier tests need the same treatment.

**Fix (pick one):**
1. Stub `Math.random` to return a fixed value so delays are deterministic:
```js
sinon.stub(Math, 'random').returns(0.5); // jitter = 1.0, delays are exact base values
```
2. Use regex matching for the log assertion:
```js
expect(mockLogger.warn).to.have.been.calledWith(
  sinon.match(/S3 operation failed with EBUSY, retrying in \d+ms/)
);
```
Option 1 is cleaner since it makes all delay assertions deterministic without changing the assertion style.

MysticatBot (changes requested):
Hey @danielbatica,

**Verdict: Request changes** - test assertions use hardcoded delay values that conflict with the jitter implementation, causing flaky CI failures.
_Changes: Adds exponential backoff retry with jitter for transient DNS/network errors in ConfigurationCollection S3 operations (2 files)._

## Must fix before merge
1. [Critical] Test log assertions use exact delay values ("retrying in 100ms") but production jitter makes actual delays range 80-120ms - will flake in CI - `configuration.collection.test.js:159` (details inline)

<details>

- nit: warn log omits `error.message`, making the log unhelpful when `error.code` is undefined and `error.name` is generic "Error" - `configuration.collection.js:43`
- nit: "throws after max retries" asserts on the outer catch wrapper message rather than verifying the original error code/cause is preserved - `configuration.collection.test.js:256`
</details>

- Null guard on log parameter added (`if (log)` check)
- Jitter added to prevent thundering herd (+-20%)
- Parameter API refactored to options object pattern
- Unreachable `throw lastError` dead code removed

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 26s | <b>Cost:</b> $3.53 | <b>Commit:</b> <code>dbff77e15d5aef426d9f23ad49ef60f7c155293f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @danielbatica,

**Verdict: Comment** - all prior blocking findings are resolved; only minor polish items remain.
_Changes: Adds exponential backoff retry with jitter for transient DNS/network errors in ConfigurationCollection S3 operations (2 files)._

<details>

- nit: warn log shows `error.code || error.name` but not the message - when matching via the `getaddrinfo` message check with no distinctive code, the log reads "S3 operation failed with Error" which is unhelpful for correlation in Coralogix. Consider `error.code || error.name || error.message?.slice(0, 50)` - `configuration.collection.js:56`
- nit: jitter test manually restores the shared `randomStub` at line 459 and re-creates it at line 482; if the test fails between restore and re-stub, the shared `afterEach` operates on a stale reference. A dedicated `describe` block without the shared `beforeEach` stub would decouple the lifecycle - `configuration.collection.test.js:459`
</details>

- Null guard on log parameter added (`if (log)` check)
- Jitter (+-20%) added to prevent thundering herd
- Parameter API refactored to options object pattern
- Unreachable `throw lastError` dead code removed
- Test assertions now use `Math.random` stub for deterministic delays

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 53s | <b>Cost:</b> $3.35 | <b>Commit:</b> <code>bc4904bb3546a64f51c23b58754faa6eaa2678e0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-data-access/src/service/index.js
```diff
@@ -78,6 +79,34 @@ const createPostgrestService = (config, client = undefined) => {
...
+ * The AWS SDK's default retry strategy handles ECONNRESET, ETIMEDOUT, ENOTFOUND, etc.,
+ * but does NOT retry EBUSY DNS errors. This class adds EBUSY to the retryable set.
+ */
+class EbusyRetryStrategy extends StandardRetryStrategy {
+  constructor(maxAttempts = 4) {
+    super(maxAttempts);
+  }
+
+  async retry(next, args) {
+    const { error } = args.response || {};
```
MysticatBot: issue (blocking): `StandardRetryStrategy` in `@smithy/util-retry` v3 implements the `RetryStrategyV2` interface, which exposes `acquireInitialRetryToken`, `refreshRetryTokenForRetry`, and `recordSuccess`. It does NOT have a `retry(next, args)` method. The SDK's retry middleware detects the V2 interface and runs its own retry loop internally - it never calls `retry()` on the strategy.

This means the EBUSY classification logic here is dead code. `super.retry(next, args)` would also throw at runtime since the method does not exist on the superclass.

**How to fix (two options):**

1. Override `refreshRetryTokenForRetry(token, errorInfo)` and reclassify EBUSY errors before delegating to super:
```js
async refreshRetryTokenForRetry(token, errorInfo) {
  const { error } = errorInfo;
  const isEbusy = error?.code === 'EBUSY'
    || (error?.message?.includes('getaddrinfo') && error?.message?.includes('EBUSY'));
  if (isEbusy) {
    return super.refreshRetryTokenForRetry(token, { ...errorInfo, errorType: 'TRANSIENT' });
  }
  return super.refreshRetryTokenForRetry(token, errorInfo);
}
```

2. Use a middleware plugin placed before the retry middleware that sets `error.$retryable = {}` on EBUSY errors (since `isTransientError` already checks `isRetryableByTrait`).

packages/spacecat-shared-data-access/package.json
```diff
@@ -44,6 +44,7 @@
     "@adobe/spacecat-shared-utils": "1.105.0",
     "@supabase/postgrest-js": "2.106.2",
     "@aws-sdk/client-s3": "^3.940.0",
+    "@smithy/util-retry": "^3.0.0",
```
MysticatBot: issue (blocking): `@smithy/util-retry` was added here but `package-lock.json` was not regenerated. CI runs `npm ci` which requires lock file parity. The CI logs show:

```
npm error Missing: @smithy/util-retry@3.0.11 from lock file
```

Run `npm install` at the repo root and commit the updated `package-lock.json`.

packages/spacecat-shared-data-access/src/service/index.js
```diff
@@ -78,6 +79,34 @@ const createPostgrestService = (config, client = undefined) => {
   });
 };
 
+/**
+ * Custom retry strategy that extends StandardRetryStrategy to include EBUSY errors.
+ * The AWS SDK's default retry strategy handles ECONNRESET, ETIMEDOUT, ENOTFOUND, etc.,
+ * but does NOT retry EBUSY DNS errors. This class adds EBUSY to the retryable set.
+ */
+class EbusyRetryStrategy extends StandardRetryStrategy {
```
MysticatBot, line 91: issue (blocking): `EbusyRetryStrategy` is 28 lines of branching production code with zero test coverage. The repo enforces 100% lines/statements and 97% branches per package (documented in CLAUDE.md and each package's `.nycrc.json`). Even after the lockfile fix, coverage will fail here.

Once the API contract is corrected (see the Critical finding on line 93), add tests that verify:
- EBUSY errors (by code and by message pattern) are classified as retryable
- Non-EBUSY errors pass through unchanged
- The strategy delegates to the parent correctly

Note: the class is not currently exported, so you'll need to either export it for direct testing or test it indirectly via integration with a mocked S3 transport.

MysticatBot (changes requested):
Hey @danielbatica,

**Verdict: Request changes** - the retry strategy overrides a method that the SDK never calls, so EBUSY errors are still not retried.
_Changes: Replaces application-level retryWithBackoff with SDK-level EbusyRetryStrategy extending StandardRetryStrategy for transient DNS error handling (4 files)._

Note: CI checks are currently failing - resolve before merge.

## Must fix before merge
1. [Critical] `EbusyRetryStrategy.retry(next, args)` overrides a non-existent method on StandardRetryStrategy v3 - EBUSY retry logic is dead code - `src/service/index.js:93` (details inline)
2. [Important] `@smithy/util-retry` added to package.json but package-lock.json not regenerated - CI fails with "Missing from lock file" - `package.json:47` (details inline)
3. [Important] `EbusyRetryStrategy` has 28 lines of branching production code with zero test coverage - repo enforces 100% lines/statements - `src/service/index.js:87` (details inline)

<details>

- nit: `maxAttempts: 4` in S3Client options is ignored when a custom `retryStrategy` is provided - remove to avoid confusion about which controls attempt count - `src/service/index.js:125`
</details>

- Null guard on log parameter (app-level retry removed entirely)
- Jitter added to prevent thundering herd (now handled by SDK built-in backoff)
- Parameter API refactored to options object pattern (no longer applicable)
- Unreachable `throw lastError` dead code removed (function deleted)
- Test assertions use deterministic Math.random stub (retry tests removed)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 46s | <b>Cost:</b> $4.45 | <b>Commit:</b> <code>d5f563439be9142d8b5dce6c83a94790f2f31657</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (changes requested):
Hey @danielbatica,

**Verdict: Request changes** - the core retry mechanism still overrides a non-existent SDK method, so EBUSY errors remain unretried in production.
_Changes: Adds test coverage for EbusyRetryStrategy and S3 error handling in ConfigurationCollection (5 files)._

Note: CI checks are currently failing - resolve before merge.

## Must fix before merge
1. [Critical] `EbusyRetryStrategy.retry(next, args)` overrides a method that does not exist on `StandardRetryStrategy` v3 - entire EBUSY retry mechanism is dead code in production - `src/service/index.js:93` (details inline)
2. [Important] `package-lock.json` not regenerated after adding `@smithy/util-retry` - CI cannot pass - `package.json:47` (details inline)
3. [Important] Tests call `strategy.retry(mockNext, mockArgs)` directly but the SDK never invokes this method - tests validate a dead code path and give false confidence - `test/unit/service/index.test.js:236` (details inline)

<details>

- nit: `maxAttempts: 4` in S3Client options is ignored when a custom `retryStrategy` is supplied (the strategy constructor already receives `4`) - remove to avoid confusion - `src/service/index.js:125`
- nit: `import { S3Client } from '@aws-sdk/client-s3'` added to test file but never referenced in assertions - unused import - `test/unit/service/index.test.js:15`
- suggestion: "S3 service with EbusyRetryStrategy" tests (lines 173-204) assert only that `dataAccess.Configuration` exists, not that the retry strategy is configured correctly - after fixing the production code, assert on `s3Client.config.retryStrategy` directly
</details>

- Zero test coverage for EbusyRetryStrategy (tests added in this commit, though they will need rewriting once the production API contract is corrected)
- Null guard on log parameter (app-level retry removed in earlier commit)
- Jitter added to prevent thundering herd (now handled by SDK built-in backoff)
- Unreachable `throw lastError` dead code removed

packages/spacecat-shared-data-access/src/service/index.js
MysticatBot: issue (blocking): `StandardRetryStrategy` in `@smithy/util-retry` v3 implements the `RetryStrategyV2` interface. Its public API is `acquireInitialRetryToken`, `refreshRetryTokenForRetry`, and `recordSuccess`. There is no `retry(next, args)` method. The SDK retry middleware never calls this method - it uses the token-based V2 interface internally.

The tests pass because they call `strategy.retry(mockNext, mockArgs)` directly, but in production the SDK never invokes this code path. EBUSY errors will continue failing without retry.

This was flagged in the prior review and confirmed by @ekremney's analysis of the SDK internals. It remains unresolved.

**Fix:** Override `refreshRetryTokenForRetry(token, errorInfo)` instead:

```js
async refreshRetryTokenForRetry(token, errorInfo) {
  const { error } = errorInfo;
  const isEbusy = error?.code === 'EBUSY'
    || (error?.message?.includes('getaddrinfo') && error?.message?.includes('EBUSY'));
  if (isEbusy) {
    return super.refreshRetryTokenForRetry(token, { ...errorInfo, errorType: 'TRANSIENT' });
  }
  return super.refreshRetryTokenForRetry(token, errorInfo);
}
```

packages/spacecat-shared-data-access/package.json
MysticatBot: issue (blocking): `@smithy/util-retry` was added here but `package-lock.json` was not regenerated. `npm ci` requires lock file parity - CI fails with "Missing from lock file".

**Fix:** Run `npm install` at the monorepo root and commit the updated `package-lock.json`.

packages/spacecat-shared-data-access/package.json
```diff
@@ -44,6 +44,7 @@
     "@adobe/spacecat-shared-utils": "1.105.0",
     "@supabase/postgrest-js": "2.106.2",
     "@aws-sdk/client-s3": "^3.940.0",
+    "@smithy/util-retry": "^4.0.0",
```
MysticatBot, line 47: issue (blocking): `@smithy/util-retry` ^4.0.0 added here but `package-lock.json` was not regenerated. `npm ci` requires lockfile parity and CI fails with "Missing from lock file".

This is the sole remaining blocker. Run `npm install` at the monorepo root and commit the updated `package-lock.json`.

MysticatBot (changes requested):
Hey @danielbatica,

**Verdict: Request changes** - lockfile must be committed for CI to pass; the production code itself is correct.
_Changes: Adds SDK-level EBUSY retry strategy for S3 DNS failures by extending StandardRetryStrategy's refreshRetryTokenForRetry (5 files)._

Note: CI checks are currently failing - resolve before merge.

## Must fix before merge
1. [Important] Missing `package-lock.json` update after adding `@smithy/util-retry` - CI fails with lockfile parity error - `packages/spacecat-shared-data-access/package.json:47` (details inline)

<details>

- nit: `maxAttempts: 4` in S3Client options is ignored when a custom `retryStrategy` is provided (the strategy constructor already controls attempt count) - remove to avoid a false affordance for future maintainers - `src/service/index.js:128`
- nit: "S3 service with EbusyRetryStrategy" tests (lines 169-199) only assert `dataAccess.Configuration` exists - they would pass identically if the retry strategy were removed; the test named "configures S3Client with maxAttempts: 4" asserts nothing about maxAttempts - `test/unit/service/index.test.js:181`
</details>

- `retry()` dead code replaced with correct `refreshRetryTokenForRetry` override
- Zero test coverage addressed with comprehensive unit tests
- Tests now exercise the correct SDK V2 interface
- Null guard, jitter, and options-object API from app-level retry iteration removed (no longer applicable after SDK-level approach)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 17s | <b>Cost:</b> $3.28 | <b>Commit:</b> <code>b367b73ced6e3554828a4b320fcbc74c2990999b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @danielbatica,

**Verdict: Comment** - production code is correct; only minor polish items remain.
_Changes: Adds SDK-level EbusyRetryStrategy extending StandardRetryStrategy to retry EBUSY DNS errors on S3 operations (6 files)._

Note: CI checks are currently failing (integration test DB seeding timeout/duplicate-key issues, unrelated to this PR) - resolve before merge.

<details>

- nit: `maxAttempts: 4` in S3Client options is ignored when a custom `retryStrategy` is provided - the strategy constructor already controls attempt count; remove to avoid a false affordance for future maintainers - `src/service/index.js:128`
- nit: "S3 service with EbusyRetryStrategy" tests only assert `dataAccess.Configuration` exists, which passes even without the retry strategy - assert `s3Client.config.retryStrategy instanceof EbusyRetryStrategy` for regression protection - `test/unit/service/index.test.js:69`
- nit: `superRefreshStub` targets `Object.getPrototypeOf(Object.getPrototypeOf(strategy))` which is fragile across SDK upgrades - consider asserting the stub target IS `StandardRetryStrategy.prototype` as a test precondition - `test/unit/service/index.test.js:121`
- suggestion: Export `EbusyRetryStrategy` so other S3 consumers (audit-worker, api-service, import-worker per the PR description) can reuse it without duplicating the pattern - `src/service/index.js:94`
</details>

- `retry()` dead code replaced with correct `refreshRetryTokenForRetry` override
- Zero test coverage addressed with comprehensive unit tests
- Tests now exercise correct SDK V2 interface
- Package-lock.json committed
- Null guard, jitter, and options-object API from earlier app-level retry iteration removed (no longer applicable after SDK-level approach)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 3s | <b>Cost:</b> $4.47 | <b>Commit:</b> <code>f0c30455ebdda73c74efe33f014e2888762ec773</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1663, opened 2026-06-11
feat(athena-client): surface structured AthenaError on query failure

akshaymagapu:
## What

When an Athena query reaches `FAILED`/`CANCELLED`, `#pollToCompletion` now attaches Athena's **structured error** to the thrown `Error`:

- `error.athenaError` — the full `AthenaError` object (`ErrorCategory`, `ErrorType`, `Retryable`)
- `error.retryable` — its `Retryable` boolean

…when Athena returns an `AthenaError`. Adds an exported `AthenaQueryError` type.

## Why

Callers (e.g. the `cdn-logs-analysis` audit retry logic) currently classify transient-vs-permanent Athena failures by **matching the `StateChangeReason` string**. Athena already exposes an authoritative `Retryable` boolean on `QueryExecution.Status.AthenaError` — keying off it is more robust than substring matching.

## Non-breaking

Additive only: the error **message is unchanged**, and the new properties are present **only** when Athena returns an `AthenaError`. Existing consumers that read `e.message` are unaffected.

```js
// FAILED/CANCELLED branch
const error = new Error(StateChangeReason || `Query ${State}`);
const { AthenaError: athenaError } = status;
if (athenaError) {
  error.athenaError = athenaError;
  error.retryable = athenaError.Retryable === true;
}
throw error;
```

## Tests

3 new cases (retryable=true attached, retryable=false, and no AthenaError → properties unset). Package remains at **100%** coverage.

## Rollout

Consumer side (`spacecat-audit-worker` `cdn-logs-analysis`) already prefers `error.retryable` when present and falls back to string matching otherwise — so it self-upgrades once this releases and the dependency is bumped, with no regression in the gap.

### DIFF
--- a/packages/spacecat-shared-athena-client/src/index.d.ts
+++ b/packages/spacecat-shared-athena-client/src/index.d.ts
@@ -10,7 +10,17 @@
  * governing permissions and limitations under the License.
  */
 
-import { AthenaClient } from '@aws-sdk/client-athena';
+import { AthenaClient, AthenaError } from '@aws-sdk/client-athena';
+
+/**
+ * Error thrown when an Athena query reaches a FAILED/CANCELLED state. Extends Error
+ * with Athena's structured error when available, so callers can branch on the
+ * authoritative `retryable` flag rather than matching the message string.
+ */
+export interface AthenaQueryError extends Error {
+  athenaError?: AthenaError;
+  retryable?: boolean;
+}
 
 export interface AthenaClientOptions {
   backoffMs?: number;
--- a/packages/spacecat-shared-athena-client/src/index.js
+++ b/packages/spacecat-shared-athena-client/src/index.js
@@ -136,7 +136,16 @@ export class AWSAthenaClient {
         return;
       }
       if (State === QueryExecutionState.FAILED || State === QueryExecutionState.CANCELLED) {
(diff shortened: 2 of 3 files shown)


MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - clean, well-scoped additive change with solid test coverage.
_Changes: Surfaces Athena's structured `AthenaError` (including the authoritative `Retryable` flag) on query failure so callers can branch on it instead of matching the message string (3 files)._

No issues found. The implementation is straightforward and backward-compatible - existing catch handlers keying off `e.message` are unaffected, and the new properties are conditionally present only when Athena returns structured error metadata. The `=== true` coercion on `Retryable` is a good defensive choice against unexpected truthy values from the SDK. Tests cover the three meaningful branches (retryable true, retryable false, no AthenaError present) with clear assertions on observable state.

Note: CI "Test" check was still pending at review time.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 38s | <b>Cost:</b> $2.17 | <b>Commit:</b> <code>20e61aba1418ee272a2fc37bd5af626d410c02fe</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1660, opened 2026-06-10
feat(project-engine-client): foundation — vendor spec + generation pipeline (LLMO-5461)

aliciadriani:
## What

Foundation slice for the new `@adobe/spacecat-shared-project-engine-client` package — the typed integration with the Semrush **Project Engine API** (`/enterprise/projects/api`).

Part of **LLMO-5459** (epic) / **LLMO-5461** (client + generated types). This PR is **foundation only**: vendor the spec and wire the conversion-and-type-generation pipeline. The client wrapper, IMS handler move, and stateful mock land in follow-up PRs.

## Scope of this PR

- **Vendored spec** — `spec/projectengine_swagger_public.yaml` (Swagger 2.0, `basePath: /enterprise/projects/api`). Manual-refresh only; Semrush provides the file, no endpoint access. Not patched/edited.
- **Generation pipeline** (npm scripts):
  - `spec:convert` — swagger2openapi `--patch` v2→3.x → `build/openapi3.json` (intermediate, gitignored). Exists **only** to feed the type generators.
  - `generate:ts` — openapi-typescript → `src/generated/types.ts`
  - `generate:pydantic` — datamodel-code-generator → `python/serenity_project_engine/` (Pydantic v2)
  - `mock` — Counterfact reads the **raw v2** spec directly (`:4010`), never the converted artifact.
- **Conventions** — follows `spacecat-shared`: JS + ESM, JSDoc-typed source, mocha + chai + c8, `@adobe/eslint-config-helix`. The scaffold's TS surface is ported to JS+JSDoc; `types.ts` is a type artifact only.
- **Smoke test** — `test/foundation.test.js` asserts the vendored spec invariants (Swagger 2.0, basePath, `Auth-Data-Jwt` header) and that generated `types.ts` carries the expected `paths` interface + known v1/v2 routes.

## Commit structure

1. `feat(...)` — vendored spec + generation pipeline + config + smoke test
2. `chore(...) [skip ci]` — generated TS + Pydantic types, committed separately and marked `linguist-generated` (collapsed in review, excluded from diff counts)

## Validation

- `npm run generate` runs end-to-end (build/openapi3.json → types.ts → python package)
- `npm test` — 6 smoke tests pass
- `npm run lint` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1665, opened 2026-06-11
feat(project-engine-client): stateful mock core — store, resource ops, seeds (LLMO-5460)

aliciadriani:
## What
The stateful Project Engine Counterfact mock — LLMO-5460 (epic LLMO-5459). A generic in-memory store + the confirmed stateful resource ops, the live Counterfact runner that wires them into per-path handlers, named seeds, typed mock factories, AI-unit quota metering, bearer-auth enforcement, a test-only control surface, and an in-package E2E that drives the real client against a booted mock.

Now based directly on `main` (the earlier stack #1660 → #1661 has merged).

## Spike outcome (confirmed against both consumers)

Grepped the real consumers to set the AC floor (`docs/mock-statefulness.md`):
- `spacecat-api-service` is the only Project Engine consumer; `llmo-data-retrieval-service` makes zero calls.
- The write-then-read spine is **projects**, **ai_models** (per project), **prompts** (aio, per project). The prompt slice was reconciled against the consumer's actual call sites (`rest-transport.js`) and the spec: create is `POST .../aio/prompts/tagged`, list is `POST .../aio/prompts/by_tags`, delete is `DELETE .../aio/prompts`.

## Live-fidelity validation (against the real Semrush Project Engine API)

Beyond the consumer grep, every metered write + read was replayed against the **real** API (prod test workspace) and the mock corrected to match — so a consumer that passes against the mock passes against live:
- Confirmed live shapes/statuses: `createProject` **201** (nests request fields under `settings.ai`, marks draft), `createTaggedPrompts` 201 `{ ids, existing_count }`, `publish` empty **202**, `createBenchmarks`/`createBrandUrls` 200 `{ ids, existing_count }`, `deletePromptsByIds` 204, `deleteBenchmarks`/`deleteBrandUrls` empty **202**, `updateCiCompetitors` 200, `addAiModel` resolves catalog `name`+`icon`.
- Metered **405** = "payment required" (AI-unit over-allocation), not route-gating — modelled by the quota meter.
- Eventual consistency observed (async main-brand benchmark; create-then-read can be empty) — documented, not faked with timers.
- Live-only fields folded into the overlay: **CR9** (`settings.ai.primary_url`) and **CR10** (`AIOBenchmarkWithCounters.primary_url`/`root_domain`). These regenerate `src/generated/types.ts` + the python model, so this PR **does ship a (mock-fidelity) type change**.

### Mock library (`mock/`)
- **`store.js`** — generic `InMemoryStore`: collection-keyed CRUD + seed/reset, deep-clone on every read/write (a loaded seed is never mutated; `reset()` restores a pristine copy). Throws on explicit id collision.
- **`stateful.js`** — the confirmed stateful set as pure ops over the store (per-workspace/per-project scoping). No Counterfact/HTTP coupling, so unit-tested directly.
- **`factories.js`** — the typed mock-factory layer, each `(Partial<T>) => T` typed against the generated schemas — the single tsc-checked source of truth for every entity shape the mock emits: `createProjectMock`, `createProjectResponseFromRequest` (the create-response shape), `applyProjectUpdate` (the PATCH-response shape), `createProjectAiModelMock`, `createAiModelMock`, `createPromptMock`, `createBenchmarkMock`, `createBrandUrlMock`, `createLanguageMock`, `createTagNodeMock`, `createBrandTopicMock`, `createCiCompetitorMock`, `createBasicResponseMock`, `createInitStatusMock`.
- **`seeds.js`** — named seed sets (`empty-workspace`, `workspace-with-data`), `DEFAULT_SEED`, `SEED_IDS`, all built through the factories.
- **`quota.js`** — AI-unit quota meter (the disguised-405 the live API returns on over-allocation), enforced on project create / prompt write / publish; configurable via `buildSeed({ quota })` or the `POST /__quota` control route.
- **`auth.js`** — bearer-auth model: every real route requires `Authorization: Bearer <token>` (presence, not validity) or returns `401 { detail: 'Not authenticated' }`, exactly like the live gateway; the `__*` control routes are exempt.
- **`context.js`** — the Counterfact per-path Context: wires the pure store + ops + factories + quota + auth to a seed selected at startup and exposes `reset()`.
- **`run.js`** — the live runner. Materializes the committed handlers into a gitignored `.counterfact/` tree (as `.ts`), injecting the bearer-auth guard onto every handler at the single materialization seam (`injectAuthGuard`, so no handler can forget it), and launches Counterfact with `--serve` against the corrected OAS3 artifact (`build/openapi3.json`) under `--prefix /enterprise/projects/api`. `npm run generate` is required before first boot. `npm run mock`.

### Counterfact route handlers (`mock/counterfact/routes/`)
- **Stateful spine:** `v1/workspaces/{id}/projects` (GET/POST), `.../{project_id}` (GET/PATCH/DELETE), `.../ai_models` (GET/POST), `v2/.../aio/prompts/tagged` (POST create), `.../by_tags` (POST list), `.../aio/prompts` (DELETE).
- **Project surface:** v1/v2 `ai_models` + `ai_models/benchmarks`, `aio/benchmarks/{benchmark_id}/brand_urls`, `aio/tags`, `brand-topics`, `ci/competitors`, `publish`, `aio/init_status`.
- **Global catalogs:** `v1/ai_models` and `v1/languages` return the **full live taxonomy** (11 global models, 38 languages — real ids/keys/icons/codes, captured verbatim) mapped through the factories.
- **Control routes (`__*`, auth-exempt):** `__reset` (seed restore), `__seed` (load a named seed), `__quota` (set quota), `__dump` (inspect store state).

### Tests, docs, CI
- **`test/mock/*.test.js`** — unit tests for store, stateful ops, factories, seeds, context, quota, and auth (100% coverage; `counterfact/routes/**` + `run.js` excluded, validated by E2E).
- **`test/e2e/project-engine-mock.e2e.js`** — boots the live mock and drives it through the real client; gated behind `MOCK_E2E=1`, outside the default `npm test` glob. Exercises the prompt write-then-read spine (`tagged → by_tags`), the quota 405, the bearer-auth 401, and the live-corrected shapes (create `settings.ai`, benchmark re-derive), with response validation on.
- **`test/types/*.type-test.ts`** — compile-only guards: the public type surface stays assignable to `Client<paths>`, and the factories stay schema-faithful (drift + wrong-field-type are `@ts-expect-error` canaries).
- **`docs/mock-usage.md`** — usage manual for humans + agents (quick start, bearer auth, full endpoint inventory, seeds, control routes, quota, driving from tests, internals, **§9 extend-the-mock recipe**, **§10 capturing real API responses to build correct mocks**, troubleshooting).
- **`spec/overlays/corrections.yaml`** — spec corrections CR1–CR10 (the vendored swagger is never edited); `build/openapi3.json` + `src/generated/types.ts` + the python model are regenerated from it.
- **CI** — a path-gated `E2E (project-engine mock)` job (type-surface check + live E2E), scoped via a merge-base diff so it only runs when this package changes.

## Validation
- `npm test` — 170 passing, 100% statements / branches / functions / lines
- `npm run lint` — clean
- `npm run test:types` — clean
- `npm run test:e2e` — 34 passing against the live mock (response validation on)

## Follow-up (separate PR)
Adopting the mock into `spacecat-api-service`'s own E2E/local-dev (running its Project Engine calls against this mock) lands with the adoption PR — out of scope here, which delivers the mock and its self-contained E2E. The pre-existing consumer bug `getInitStatus` hits the v1 `aio/init_status` path (404s live; v2 is correct) and is tracked for that adoption PR. The opt-in "unprocessed benchmark → `listBrandUrls` 404" deterministic control also lands there, alongside the consumer E2E that needs it.

🤖 Generated with Claude Code


packages/spacecat-shared-project-engine-client/src/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/prompts.js
```diff
@@ -0,0 +1,40 @@
...
+
+/** POST — create prompts (body: { items: [text], tag_ids }). Returns the first as StringIDName. */
+export function POST($) {
+  const { path, body, context } = $;
+  const created = context.ops.prompts.createMany(
+    { workspaceId: path.id, projectId: path.project_id },
+    (body?.items ?? []).map((text) => ({ name: text })),
+  );
+  const first = created[0] ?? {};
+  return $.response[200].json({ id: first.id, name: first.name });
```
aliciadriani: **Should Fix — this create handler is mounted on a path the spec and the consumer don't use.**

The spec defines only `delete` on `/aio/prompts` (`aio-delete-prompt-by-ids-v2`) — there is no `POST` here. The real create is `POST /aio/prompts/tagged` (`aio-create-prompts-with-tags`) and the real read is `POST /aio/prompts/by_tags` (`aio-list-prompts-by-tag-ids`) — exactly what `docs/mock-statefulness.md` lists as the consumer's stateful prompt ops.

Consequences as written:
- The E2E `creates aio prompts (v2)` case drives a non-spec `POST`, so there's no `200` response schema for Counterfact to validate the `{ id, name }` envelope against — the "response validation stays on" guarantee doesn't cover this handler.
- When `spacecat-api-service` runs against the mock, prompt **create** (`tagged`) and **read** (`by_tags`) hit Counterfact's stateless stubs, so the documented prompt write-then-read spine isn't actually stateful.

**Fix:** keep `DELETE` in this file, move create into a new `aio/prompts/tagged.js`, and add a `aio/prompts/by_tags.js` read handler (the store + ops already support both via `createMany` / `list`). Repoint the E2E case at `/v2/.../aio/prompts/tagged`.

```js
// aio/prompts/tagged.js — POST create (operationId aio-create-prompts-with-tags)
export function POST($) {
  const { path, body, context } = $;
  const created = context.ops.prompts.createMany(
    { workspaceId: path.id, projectId: path.project_id },
    (body?.items ?? []).map((text) => ({ name: text })),
  );
  const first = created[0] ?? {};
  return $.response[200].json({ id: first.id, name: first.name });
}
```
```js
// aio/prompts/by_tags.js — POST list/read (operationId aio-list-prompts-by-tag-ids)
// shape the envelope to match the spec's 200 response for this op
export function POST($) {
  const { path, context } = $;
  const items = context.ops.prompts.list({ workspaceId: path.id, projectId: path.project_id });
  return $.response[200].json({ items, page: 1, total: items.length });
}
```

If this is intentionally deferred to the adoption PR, please scope it out in the description and drop the `POST /aio/prompts` row from the README's stateful-endpoints table, which currently advertises it as "create prompts."

packages/spacecat-shared-project-engine-client/src/mock/seeds.js
```diff
@@ -0,0 +1,60 @@
...
+ */
+export const WORKSPACE_WITH_DATA = Object.freeze({
+  [collectionKey('projects', { workspaceId: WORKSPACE_ID })]: [
+    { id: PROJECT_ID, name: 'Seeded Project', workspace_id: WORKSPACE_ID },
+  ],
+  [collectionKey('ai_models', { workspaceId: WORKSPACE_ID, projectId: PROJECT_ID })]: [
+    { id: 'model-gpt4o', name: 'gpt-4o' },
+  ],
+  [collectionKey('prompts', { workspaceId: WORKSPACE_ID, projectId: PROJECT_ID })]: [
+    { id: 'prompt-1', text: 'What is the best running shoe?' },
```
aliciadriani: **Nit — seed/created prompt field mismatch.** Seed prompts use `text`, but `prompts.createMany` stores `{ name }` and the create handler returns `{ id, name }`. If a `by_tags` read handler is added (see the prompts.js comment), seeded vs. created prompts will have inconsistent shapes. Align the seed to `name`:

```suggestion
    { id: 'prompt-1', name: 'What is the best running shoe?' },
```

packages/spacecat-shared-project-engine-client/mock/store.js
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * @param {Partial<Entity> & Record<string, unknown>} entity
+   * @param {string} name
+   * @returns {Entity} the stored entity (cloned)
+   */
+  create(name, entity) {
+    const id = entity.id ?? InMemoryStore.#generateId();
+    /** @type {Entity} */
+    const stored = { ...entity, id };
+    this.#collection(name).set(id, stored);
```
aliciadriani, line 84: **Nit (optional, no change required).** `create` with an explicit existing `id` silently overwrites via `Map.set` (a real API would `409`). Fine for a mock; flagging only because seeds use fixed ids, so a seed-id collision would be swallowed.

_If_ you want collisions to surface during dev, the minimal guard is:
```js
if (entity.id && this.#collection(name).has(entity.id)) {
  throw new Error(`duplicate id ${entity.id} in collection ${name}`);
}
```
Otherwise leave as-is — overwrite-on-duplicate is acceptable mock behaviour.

packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v1/workspaces/{id}/projects/{project_id}/ai_models.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/** POST — add an AI model to the project (body: { model_id }). */
+export function POST($) {
+  const { path, body, context } = $;
+  const created = context.ops.ai_models.add(
+    { workspaceId: path.id, projectId: path.project_id },
+    { model: { id: body.model_id }, prompts_count: 0 },
+  );
+  return $.response[200].json(created);
```
aliciadriani: **Nit (no change required).** This `POST` returns `200` while the `projects` create returns `201`. CI response-validation is on and green, so `200` is what the spec declares for `add ai_model` — the inconsistency is spec-driven, not a bug. Noting only for awareness.

aliciadriani (commented):
Inline comments below carry the proposed fix for each finding from the review.

**Process — Should Fix (not inline):** the PR description is stale relative to the branch. The body says step 5 (Counterfact runner) and step 8 (E2E + CI wiring) are "Deferred to a follow-up PR," but this branch already contains `run.js`, the route handlers, `__reset.js`, the full E2E suite, and the new `E2E (project-engine mock)` CI job. The "Contents" list also omits `context.js`, `run.js`, the handlers, and the type-surface test. Please refresh the description to match the 24-file / +1579 diff so reviewers can size scope.

_Fix:_ update the PR body — move steps 5 & 8 from "Deferred" to "Contents," and add `context.js`, `run.js`, the counterfact handlers, `__reset.js`, the E2E suite, and `test/types/index.type-test.ts` to the contents list.

packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/prompts.js
```diff
@@ -0,0 +1,40 @@
...
+  const created = context.ops.prompts.createMany(
+    { workspaceId: path.id, projectId: path.project_id },
+    (body?.items ?? []).map((text) => ({ name: text })),
+  );
+  const first = created[0] ?? {};
+  return $.response[200].json({ id: first.id, name: first.name });
+}
+
+/** DELETE — batch-delete prompts (body: { ids }). */
+export function DELETE($) {
```
aliciadriani, line 22: **Nit (no change required).** `prompts.removeMany` (and `ai_models.removeMany`) compute a removed count that this handler discards — it always returns `204`. The spec DELETE is `204`, so this is correct; the count is exercised at the ops layer in `stateful.test.js`. Flagging only so the unused return value reads as deliberate.

packages/spacecat-shared-project-engine-client/src/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/prompts.js
aliciadriani: Resolved in `cbe35f3` — the bogus `POST /aio/prompts` handler was removed and the correct paths were wired:
- `aio/prompts/tagged.js` → `POST` create (201, `IDsWithStatsResponse`)
- `aio/prompts/by_tags.js` → `POST` list (200, `AIOPromptsWithStatusListResponse`)
- `aio/prompts.js` retains only `DELETE` → 204

The E2E now drives `tagged → by_tags` with a real write-then-read assertion and response validation on, closing the false-coverage gap.

packages/spacecat-shared-project-engine-client/src/mock/seeds.js
aliciadriani: Resolved in `cbe35f3` — seed prompts now use `name` (+ `tags: []`) matching the `AIOPromptWithStatus` shape, consistent with what `createMany` stores and `by_tags` returns.

packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v1/workspaces/{id}/projects/{project_id}/ai_models.js
aliciadriani: Acknowledged — spec-driven inconsistency, no action needed.

packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/prompts.js
aliciadriani: Acknowledged — the count is exercised at the ops layer; the handler deliberately discards it and always returns 204 per spec.

packages/spacecat-shared-project-engine-client/mock/store.js
aliciadriani: Added in `1f5c8cf` — the guard throws on an explicit id collision:
```js
if (entity.id && this.#collection(name).has(entity.id)) {
  throw new Error(`duplicate id ${entity.id} in collection ${name}`);
}
```
Seed-id collisions now surface immediately instead of being silently swallowed.

rainer-friederich (dismissed):
Hey @aliciadriani,

**Verdict: Request changes** - solid, well-layered mock with a clean published-surface boundary and a faithful auth model; the blocking items are all small (a cross-cutting auth-guard hardening, two real test gaps, and a self-consistency fix), none are hard bugs.
_Changes: adds a stateful Counterfact mock of the Semrush Project Engine API to project-engine-client (in-memory store, resource ops, typed factories, seeds, AI-unit quota, bearer auth, a live runner + route handlers, an in-package E2E), plus published-client `@ts-check`/type regen and a generated python model package. (58 files)._

## Must fix before merge

1. **[Important] Auth-guard injection fails open on handler-style drift** - `mock/run.js:111` (`injectAuthGuard`). The bearer guard is injected by regex-matching `export [async] function GET|POST|PUT|PATCH|DELETE($) {`. Every committed handler matches (all 30 verified, so no live bug), but a future handler in another style (`export const GET = ($) =>`, destructured params, a new `OPTIONS`/`HEAD`) gets no guard and serves unauthenticated silently - and `run.js` is coverage-excluded, exercised only by the gated E2E. Fix: assert at materialization that every real-route method received the guard (throw on miss), or add a unit test enumerating `counterfact/routes/**`.

2. **[Important] Child-resource writes skip parent-project existence** - `mock/counterfact/routes/v2/.../ai_models.js:25-34` (and the v1 sibling / prompts / benchmarks). `POST .../ai_models` calls `ops.ai_models.add` with no check the project exists, returning 201 where the live API 404s. A consumer flow that adds a model to a deleted project passes the mock but fails prod - the exact bug class the mock exists to catch. Fix: a shared `requireProject(scope)` guard the child writers run first, or document the gap in `mock-statefulness.md`.

3. **[Important] `by_tags` tag-filter branch untested even by the E2E** - `mock/counterfact/routes/v2/.../aio/prompts/by_tags.js:29-31`. The E2E only calls `by_tags` with `tag_ids: []` (the list-all branch); the filter branch - the consumer's actual "list by tags" read path, with OR semantics - is never executed, and handlers are coverage-excluded so nothing flags it. Fix: one test creating tagged prompts under two tags and asserting a non-empty `tag_ids` returns only the matching subset.

4. **[Important] Published `src/internal.js` retry change shipped without a test** - `src/internal.js:153`. `result && typeof result.catch === 'function'` became `result instanceof Promise`, a behavior tightening (a non-Promise thenable's rejection now escapes the swallow). The existing test stays green because a native `async` fn returns a real Promise, so the change is unpinned. It is defensible (out-of-contract input) but it ships in published code. Fix: add an assertion pinning the chosen behavior, or revert to the duck-typed thenable check.

5. **[Important] Factory-pattern doc/code inconsistency** - `init_status.js:22` (`{ initialized: true }`), `publish.js:32` (`{ message: 'publish accepted' }`), `brand_urls.js` DELETE. These emit inline response literals, but the package `CLAUDE.md` states the lone exception to the factory rule is `ci/competitors`. Fix: add `createBasicResponseMock`/`createInitStatusMock` and route through them, or amend the CLAUDE.md exception list to name the scalar wrappers.

<details>
<summary>Non-blocking (13): minor issues and suggestions</summary>

- suggestion: CI workflow has no `timeout-minutes` (`project-engine-mock-e2e.yaml`, job `runs-on` at line 37) - a wedged boot inherits GitHub's 6h default; add `timeout-minutes: 15`.
- suggestion: E2E spawns the server with `stdio: 'ignore'`, so a boot/transpile failure surfaces only as the opaque "mock did not become ready in time"; capture child stderr and include the tail in the readiness-timeout error.
- nit: `.nycrc.json` excludes `run.js` + `counterfact/routes/**` (justified) - add a one-line comment pointing at the E2E as the compensating control.
- suggestion: E2E teardown does `Promise.race([exited, sleep(5000)])` but never escalates to `SIGKILL` if SIGTERM is ignored.
- suggestion: `LIB_FILES` in `run.js` is a hand-maintained list; it could self-validate by parsing `context.js`'s sibling imports (fails loudly at boot today, but far from the edit).
- nit: `ci/competitors` is the one handler still using an inline literal (`CICompetitor` has no factory) - add `createCiCompetitorMock` to close the single-source-of-truth loop.
- nit: package `CLAUDE.md` says `globalThis.crypto` is typed via lib `dom`, but `tsconfig.json` uses `"lib": ["esnext"]`; the type comes from `@types/node`.
- suggestion: store deep-clone isolation has no test mutating a `create()`/`get()`/`list()` return and re-reading (the "mutate a returned entity must not corrupt the store" case); the clone path is covered, so this is a hardening gap.
- nit: `__quota` with a missing `workspaceId` calls `quota.set(undefined, ...)` and returns `{ id: undefined }`; a one-line 400 guard would make the control route self-explaining.
- nit: E2E `baseUrl` uses `http://localhost:${port}` while `pickPort` binds on `127.0.0.1`; pin both to `127.0.0.1` to drop a DNS variable.
- suggestion: `auth.js` - add a whitespace-only-token canary test (`'Bearer    '`) and a comment that `.trim()` is what makes a trailing-whitespace token fail, locking the `.trim()`+`\S` contract.
- suggestion: the E2E readiness probe could hit the auth-exempt `/__dump` instead of a seeded route, removing the `Bearer readiness-probe` workaround and the seed dependency.
- nit: `python/serenity_project_engine/http_server.py` is an inert datamodel-codegen pydantic model (no server) - the filename misleads.

</details>

- The api-service consumer `getInitStatus` calls the dead `/v1` init_status path (degrades to `initialized: null`); the mock faithfully reproduces the live 404 (overlay CR8). The real fix is a spacecat-api-service change - worth a tracking ticket so this PR's CR8 note isn't the only record.
- The E2E proves mock-vs-spec, not consumer-wiring end-to-end; that lands with the adoption PR.

---

No Critical issues, and the architecture/layering, published-surface boundary (`files: ["src"]`, counterfact is a devDependency), supply chain, and bearer-auth model all reviewed clean (the auth model was independently confirmed faithful). The Request-changes state is driven entirely by the 5 Important items above, all fixable on-branch.

rainer-friederich:
Ran a second local multi-agent review (6 agents: staff-engineer, architect, project-conventions, security, sre, qa) over the post-fix tree (this PR's ~4.9k-LOC diff is over the MysticatBot limit, so the local pass is the stand-in). All five prior-review fixes were independently verified as correctly resolved. The new findings are addressed in `dbcd395d`.

Addressed:

- [Important] Design-of-record understated the shipped stateful set - `docs/mock-statefulness.md` listed 3 stateful resources but the implementation ships 5 (`STATEFUL_RESOURCES` = projects, ai_models, prompts, benchmarks, brand_urls). Done: added benchmarks + brand_urls to the consumer-inventory table and the confirmed-set line, with the consumer write-then-read justification; the implementation was the correct side.
- nit: `injectAuthGuard`'s comment claimed the throw catches arrow handlers and `OPTIONS`/`HEAD`, but the regexes caught neither. Done: extracted it to a pure `mock/inject-auth-guard.js` and broadened the declared-method count to every shape the comment names (`export const VERB =`, plus `OPTIONS`/`HEAD`), so an unguardable shape now trips the count-vs-guarded throw - the guarantee matches the code.
- suggestion: the auth-bypass safety net had no unit test (`run.js` launches the server at import and is coverage-excluded). Done: the extraction makes it a pure transform; added `test/mock/inject-auth-guard.test.js` (conforming handler guarded, `__*` untouched, arrow/OPTIONS/renamed-param throw) at 100%.
- suggestion: the benchmark `PUT` (`updateBenchmark`) was the only consumer-critical stateful route the E2E never drove. Done: added a create -> PUT `brand_aliases` -> list-reflects case.
- nit: `mock/quota.js` described the over-allocation body as nginx-style, but handlers emit JSON `{ message }`. Done: corrected the comment (consumer keys on the 405 status, not the body).
- nit: package enforces `branches: 100` vs the root CLAUDE.md's 97%. Done: documented the deviation + the coverage exclusions in the package CLAUDE.md.
- suggestion: `existing_count` is hard-wired to 0 (no dedup modeled). Done: documented next to the other fidelity simplifications.
- suggestion: clarify `./mock/*` is intentionally unpublished. Done: strengthened the package CLAUDE.md note (workspace-symlink-only, never the registry tarball).
- (tracking) unauthenticated `__*` control routes if ever shared-hosted. Done as a doc caveat: added a bind-to-localhost note in `docs/mock-usage.md`.

Declined / no change:
- `launch()` handles `exit` but not `error` (sre) - declined: a spawn failure's stderr still reaches the captured stream and is dumped on readiness timeout, so it stays diagnosable; no silent path.
- v1/v2 `ai_models` POST handlers are near-identical (staff) - declined: extracting a shared helper across two untyped materialized handlers adds more than it saves; left as-is.
- E2E rebuilds `build/openapi3.json` every CI run rather than caching (sre) - out of scope: CI-cost only, not correctness; not changed here.

Gates after `dbcd395d`: types, lint, unit 162 @ 100%, e2e 31.

packages/spacecat-shared-project-engine-client/test/e2e/project-engine-mock.e2e.js
```diff
@@ -0,0 +1,732 @@
...
+    expect(created.existing_count).to.equal(0);
+
+    const { data: listed } = await client.GET(
+      '/v1/workspaces/{id}/projects/{project_id}/ai_models/benchmarks',
+      { params: { path: { id: SEED_WORKSPACE, project_id: SEED_PROJECT } } },
+    );
+    // seeded own-brand + the new competitor
+    expect(listed.aio_benchmarks.map((b) => b.id)).to.include(created.ids[0]);
+
+    await client.DELETE('/v1/workspaces/{id}/projects/{project_id}/ai_models/benchmarks', {
```
rainer-friederich: issue (blocking): The empty-body 202 contract is locked only for `publish` (the raw-fetch `expect(await rawPub.text()).to.equal('')` at ~line 600). The comment at line 594 says delete-benchmarks, update-benchmark, AND delete-brand-urls all return a 202 with an EMPTY body, but those three tests assert state only - `client.DELETE`/`client.PUT` swallow the response body, so a regression that drops `body: ''` and returns a `{ message }` envelope (the exact Counterfact `{status:202}` -> "Accepted" gotcha this PR fixes) would still pass. These handlers are coverage-excluded, so the E2E is their only gate. Add a raw-fetch `status === 202 && (await res.text()) === ''` assertion to the benchmark-delete, benchmark-PUT, and brand-url-delete cases, mirroring the publish test.

packages/spacecat-shared-project-engine-client/mock/seeds.js
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * Pass `quota` to mirror the AI-unit allocation the sub-workspace was granted (via user-manager);
+ * the project-engine mock then meters project/prompt creates + publish against it. Omit for the
+ * unlimited (limits-disabled) default.
+ *
+ * @param {{ workspaceId: string, projects?: SeedProject[],
+ *   quota?: { projects?: number | null, prompts?: number | null } }} spec
+ * @returns {import('./store.js').Snapshot}
+ */
+export function buildSeed({ workspaceId, projects = [], quota }) {
```
rainer-friederich, line 147: issue (blocking): `buildSeed`'s docstring says it lets a caller "mirror the rows it inserted into Postgres," but `SeedProject` only accepts `aiModels` and `prompts` - benchmarks and brand_urls (both now part of the stateful set, and both present in `WORKSPACE_WITH_DATA` above) can't be expressed. The cross-repo harness this helper exists for would have to hand-write raw `collectionKey(...)` snapshot JSON, bypassing the factory drift-guard. Either extend `SeedProject` with optional `benchmarks`/`brandUrls` routed through `collectionKey` (the same way `aiModels`/`prompts` are handled on lines 152-153), or narrow the docstring to state the helper covers projects/ai_models/prompts only.

rainer-friederich (dismissed):
Hey @aliciadriani,

**Verdict: Request changes** - the shipping mock code is correct and live-verified; two small, in-scope gaps are worth closing first (a test that doesn't lock a contract it claims to, and a seed helper that can't express two of its own resource types).
_Changes: adds a stateful Counterfact mock (store, stateful ops, typed factories, seeds, quota meter, bearer-auth guard, route handlers, runner), a spec overlay, generated types, unit + gated-E2E tests, docs, and a path-gated CI job to spacecat-shared-project-engine-client (64 files)._
Note: this PR's diff (~5.3k LOC) is over the MysticatBot limit, so this is the local multi-agent stand-in (7 reviewers). CI is green.

## Must fix before merge
1. [Important] E2E asserts state only for delete-benchmarks / update-benchmark / delete-brand-urls - the empty-body 202 contract is locked only for `publish` - `packages/spacecat-shared-project-engine-client/test/e2e/project-engine-mock.e2e.js:516` (details inline)
2. [Important] `buildSeed` can't seed benchmarks/brand_urls despite its docstring's "mirror the rows it inserted into Postgres" promise - `packages/spacecat-shared-project-engine-client/mock/seeds.js:143` (details inline)

<details>
<summary>Non-blocking (9): suggestions and nits</summary>

- suggestion: derive `LIB_FILES` from `context.js`'s actual imports (or assert the `_lib/` set equals them after materialize) instead of the hand-maintained list - a missing entry fails the E2E boot, but the manual sync is a footgun the next contributor will hit - `mock/run.js:87`.
- suggestion: `QUOTA_COLLECTION` test asserts the constant equals `'quota'` (tautological); instead `quota.set(...)` then `store.list(QUOTA_COLLECTION)` to prove they're wired at runtime - `test/mock/quota.test.js`.
- suggestion: `factories.type-test.ts` type-asserts only 5 of 12 factories; add `createBenchmarkMock` and `createBrandUrlMock` (the ones the overlay drift-guard exists for) - `test/types/factories.type-test.ts`.
- suggestion: add a `push: { branches: [main] }` trigger (or keep the documented PR-only-signal intent on purpose) so the live-E2E + type-surface check also run post-merge, not just on PRs - `.github/workflows/project-engine-mock-e2e.yaml:16`.
- nit: `beforeEach`'s `POST /__reset` throws an opaque "fetch failed" if the mock died mid-suite; guard on the existing `serverExited` flag for an actionable message - `test/e2e/project-engine-mock.e2e.js`.
- nit: copy the lib files with `readFileSync(join(here, file), 'utf8')` - the current call returns a Buffer; it works only because `writeFileSync` writes the bytes - `mock/run.js:133`.
- nit: one-line comment on `GUARDABLE_METHOD` explaining why `OPTIONS`/`HEAD` are in `DECLARED_METHOD` but intentionally absent here (the fail-closed asymmetry) - `mock/inject-auth-guard.js`.
- nit: the PR body "## Validation" counts (147 unit / 29 e2e) are stale vs HEAD (the latest fix round reports 165 unit / 33 e2e); refresh them.
- nit: `quota.set` has a read-then-write that's a TOCTOU only if the store ever becomes multi-writer; a one-line note on the single-writer invariant would age well - `mock/quota.js`.
</details>

- The overlay (`corrections.yaml`, 9 corrections) is verified manually and timestamped inline; there is no automated freshness gate, so a future upstream Semrush change can silently diverge from a correction. A scheduled contract test against the sandbox workspace would close the loop - belongs with the adoption PR / the Jira epic, not here.
- CI uses floating action tags (`actions/checkout@v6`, etc.); this matches the repo-wide `main.yaml` pattern, so SHA-pinning is a repo-wide pass, not this PR.
- The e2e job passes `MYSTICAT_DATA_SERVICE_REPO_READ_TOKEN` through the shared `setup-node-npm` action (same as `main.yaml`). It's the monorepo install convention; if the root `npm ci` provably doesn't need it for this workflow it could be dropped, but that's an action-level change, not introduced here.

The architecture is sound and the review surfaced no bugs in shipping code: the generic-store -> stateful-ops -> typed-factories -> Counterfact-handlers layering is clean, the auth guard is fail-closed at the single materialization seam (the prior fail-open drift bug is fixed and the drift vectors are unit-proven), the overlay-as-source-of-truth with a gitignored build artifact is the right call, and the 100%-unit / E2E-for-the-server-surface split is the correct trade-off for a mock that can't run under `npm test`. The live-fidelity verification (timestamped against a real workspace) is exemplary.

packages/spacecat-shared-project-engine-client/mock/seeds.js
rainer-friederich: Fixed in `81ff578f` — `SeedProject` now accepts `benchmarks` and `brandUrls` (grouped by benchmark id), routed through `collectionKey` like `aiModels`/`prompts`. Added a seeds.test case that loads both and reads them back via the stateful ops.

packages/spacecat-shared-project-engine-client/test/e2e/project-engine-mock.e2e.js
rainer-friederich: Fixed in `81ff578f` — the delete-benchmarks, update-benchmark (PUT v1), and delete-brand-urls e2e cases now raw-fetch and assert `status === 202 && (await res.text()) === ""`, matching the publish lock. A regression to a `{message}` body now fails the suite.

rainer-friederich (approved):
LGTM - follow ups will come

MysticatBot (approved):
Hey @aliciadriani,

**Verdict: Approve** - all prior blocking findings addressed; the mock is well-layered, live-verified, and ships with strong type-level and E2E coverage.
**Complexity: HIGH** - large diff (5300+ LOC, 60 files); auth + dependencies + CI/CD + API surface signals.
_Changes: adds a stateful Counterfact mock of the Semrush Project Engine API to project-engine-client (in-memory store, resource ops, typed factories, seeds, AI-unit quota, bearer auth, a live runner + route handlers, an in-package E2E), plus published-client type regen and a path-gated CI job (60 files)._

<details>
<summary>Non-blocking (4): suggestions and nits</summary>

- suggestion: E2E `baseUrl` uses `http://localhost:${port}` while `pickPort` binds on `127.0.0.1`; pin both to `127.0.0.1` to eliminate a DNS variable - `test/e2e/project-engine-mock.e2e.js`
- suggestion: E2E spawns the server with `stdio: 'ignore'`, so a boot/transpile failure surfaces only as an opaque readiness timeout; capturing child stderr and including the tail in the timeout error would speed debugging - `test/e2e/project-engine-mock.e2e.js`
- nit: `python/serenity_project_engine/http_server.py` is an inert pydantic model (no server logic); the filename misleads - rename to `server_models.py` or similar
- nit: the PR body "## Validation" counts (170 unit / 34 e2e) may be stale vs the final HEAD after fix rounds; refresh before merge so the archive is accurate

</details>

- The api-service consumer `getInitStatus` calls the dead `/v1` init_status path (degrades to `initialized: null`); the mock faithfully reproduces the live 404 (overlay CR8). The fix is a spacecat-api-service change.
- The overlay (`corrections.yaml`, 10 corrections) has no automated freshness gate against the upstream Semrush spec; a scheduled contract test would close the loop.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 4s | <b>Cost:</b> $9.73 | <b>Commit:</b> <code>55316f37fa02c84b8e3144efe9e864692cc3868f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1667, opened 2026-06-11
feat(brand-governance): replace /checks with /profile endpoint in BrandGovernanceClient

anagarwa:
## Summary

- Replaces the paginated `/checks?scope=COPY` endpoint with the single `/profile` endpoint in `BrandGovernanceClient.getBrandGuidelinesForUrl()`
- Returns the raw Brand Governance Agent profile response (`{ data: { identity, voice_and_tone, language, messaging, editorial } }`) directly to callers — no mapping needed server-side
- Returns `null` on 404 from either `/from-url` or `/profile`, preserving the existing fallback contract for callers

## Test plan

- [x] Replaced `mockBrandChecks` helper with `mockBrandProfile` in test suite
- [x] Updated success test to verify raw profile data is returned as-is
- [x] Added test for 404 from `/profile` endpoint (returns null)
- [x] Updated header verification test to assert against `/profile` endpoint
- [x] Updated fresh-token test to use `/profile` mock
- [x] All 37 tests passing, 100% line/branch/function coverage

### DIFF
--- a/packages/spacecat-shared-brand-client/src/brand-governance-client.js
+++ b/packages/spacecat-shared-brand-client/src/brand-governance-client.js
@@ -17,16 +17,14 @@ import { ImsClient } from '@adobe/spacecat-shared-ims-client';
 const HTTP_BAD_REQUEST = 400;
 const HTTP_NOT_FOUND = 404;
 
-const CHECKS_PAGE_SIZE = 100;
-const MAX_PAGES = 20;
-
 const API_GET_BRAND_FROM_URL = (url) => `/api/v1/brands/from-url?url=${encodeURIComponent(url)}`;
-const API_GET_BRAND_CHECKS = (brandId, page) => `/api/v1/brands/${brandId}/checks?status=ACTIVE&type=BRAND&scope=COPY&pageSize=${CHECKS_PAGE_SIZE}&page=${page}`;
+const API_GET_BRAND_PROFILE = (brandId) => `/api/v1/brands/${brandId}/profile`;
 
 /**
  * Client for the Adobe Brand Governance Agent API.
- * Resolves brand guidelines by site URL using COPY-scoped checks.
- * Returns null when the brand is not registered — caller falls back to Brand Publish.
+ * Resolves brand profile by site URL using the /profile endpoint.
+ * Returns null when the brand is not registered or has no profile — caller falls back to
+ * Brand Publish.
  */
 export class BrandGovernanceClient {
   static createFrom(context) {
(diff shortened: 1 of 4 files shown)


packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -139,45 +137,19 @@ export class BrandGovernanceClient {
...
-    return {
-      id: brand.id,
-      name: brand.name,
-      imsOrgId,
-      createdAt: brand.createdAt,
-      updatedAt: brand.updatedAt,
-      guidelines,
-    };
+    }
+    return profileResponse.json();
```
MysticatBot, line 153: issue (blocking): This return statement changes the public API contract from `{ id, name, imsOrgId, createdAt, updatedAt, guidelines: [{name, text}] }` to the raw upstream `/profile` response (`{ data: { identity, voice_and_tone, ... }, meta }`). Every consumer accessing `.guidelines`, `.id`, `.name` etc. will get `undefined` at runtime.

Since this is a shared library consumed by multiple services and semantic-release will produce a minor bump from the `feat()` prefix, consumers on `^` ranges will auto-upgrade and break silently.

Fix: Use `feat!:` or add a `BREAKING CHANGE:` footer to the commit message so semantic-release produces a major version bump. Coordinate downstream consumer updates before or alongside this release.

packages/spacecat-shared-brand-client/src/brand-governance-client.js
```diff
@@ -17,16 +17,14 @@ import { ImsClient } from '@adobe/spacecat-shared-ims-client';
...
-
 const API_GET_BRAND_FROM_URL = (url) => `/api/v1/brands/from-url?url=${encodeURIComponent(url)}`;
-const API_GET_BRAND_CHECKS = (brandId, page) => `/api/v1/brands/${brandId}/checks?status=ACTIVE&type=BRAND&scope=COPY&pageSize=${CHECKS_PAGE_SIZE}&page=${page}`;
+const API_GET_BRAND_PROFILE = (brandId) => `/api/v1/brands/${brandId}/profile`;
 
 /**
  * Client for the Adobe Brand Governance Agent API.
- * Resolves brand guidelines by site URL using COPY-scoped checks.
- * Returns null when the brand is not registered — caller falls back to Brand Publish.
+ * Resolves brand profile by site URL using the /profile endpoint.
```
MysticatBot, line 25: issue (blocking): The method is still named `getBrandGuidelinesForUrl` but now returns a brand profile (identity, voice_and_tone, language, editorial) - not guidelines. The JSDoc was updated to say "profile" but the method name was not.

Fix: Rename to `getBrandProfileForUrl` (breaking regardless, so bundle with the major bump). If backward compat is needed short-term, add the new name and deprecate the old.

MysticatBot (dismissed):
Hey @anagarwa,

**Verdict: Request changes** - the return type change is a breaking change that needs a major version signal and TypeScript declaration update.
_Changes: Replaces paginated `/checks` endpoint with single `/profile` endpoint in BrandGovernanceClient, returning raw upstream response instead of mapped guidelines object (2 files)._

## Must fix before merge
1. [Critical] Breaking change shipped under minor semver bump - consumers accessing `.guidelines`, `.id` will get `undefined` - `brand-governance-client.js:153` (details inline)
2. [Important] Method name `getBrandGuidelinesForUrl` no longer matches the profile data it returns - `brand-governance-client.js:25` (details inline)
3. [Important] TypeScript declaration (`src/index.d.ts`) still declares return type `BrandGovernanceGuidelines` with `{ id, name, imsOrgId, createdAt, updatedAt, guidelines }` - now actively incorrect after this change. Update or replace the interface to reflect the actual `/profile` response shape.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 44s | <b>Cost:</b> $2.97 | <b>Commit:</b> <code>7868be281fd6470d6e656c63526d6d9acdde39c7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

duynguyen (approved):
## Review

Clean simplification — removing the pagination loop, `CHECKS_PAGE_SIZE`, and `MAX_PAGES` constants is a nice improvement. Tests are well-structured and the new 404-from-profile test is a good addition.

**One nit:** `return profileResponse.json()` on the last line — prefer `return await profileResponse.json()` for style consistency with the rest of the file and to keep the stack trace intact if `.json()` throws.

**Heads up:** this is a breaking contract change (return type goes from `{ id, name, imsOrgId, createdAt, updatedAt, guidelines }` to the raw `{ data, meta }` response). Make sure the consuming service PRs are merged alongside this one.

anagarwa (commented):
@MysticatBot 
Thanks for the thorough review! Addressing all three points:

**1. Semver bump — no major needed**
There are no existing consumers of `BrandGovernanceClient.getBrandGuidelinesForUrl()` in production today. This is the first integration — the companion PRs in `spacecat-api-service` and `mystique` are being merged alongside this one. Since no deployed service is currently reading `.guidelines`, `.id`, or `.name` from the return value, there is no consumer to silently break. A minor bump is appropriate here.

**2. Method name — keeping `getBrandGuidelinesForUrl`**
The method name reflects its role in the **SpaceCat Brand Guidelines API**, not the Brand Governance Agent's internal endpoint name. From the caller's perspective (spacecat-api-service `/sites/:siteId/brand-guidelines`), this method resolves brand guidelines for a URL — which is exactly what it does. Renaming it to `getBrandProfileForUrl` would leak the Brand Governance Agent's internal concept into our API surface. The method name stays.

**3. TypeScript declaration — updated**
Fixed in commit `9e42c012`. Replaced the stale `BrandGovernanceGuideline` + `BrandGovernanceGuidelines` interfaces (which reflected the old `/checks`-based shape) with a minimal `BrandGovernanceProfile { data: Record<string, unknown>; meta?: Record<string, unknown> }`. Since the client is a raw passthrough of the `/profile` response, we keep the type intentionally loose rather than over-specifying every nested field.

@MysticatBot review again ?

ymadobe (approved):

ymadobe (approved):

PR #1670, opened 2026-06-12
fix: increase DRS createschedule api timeout

dipratap:

## Related Issues

Thanks for contributing!


MysticatBot (approved):
Hey @dipratap,

**Verdict: Approve** - clean, well-scoped change with only minor tightening suggestions.
_Changes: Adds an optional `timeout` parameter to `createExperimentSchedule` that forwards through to the underlying fetch call (3 files)._

<details>

- suggestion: `#request` accepts an open-ended `fetchOptions` bag and spreads it after `headers` - since only `timeout` is needed, narrowing the signature to `async #request(method, path, body = undefined, { timeout } = {})` would prevent accidental header override by future callers - `packages/spacecat-shared-drs-client/src/index.js:138`
- nit: test "forwards timeout to fetch when provided" only proves the call does not break with `timeout` set; nock cannot observe fetch options, so a regression that drops the timeout would pass silently - a spy on `tracingFetch` asserting `{ timeout: 12000 }` in the options would close the gap - `packages/spacecat-shared-drs-client/test/index.test.js:937`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 6s | <b>Cost:</b> $3.19 | <b>Commit:</b> <code>91f9649e80e2eecb791de49514a2efdea91ea850</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1671, opened 2026-06-12
feat(utils): add LLMO claims guidance schema

davidaurelio:
## Summary
- Adds optional `claims.brandContext` and `claims.sentimentGuidance` to the LLMO config schema with a 4000 character max per field.
- Preserves claims guidance through S3 LLMO config read/write validation.
- Adds schema and S3 read/write tests for valid claims, empty claims, overlong values, and wrong types.

## Tests
- Not run (blocked): `npm test --workspace @adobe/spacecat-shared-utils -- test/schemas.test.js test/llmo-config.test.js` failed because the local workspace cannot resolve `ipaddr.js` from `spacecat-shared-utils/src/network-policy.js`.
- Not run (blocked): `CI=true pnpm --filter @adobe/spacecat-shared-utils exec mocha ...` failed because the root `pnpm-lock.yaml` is out of sync with the root `package.json` and pnpm aborts before running tests.


packages/spacecat-shared-utils/src/schemas.js
```diff
@@ -38,6 +38,7 @@ const nonEmptyString = z.string().min(1);
 
 const region = z.string().length(2).regex(/^[a-z][a-z]$/i);
 const awsRegion = z.string().regex(/^[a-z]{2}(?:-[a-z]+)+-\d+$/i);
+const claimsGuidance = z.string().max(4000);
```
akshaymagapu, line 41: `z.string().min(1).max(4000)`
small nit: should we have min check as well?

akshaymagapu (approved):

packages/spacecat-shared-utils/src/schemas.js
davidaurelio: no, it's allowed to be empty

akshaymagapu (approved):

PR #1674, opened 2026-06-15
feat(launchdarkly-client): eliminate Lambda log noise (cache + polling + init timeout)

solaris007:
## Problem

LaunchDarkly SDK lines dominate SpaceCat prod `error` logs. Over a representative window (Coralogix prod, `spacecat-services-prod`, `level=error`, 2026-06-15 00:00:00Z–05:45:00Z, ~5.75h): **105,225 of 148,521 error events (~71%) are LaunchDarkly noise** (`will retry stream connection`, `waitForInitialization ... without a timeout`, `received I/O error ... will retry`). `api-service` alone accounts for 105,912 of those errors. These are not real errors; they drown genuine operational signal and inflate error-rate metrics.

## Root cause

1. The auth-path LD client running in prod is `launchdarkly-client@1.0.4` (a **nested** copy under `spacecat-shared-http-utils`, which hard-pins `1.0.4`). That version passes **no `logger`** to `ld.init()`, so the SDK uses `basicLogger`, which writes every level to stderr → Lambda tags it all `error` (hence the `info:`/`warn:` prefixes in the messages).
2. The IMS auth handler creates a **new streaming LD client on every authenticated request** (`http-utils` `ims.js` / `read-only-admin-wrapper.js`, no caching), each opening a streaming connection and calling `waitForInitialization()` with no timeout. Streaming can't survive Lambda freeze/thaw, so each invocation reconnects and retries.
3. The prior logger fix (#1511, `1.2.x`) never reached prod because the nested `1.0.4` pin shadows it.

## What this PR does (Phase 1 of 3)

Behavior change is confined to `spacecat-shared-launchdarkly-client`:

- **Singleton cache per `sdkKey`** at module scope — one SDK client per warm Lambda container instead of one per request. Same-instance and cross-instance/in-flight callers reuse it; a failed init clears the cache entry so the next call retries.
- **Polling instead of streaming** (`stream: false`, `pollInterval: 30`) — removes the streaming reconnect storm and drops `launchdarkly-eventsource` from the hot path. Uses the **stable** top-level options; the newer `dataSystem` config is still `@experimental`, so it is deliberately avoided.
- **Bounded init** — `waitForInitialization({ timeoutSeconds: 5 })`, removing the "without a timeout" warning and capping cold-start latency; fail-open is preserved (the auth call sites already catch).
- **Logger mapping retained** (`error → log.error`; `warn`/`info`/`debug → log.debug`) as defense-in-depth.
- **SDK bump** `@launchdarkly/node-server-sdk` `^9.9.7 → ^9.11.2` (routine hygiene, behavior-neutral; verified `stream`/`pollInterval` stable and `dataSystem` still experimental on the bumped version).

## Cascade (this PR is only step 1)

The fix only reaches prod after the version cascade — Phase 1 alone does nothing in prod until consumers pick it up:

1. **This PR** → publishes `launchdarkly-client` (minor, → `1.3.0`).
2. **Next:** bump the `launchdarkly-client` pin in `spacecat-shared-http-utils` (`1.0.4 → ^1.3.0`) — this is what removes the shadowing nested copy on the auth path.
3. **Then:** bump `spacecat-api-service` to consume the new `http-utils` + dedupe to a single hoisted client, redeploy, and verify in Coralogix (dev → prod).

## Testing

- `npm test -w packages/spacecat-shared-launchdarkly-client` → 43 passing, **100% statements/branches/functions/lines**.
- `npm run lint -w packages/spacecat-shared-launchdarkly-client` → clean.
- New tests: cross-instance cache reuse, in-flight init join, polling-mode + init-timeout options, and cache-cleared-on-failure retry.

## Out of scope

The `url.parse` `DEP0169` DeprecationWarning and LD daemon mode (DynamoDB feature store) are explicitly out of scope (see spec).

## Design docs (included in this PR)

- `docs/plans/2026-06-15-launchdarkly-lambda-noise-design.md`
- `docs/plans/2026-06-15-launchdarkly-lambda-noise-plan.md`

### DIFF
--- a/packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
+++ b/packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
@@ -87,7 +87,6 @@ class LaunchDarklyClient {
       debug: (...args) => log.debug('[LaunchDarkly]', ...args),
     };
     this.client = null;
-    this.initPromise = null;
   }
 
   /**
@@ -113,9 +112,11 @@ class LaunchDarklyClient {
 
     const initPromise = (async () => {
       const client = ld.init(this.sdkKey, {
+        ...this.options,
+        // Forced after the spread: a consumer's options must not be able to
+        // re-enable streaming (the whole point of this change) or swap the logger.
         stream: false,
         pollInterval: DEFAULT_POLL_INTERVAL_SECONDS,
-        ...this.options,
         logger: this.sdkLogger,
       });
       await client.waitForInitialization({ timeoutSeconds: DEFAULT_INIT_TIMEOUT_SECONDS });
--- a/packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
+++ b/packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
@@ -122,6 +122,9 @@ class LaunchDarklyClient {
       return client;
     })();
 
+    // The first caller's sdkLogger is baked into the cached client; later
+    // callers reuse it. Per-instance this.log is still used for flag-evaluation
+    // logging in variation(), so request-scoped logs are unaffected.
     sdkClientCache.set(this.sdkKey, { initPromise });
 
     try {
--- a/packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
+++ b/packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
@@ -13,6 +13,21 @@
 import * as ld from '@launchdarkly/node-server-sdk';
 
 const DEFAULT_API_BASE_URL = 'https://app.launchdarkly.com';
+const DEFAULT_POLL_INTERVAL_SECONDS = 30;
+const DEFAULT_INIT_TIMEOUT_SECONDS = 5;
+
+// Cache of initialized SDK clients, keyed by sdkKey. A warm Lambda container
+// reuses one client across invocations instead of opening a new connection per
+// request. Each entry holds the in-flight/resolved init promise.
+const sdkClientCache = new Map();
+
+/**
+ * Resets the module-scope SDK-client cache. Exported for tests; not part of the
+ * normal runtime flow.
+ */
+export function clearClientCache() {
+  sdkClientCache.clear();
+}
 
 /**
  * LaunchDarkly Client wrapper for SpaceCat services.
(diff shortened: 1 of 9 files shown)


packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
```diff
@@ -89,23 +104,38 @@ class LaunchDarklyClient {
...
-        this.log.info('LaunchDarkly client initialized successfully');
-      } catch (error) {
-        this.log.error('Failed to initialize LaunchDarkly client:', error);
-        throw error;
-      }
+    const initPromise = (async () => {
+      const client = ld.init(this.sdkKey, {
+        stream: false,
+        pollInterval: DEFAULT_POLL_INTERVAL_SECONDS,
+        ...this.options,
```
MysticatBot: issue (blocking): The spread order `{ stream: false, pollInterval: 30, ...this.options, logger: ... }` places `stream` and `pollInterval` before `...this.options`, so any consumer that passes `options: { stream: true }` will silently re-enable streaming and defeat the fix this PR exists to ship.

Since this library is consumed by multiple services and the entire purpose of this change is to enforce polling mode, the fix should not be overridable by accident.

**Suggested fix** - move `stream` and `pollInterval` after the spread (alongside `logger`):

```js
const client = ld.init(this.sdkKey, {
  ...this.options,
  stream: false,
  pollInterval: DEFAULT_POLL_INTERVAL_SECONDS,
  logger: this.sdkLogger,
});
```

Or, if consumer override is intentional (e.g. for local dev with streaming), document that explicitly in a code comment so the next reader knows the position is deliberate.

MysticatBot (changes requested):
Hey @solaris007,

**Verdict: Request changes** - one structural concern with the options spread that can silently defeat the streaming fix.
_Changes: adds a module-scope SDK-client cache with polling mode and bounded init timeout to eliminate LaunchDarkly log noise in Lambda (9 files)._

## Must fix before merge
1. [Important] Options spread lets callers silently re-enable streaming - `packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js:118` (details inline)

<details>

- nit: Dead field `this.initPromise = null` still in constructor but no longer used after the cache refactor - remove it to avoid misleading future readers into thinking per-instance init tracking is still active - `packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js`
- nit: Second caller joining a failing cached `initPromise` propagates the error without logging (first caller already logs; consider wrapping the cached branch in try/catch for observability on this path) - `packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js:110`
- suggestion: Add a test with two different `sdkKey` values to verify separate cache entries are maintained (the Map keying contract is currently only tested with a single key) - `packages/spacecat-shared-launchdarkly-client/test/index.test.js`
- suggestion: `clearClientCache()` could call `client.close()` on each cached entry before clearing, to avoid leaked polling connections in non-Lambda contexts - `packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js:28`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 30s | <b>Cost:</b> $3.61 | <b>Commit:</b> <code>78fff88535da97dca592ef9271ab3609b9582b10</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
solaris007: Fixed in `1fde0484`. Moved `...this.options` before the forced settings so `stream: false`, `pollInterval`, and `logger` now win over any consumer-supplied options - streaming can no longer be re-enabled by accident. Added a test (`forces polling and cannot be overridden by consumer options`) that passes `options: { stream: true, pollInterval: 999 }` and asserts `ld.init` still receives `stream: false, pollInterval: 30`.

MysticatBot (approved):
Hey @solaris007,

**Verdict: Approve** - prior blocking finding is fully resolved, no new issues.
_Changes: adds a module-scope SDK-client cache with polling mode and bounded init timeout to eliminate LaunchDarkly log noise in Lambda (9 files)._

<details>

- nit: Design doc states `SDK in use: @launchdarkly/node-server-sdk@9.10.10` but implementation pins `^9.11.2` and the lockfile resolves to 9.11.2 - internal inconsistency in `docs/plans/2026-06-15-launchdarkly-lambda-noise-design.md:95`
- suggestion: Add a test for late-joiner failure propagation (two concurrent callers where init rejects - both should see the rejection via the shared promise) - `packages/spacecat-shared-launchdarkly-client/test/index.test.js`
- suggestion: Reset stub behavior (not just history) in `afterEach` to eliminate ordering fragility from the manual `mockClient.waitForInitialization.reset()` at the end of the failure test - `packages/spacecat-shared-launchdarkly-client/test/index.test.js`

</details>

- Options spread order now places forced settings (`stream: false`, `pollInterval`, `logger`) after `...this.options` so consumers cannot re-enable streaming; dedicated test added.
- Dead field `this.initPromise = null` removed.
- Separate-sdkKey cache isolation test added (`maintains separate cache entries per sdkKey`).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 51s | <b>Cost:</b> $3.58 | <b>Commit:</b> <code>1fde0484715865bc78dd94a57a4c9cca13299052</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1675, opened 2026-06-15
fix(http-utils): pin launchdarkly-client to 1.3.0 (ship Lambda log-noise fix)

solaris007:
## What

Bumps `spacecat-shared-http-utils`'s dependency on `@adobe/spacecat-shared-launchdarkly-client` from `1.0.4` to `1.3.0` (exact pin, no caret).

## Why

This is **Phase 2** of the LaunchDarkly Lambda log-noise cascade (Phase 1 = #1674, published `launchdarkly-client@1.3.0`).

`http-utils` hard-pinned `launchdarkly-client@1.0.4`, which shipped a **nested** copy on the IMS auth path (`auth/handlers/ims.js`, `auth/read-only-admin-wrapper.js`). That `1.0.4` passed no `logger` to `ld.init()` and created a new streaming client per request - the root cause of ~71% of SpaceCat prod `error` logs being LaunchDarkly noise. The nested pin also shadowed the earlier logger fix, so it never reached prod.

Pinning to `1.3.0` removes that shadowing copy so consumers (next: `spacecat-api-service`, Phase 3) resolve a single client with the cache + polling (`stream:false`) + bounded init-timeout fix.

## Verification

- `npm install` reconciles to a single `launchdarkly-client@1.3.0`; the stale nested `1.0.4` lockfile entry is removed (no `1.0.4` launchdarkly reference remains).
- `npm test -w packages/spacecat-shared-http-utils` and `npm run lint -w packages/spacecat-shared-http-utils` pass.

## Next

Phase 3: bump `spacecat-api-service` to the new `http-utils` + `launchdarkly-client@1.3.0`, dedupe to a single hoisted copy, redeploy, then validate in Coralogix (dev → prod).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29840,7 +29840,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
-        "@adobe/spacecat-shared-launchdarkly-client": "1.0.4",
+        "@adobe/spacecat-shared-launchdarkly-client": "1.3.0",
         "@adobe/spacecat-shared-utils": "1.81.1",
         "jose": "6.2.3"
       },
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @solaris007,

**Verdict: Approve** - clean dependency pin bump, no issues found.
_Changes: Bumps the `@adobe/spacecat-shared-launchdarkly-client` dependency in `http-utils` from 1.0.4 to 1.3.0, removing the nested shadowing copy from the lockfile (2 files)._

Note: CI Test check is still pending - confirm it passes before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 23s | <b>Cost:</b> $1.88 | <b>Commit:</b> <code>6669eda1755da39d819f26ddfc4f2c45996c965c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1678, opened 2026-06-15
fix: downgrade expected RUM domainkey 404s from error to debug

solaris007:
## What

A `404` from `GET bundles.aem.page/domainkey/{domain}` means the domain was never onboarded to RUM. For SEO/LLMO-only sites this is the expected steady state, not a failure - yet it was being logged at `error` level on every audit/import run.

This was the single largest source of prod `ERROR`-severity log lines (~45%) across `spacecat-audit-worker` and `spacecat-import-worker`, spread over hundreds of distinct real customer domains.

## Changes

**`spacecat-shared-rum-api-client` (`src/client.js`)**
- `_exchangeDomainkey` now attaches `error.status = resp.status` when the admin-key fetch returns a non-ok response.
- `query`, `queryMulti`, and `queryStream` catch-blocks propagate `error.status = e.status` through the re-wrap. Previously these created a fresh `Error` and dropped all custom properties, so callers could never see the underlying HTTP status.

**`spacecat-shared-utils` (`src/url-helpers.js`)**
- `wwwUrlResolver` (the URL resolver used by 15+ audits) now checks `e.status === 404` in both catch-blocks and logs at `debug` ("site not onboarded to RUM") instead of `error`. Non-404 failures still log at `error`.
- `wwwUrlResolver` takes `rumApiClient` as an injected parameter, so this does not add a `utils -> rum-api-client` dependency.

## Tests
- rum-api-client: status attached on a 404 from `retrieveDomainkey`, and propagated through the `query` re-wrap.
- utils: 404 domainkey miss -> `debug` (not `error`) in both catch-blocks; mixed 404-then-500 case still logs `error` for the non-404.
- Both packages: 100% lines/statements, branch thresholds met. Lint clean.

## Rollout
This is Phase A of a 3-phase cascade. Once `rum-api-client` and `utils` publish, the consumers re-pin (exact, no caret):
- `spacecat-import-worker` - consume + branch the per-importer catch on `e.status === 404`.
- `spacecat-audit-worker` - re-pin both packages (no code change; the fix lives in the shared packages).

### DIFF
--- a/packages/spacecat-shared-rum-api-client/src/client.js
+++ b/packages/spacecat-shared-rum-api-client/src/client.js
@@ -92,7 +92,9 @@ export default class RUMAPIClient {
     });
 
     if (!resp.ok) {
-      throw new Error(`Error during fetching domainkey for domain '${domain} using admin key. Status: ${resp.status}`);
+      const error = new Error(`Error during fetching domainkey for domain '${domain} using admin key. Status: ${resp.status}`);
+      error.status = resp.status;
+      throw error;
     }
 
     try {
@@ -143,7 +145,9 @@ export default class RUMAPIClient {
       this.log.debug(`Query "${query}" fetched ${bundles.length} bundles`); // maybe even remove?
       return handler(bundles, opts);
     } catch (e) {
-      throw new Error(`Query '${query}' failed. Opts: ${JSON.stringify(sanitize(opts))}. Reason: ${e.message}`);
+      const error = new Error(`Query '${query}' failed. Opts: ${JSON.stringify(sanitize(opts))}. Reason: ${e.message}`);
+      error.status = e.status;
+      throw error;
     }
   }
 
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @solaris007,

**Verdict: Approve** - clean, well-tested fix that correctly addresses a high-volume log noise problem.
_Changes: Propagates HTTP status through RUM API client error re-wraps so wwwUrlResolver can downgrade expected domainkey 404s from error to debug level (4 files)._

<details>

- suggestion: Use `{ cause: e }` in the catch-rethrow blocks to preserve the original stack trace for production debugging. Currently only `e.message` (via string interpolation) and `e.status` survive the re-wrap; the original throw site is lost. Example: `new Error(\`Query '${query}' failed. ...\`, { cause: e })` - `packages/spacecat-shared-rum-api-client/src/client.js:147`
- nit: Typo "Could not retrieved" should be "Could not retrieve" in the retained error log path - `packages/spacecat-shared-utils/src/url-helpers.js:396`
- suggestion: Add one test asserting `.status` propagation through `queryMulti` or `queryStream` to match the explicit `query` path coverage - `packages/spacecat-shared-rum-api-client/test/index.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 12s | <b>Cost:</b> $2.44 | <b>Commit:</b> <code>0bfd2386433e296158ba6781f3b71e4b602fdb50</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1683, opened 2026-06-16
fix: pageCitability url-upsert + metrics-store $metadata 404 detection

solaris007:
Two recoverable-log / data-integrity fixes surfaced by the prod ERROR-noise review.

## Change A - metrics-store missing-key detection (spacecat-shared-utils)
`getStoredMetrics` distinguished a missing S3 object (expected, warn) from a genuine infra error (error) via a message-substring check. Switched to the structured AWS SDK v3 signal: `e.name === 'NoSuchKey' || e.$metadata?.httpStatusCode === 404`. More robust than substring matching; no behavior change for real S3 NoSuchKey errors.

## Change B - PageCitability url-upsert (spacecat-shared-data-access)
Prod emits ~100/hour of `[23505] duplicate key value violates unique constraint "idx_page_citabilities_url_unique"`. Root cause: a read-then-insert race - concurrent audit / prerender invocations build the existing-records map before either write lands, both see no row for a URL, both INSERT. `url` carries a global unique index, so the loser throws.

Fix: make PageCitability create idempotent via an upsert on `url`.
- `BaseCollection.create` gains an optional `onConflict` field (defaults to the PK `idName`, fully backward-compatible): `create(item, { upsert = false, onConflict } = {})`.
- `PageCitabilityCollection.create` overrides to `super.create(item, { upsert: true, onConflict: 'url' })` (mirrors ReportCollection's upsert pattern). Consumers (audit-worker) need no call-site change - the existing `PageCitability.create(...)` becomes a race-safe `ON CONFLICT (url) DO UPDATE`.

## Testing
- spacecat-shared-utils: 1172 passing; metrics-store.js 100% coverage; lint clean.
- spacecat-shared-data-access: 2324 passing; 100% lines / 97.76% branches (gate >= 97%); base.collection.js + page-citability.collection.js covered on both onConflict paths; lint clean.

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/base/base.collection.js
+++ b/packages/spacecat-shared-data-access/src/models/base/base.collection.js
@@ -731,7 +731,7 @@ class BaseCollection {
     return this.#queryByIndexKeys(keys, { ...options, limit: 1 });
   }
 
-  async create(item, { upsert = false } = {}) {
+  async create(item, { upsert = false, onConflict } = {}) {
     if (!isNonEmptyObject(item)) {
       const message = `Failed to create [${this.entityName}]: data is required`;
       this.log.error(message);
@@ -751,7 +751,7 @@ class BaseCollection {
 
       const prepared = this.#prepareItem(item);
       const payload = this.#toDbRecord(prepared);
-      const conflictKey = this.#toDbField(this.idName);
+      const conflictKey = this.#toDbField(onConflict ?? this.idName);
 
       let query = this.postgrestService.from(this.tableName);
       query = upsert ? query.upsert(payload, { onConflict: conflictKey }) : query.insert(payload);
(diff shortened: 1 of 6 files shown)


MysticatBot (approved):
Hey @solaris007,

**Verdict: Approve** - clean, well-targeted fixes with solid test coverage.
_Changes: Fixes a duplicate-key race in PageCitability via url-based upsert and switches S3 missing-key detection to the structured AWS SDK v3 `$metadata.httpStatusCode` signal (6 files)._

<details>

- suggestion: `onConflict` is silently ignored when `this.entity` exists (ElectroDB path) - consider a defensive guard that throws if `onConflict` is passed for a non-PostgREST entity, to prevent future callers from hitting a silent no-op - `packages/spacecat-shared-data-access/src/models/base/base.collection.js:734`
- nit: prototype-patching via `Object.getPrototypeOf(Object.getPrototypeOf(instance))` couples the test to the exact inheritance depth - a `sinon.stub(BaseCollection.prototype, 'create')` would be more resilient to future mixin/inheritance changes - `packages/spacecat-shared-data-access/test/unit/models/page-citability/page-citability.collection.test.js:68`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 7m 38s | <b>Cost:</b> $3.33 | <b>Commit:</b> <code>be008af65969550c2b40d56fa8a9a6b98374aeac</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1688, opened 2026-06-17
fix(utils): correct AMS detection for AEM 6.5 sites with Core Components

Kanishkavijay39:
## Summary

`detectAEMVersion` was misclassifying AEM 6.5 (AMS) sites as `aem_cs`. The root cause: several patterns the scorer treated as CS indicators are **not CS-exclusive** — Core Components (`cmp-` classes, `data-cmp-`), the `lc-[timestamp]-lc` clientlib format, and `/content/experience-fragments/` all ship on AEM 6.5 (AMS) too. An AMS page accumulates enough of these "CS-looking" points to overtake its own explicit AMS signals.

Discovered via `publish-prod-sgad-secretary-general-of-administration-digital.adobecqms.net` (AMS domain — `x-dispatcher: dispatcher3eusouth2`, `data-routing="ams=..."`) being returned as `aem_cs`.

## Changes — three weight recalibrations

1. **`lc-[timestamp]-lc` CS extra weight: +3 → +1** — this clientlib caching format was backported to AEM 6.5, so it's a weak CS signal, not a decisive one.
2. **`data-routing ams=` extra weight: +5 → +7** — `data-routing` is an explicit, site-authored declaration; the AMS variant must outweigh the CS-looking points an AMS page legitimately accumulates.
3. **`x-dispatcher` header extra weight: 0 → +6** — infrastructure header only emitted by AMS dispatcher nodes; now decisive, matching the MD5-clientlib signal.

## Why only `ams=` gets the boost (not `cs=`/`eds=`)

The contamination is **one-directional**. CS markup appears on AMS 6.5, but legacy AMS markup (`/etc/designs/`, `foundation-`, MD5-fingerprinted clientlibs) never appears on modern CS. So an AMS page competes against CS noise, while a CS page faces no equivalent AMS noise. Measured opposite-type points on live sites:

| AMS site | CS points it picks up | | CS site | AMS points it picks up |
|---|---|---|---|---|
| mptmd.gob.es | **7** | | eplan.com | 0 |
| honeywell.com | **5** | | deloitte.com | 0 |
| stryker.com | 1 | | aramark.com | 0 |
| | | | audemarspiguet.com | 2 (low-weight only) |

A CS site's `cs=` (+6 total) is never at risk; an AMS site's old `ams=` (+6) *was* losing to a 7–8 point CS pile. Hence `cs=`/`eds=` need no boost.

## Test plan

- [x] All 1177 tests pass locally
- [x] New regression tests added:
  - AMS 6.5 + Core Components + `lc-` + experience-fragments + explicit `ams=` → `aem_ams`
  - AMS 6.5 + Core Components + `lc-` + `x-dispatcher` header → `aem_ams`
  - Mixed-case `x-dispatcher` value (`Dispatcher3eusouth2`) → `aem_ams`
- [x] `mptmd.gob.es` now correctly classified as `aem_ams`
- [x] Spot-checked 20 live sites across all delivery types — no regressions

Jira: SITES-46675

### DIFF
--- a/packages/spacecat-shared-utils/src/aem.js
+++ b/packages/spacecat-shared-utils/src/aem.js
@@ -145,6 +145,13 @@ export function detectAEMVersion(htmlSource, headers = {}) {
     }
   }
 
+  // x-dispatcher is a server infrastructure header only present on AMS dispatcher nodes —
+  // give it higher weight than the MD5 fingerprint pattern to overcome accumulated
+  // CS-looking signals (cmp- classes, lc- format, experience-fragments) on AMS 6.5 sites
+  if (/^dispatcher[0-9]/i.test(headers['x-dispatcher'])) {
+    amsMatches += 6;
+  }
+
   for (const pattern of aemHeadlessPatterns) {
     if (pattern.test(normalizedHtml)) {
       aemHeadlessMatches += 1;
@@ -172,14 +179,17 @@ export function detectAEMVersion(htmlSource, headers = {}) {
     amsMatches += 5; // Increased weight since this is a very reliable AMS indicator
   }
 
-  // Give extra weight to CS clientlib format pattern as it's very distinctive
+  // lc-[timestamp]-lc format exists on AEM 6.5 (AMS) too — not CS-exclusive
   if (/\/etc\.clientlibs\/[^"']+\.lc-[a-f0-9]+-lc\.min\.(js|css)/i.test(normalizedHtml)) {
-    csMatches += 3;
+    csMatches += 1;
   }
 
   // Give significant weight to explicit RUM data-routing indicators
+  // ams= gets higher weight (+8 total) because AMS 6.5 sites running Core Components
+  // accumulate many CS-looking patterns (lc- format, cmp- classes, experience-fragments)
+  // that can overtake a lower-weighted explicit AMS declaration
   if (/data-routing="[^"]*ams=([^,"]*)/i.test(normalizedHtml)) {
-    amsMatches += 5;
+    amsMatches += 7;
   }
 
   if (/data-routing="[^"]*eds=([^,"]*)/i.test(normalizedHtml)) {
--- a/packages/spacecat-shared-utils/src/aem.js
+++ b/packages/spacecat-shared-utils/src/aem.js
@@ -102,8 +102,10 @@ export function detectAEMVersion(htmlSource, headers = {}) {
     /data-routing="[^"]*ams=([^,"]*)/i,
   ];
 
+  // Case-insensitive: HTTP header values may arrive with unexpected casing.
+  // Must stay aligned with the extra-weight check below.
   const amsHeaderPatterns = [
-    /^dispatcher[0-9].*$/,
+    /^dispatcher[0-9]/i,
   ];
 
   const aemHeadlessPatterns = [
--- a/packages/spacecat-shared-utils/test/aem.test.js
+++ b/packages/spacecat-shared-utils/test/aem.test.js
@@ -241,6 +241,26 @@ describe('AEM Detection', () => {
         expect(result).to.equal(DELIVERY_TYPES.AEM_AMS);
       });
 
+      it('should detect AMS with mixed-case x-dispatcher header value', () => {
+        // HTTP header values may arrive with unexpected casing — the pattern and the
+        // extra-weight check must both match case-insensitively to score consistently.
+        const headers = { 'x-dispatcher': 'Dispatcher3eusouth2' };
+        const htmlSource = `
+          <html>
+            <head>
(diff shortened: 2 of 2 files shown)


packages/spacecat-shared-utils/src/aem.js
```diff
@@ -145,6 +145,13 @@ export function detectAEMVersion(htmlSource, headers = {}) {
     }
   }
 
+  // x-dispatcher is a server infrastructure header only present on AMS dispatcher nodes —
+  // give it higher weight than the MD5 fingerprint pattern to overcome accumulated
+  // CS-looking signals (cmp- classes, lc- format, experience-fragments) on AMS 6.5 sites
+  if (/^dispatcher[0-9]/i.test(headers['x-dispatcher'])) {
```
MysticatBot, line 153: issue (blocking): The existing `amsHeaderPatterns` entry for x-dispatcher uses `/^dispatcher[0-9].*$/` (case-sensitive, end-anchored). This new extra weight check uses `/^dispatcher[0-9]/i` (case-insensitive, no end anchor).

This means a header value like `Dispatcher3eusouth2` (capital D) matches only the new block (+6) but not the loop (+0), giving +6 total instead of +7. Conversely, a value with unexpected trailing content could match one but not the other due to the end-anchor difference. In practice AMS dispatcher values are lowercase, but the inconsistency creates a maintenance trap.

Fix: align both regexes. Recommended approach - update `amsHeaderPatterns` to `/^dispatcher[0-9]/i` (matching the new block), since HTTP header values may arrive with unexpected casing and the case-insensitive match is the more defensive choice.

MysticatBot (changes requested):
Hey @Kanishkavijay39,

**Verdict: Request changes** - one inconsistency in the new x-dispatcher check needs alignment before merge.
_Changes: recalibrates AEM version detection weights so AMS 6.5 sites running Core Components are no longer misdetected as aem_cs (2 files)._

## Must fix before merge
1. [Important] Case-sensitivity mismatch between `amsHeaderPatterns` and new extra weight check - `packages/spacecat-shared-utils/src/aem.js:151` (details inline)

<details>

- suggestion: Add a test case with mixed-case `x-dispatcher` value (e.g., `Dispatcher3eusouth2`) to lock in intended behavior and prevent regression if someone aligns the regexes in one direction or the other - `packages/spacecat-shared-utils/test/aem.test.js:220`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 6m 53s | <b>Cost:</b> $3.88 | <b>Commit:</b> <code>17623729df4e4a20b40f684c9b50db83a25141bf</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-utils/src/aem.js
Kanishkavijay39: Fixed in f979386c. Aligned `amsHeaderPatterns` to `/^dispatcher[0-9]/i` (case-insensitive, dropped the redundant `.*$`) so it matches the extra-weight check. Added a mixed-case regression test (`Dispatcher3eusouth2`) to lock in the behavior. Both the +1 loop and +6 block now fire consistently for any casing → intended +7 total.

MysticatBot (approved):
Hey @Kanishkavijay39,

**Verdict: Approve** - focused bug fix with clean weight recalibrations and thorough regression tests.
_Changes: recalibrates AEM version detection weights so AMS 6.5 sites running Core Components are no longer misdetected as aem_cs (2 files)._

<details>

- nit: The x-dispatcher scoring is split across the `amsHeaderPatterns` loop (+1) and standalone block (+6) with duplicated regex literals - consolidating into a single `amsMatches += 7` would remove the maintenance trap of keeping two identical patterns in sync - `packages/spacecat-shared-utils/src/aem.js:150`
- suggestion: Add a test with `x-dispatcher` and otherwise neutral HTML (no cmp- classes) to document the minimum signal threshold needed for the header alone to classify as AMS - `packages/spacecat-shared-utils/test/aem.test.js:220`
</details>

- Case-sensitivity mismatch between `amsHeaderPatterns` and the extra-weight check - aligned to `/^dispatcher[0-9]/i` with mixed-case regression test added

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 7s | <b>Cost:</b> $3.96 | <b>Commit:</b> <code>f979386c2ff15e5a9f88c00cfb3c34dcd394fd75</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rpapani (approved):

PR #1689, opened 2026-06-17
feat(ims-client): include IMS response body in token-mint error messages

GeezerPelletier:
## Summary

The three token-mint paths (`getServiceAccessToken` v2, `getServiceAccessTokenV3`, `getServicePrincipalAccessToken`) currently throw `IMS X request failed with status: N` with no detail from the IMS response body. That swallows the load-bearing diagnostic info — IMS rejections include OAuth2-shape error payloads like `{ error: "invalid_scope", error_description: "..." }` that disambiguate between *wrong scope value*, *client not registered for grant*, *bad credentials*, etc. The status code alone is the same 400 for all of these.

## Motivating case

[SITES-43236](https://jira.corp.adobe.com/browse/SITES-43236) spacecat-api-service debugging hit `IMS getServiceAccessTokenV3 request failed with status: 400` in Splunk after a code change ([spacecat-api-service#2611](https://github.com/adobe/spacecat-api-service/pull/2611) merged) that started using v3 client_credentials against the `aem-project-collab-service` IMS client. The actual cause (likely scope authorization gap on that client) could only be guessed without the response body — we burned cycles trying `IMS_SCOPE='system'` ([spacecat-api-service#2627](https://github.com/adobe/spacecat-api-service/pull/2627)) before having concrete error info from IMS.

The next person hitting an IMS auth issue on any of these three endpoints should not have to go through the same guessing loop.

## Implementation

Mirrors the existing pattern in `getAccountCluster` (lines 430-441 of the same file): parse the response body as JSON, prefer the `error` field over `message`, fall back to status-only on non-JSON / unparseable bodies. Same shape, same error-format suffix (`" - <detail>"`), zero new abstractions.

```js
if (!tokenResponse.ok) {
  let errorMessage = `IMS getServiceAccessTokenV3 request failed with status: ${tokenResponse.status}`;
  try {
    const errorBody = await tokenResponse.json();
    if (hasText(errorBody.error)) {
      errorMessage += ` - ${errorBody.error}`;
    } else if (hasText(errorBody.message)) {
      errorMessage += ` - ${errorBody.message}`;
    }
  } catch (e) {
    // Response body is not JSON or cannot be parsed, ignore
  }
  throw new Error(errorMessage);
}
```

## Scope

- **In scope:** the three S2S token-mint methods that surface authentication errors (`getServiceAccessToken`, `getServiceAccessTokenV3`, `getServicePrincipalAccessToken`).
- **Out of scope:** the remaining 11 throw sites in the same file follow the same status-only pattern but aren't in the auth/mint hot path. Left untouched until they actually surface in production (YAGNI). The helper pattern is now applied at 4 call sites in the file (the original `getAccountCluster` + the three new ones), so a future refactor could extract a shared helper if the pattern proliferates further.

## Tests

- 8 new tests covering all branches per method:
  - Error field present in response body → included in thrown message
  - Message field present (no error field) → included in thrown message  
  - Both fields present → prefers `error`
  - Non-JSON / unparseable body → falls back to status-only
- All branches × all three methods covered.
- 100% line / 100% branch coverage on `ims-client.js` maintained (per `.nycrc.json`).

## Backward compatibility

Fully backward-compatible. Callers that string-match on the existing error prefix (`'IMS X request failed with status: N'`) continue to match — the new detail suffix is additive. Callers parsing the error message structurally aren't affected since the format is documented for human reading, not structured consumption.

## Test plan

- [x] `npm test -w packages/spacecat-shared-ims-client` — 89/89 pass
- [x] Coverage thresholds maintained
- [ ] Once published, dependent repos pick up the new error detail automatically. Specifically: spacecat-api-service's existing SITES-43236 debugging gets immediate diagnostic value the next time the IMS 400 surfaces.

### DIFF
--- a/packages/spacecat-shared-ims-client/src/clients/ims-client.js
+++ b/packages/spacecat-shared-ims-client/src/clients/ims-client.js
@@ -181,10 +181,14 @@ export default class ImsClient extends ImsBaseClient {
       let errorMessage = `IMS getServiceAccessToken request failed with status: ${tokenResponse.status}`;
       try {
         const errorBody = await tokenResponse.json();
+        // Cap appended detail at 200 chars — defense-in-depth against
+        // unexpectedly verbose IMS responses; typical OAuth2 error codes
+        // (`invalid_scope`, `unauthorized_client`, etc.) are well under
+        // this. Per @solaris007 review.
         if (hasText(errorBody.error)) {
-          errorMessage += ` - ${errorBody.error}`;
+          errorMessage += ` - ${errorBody.error.slice(0, 200)}`;
         } else if (hasText(errorBody.message)) {
-          errorMessage += ` - ${errorBody.message}`;
+          errorMessage += ` - ${errorBody.message.slice(0, 200)}`;
         }
       } catch (e) {
         // Response body is not JSON or cannot be parsed, ignore
@@ -225,10 +229,14 @@ export default class ImsClient extends ImsBaseClient {
       let errorMessage = `IMS getServiceAccessTokenV3 request failed with status: ${tokenResponse.status}`;
       try {
         const errorBody = await tokenResponse.json();
+        // Cap appended detail at 200 chars — defense-in-depth against
+        // unexpectedly verbose IMS responses; typical OAuth2 error codes
+        // (`invalid_scope`, `unauthorized_client`, etc.) are well under
+        // this. Per @solaris007 review.
         if (hasText(errorBody.error)) {
-          errorMessage += ` - ${errorBody.error}`;
+          errorMessage += ` - ${errorBody.error.slice(0, 200)}`;
         } else if (hasText(errorBody.message)) {
-          errorMessage += ` - ${errorBody.message}`;
+          errorMessage += ` - ${errorBody.message.slice(0, 200)}`;
         }
       } catch (e) {
         // Response body is not JSON or cannot be parsed, ignore
@@ -270,10 +278,14 @@ export default class ImsClient extends ImsBaseClient {
       let errorMessage = `IMS getServicePrincipalAccessToken request failed with status: ${tokenResponse.status}`;
       try {
         const errorBody = await tokenResponse.json();
+        // Cap appended detail at 200 chars — defense-in-depth against
+        // unexpectedly verbose IMS responses; typical OAuth2 error codes
+        // (`invalid_scope`, `unauthorized_client`, etc.) are well under
+        // this. Per @solaris007 review.
         if (hasText(errorBody.error)) {
-          errorMessage += ` - ${errorBody.error}`;
+          errorMessage += ` - ${errorBody.error.slice(0, 200)}`;
         } else if (hasText(errorBody.message)) {
-          errorMessage += ` - ${errorBody.message}`;
+          errorMessage += ` - ${errorBody.message.slice(0, 200)}`;
         }
       } catch (e) {
         // Response body is not JSON or cannot be parsed, ignore
@@ -463,10 +475,14 @@ export default class ImsClient extends ImsBaseClient {
       let errorMessage = `IMS getAccountCluster request failed with status: ${accountClusterResponse.status}`;
       try {
         const errorBody = await accountClusterResponse.json();
+        // Cap appended detail at 200 chars — defense-in-depth against
+        // unexpectedly verbose IMS responses; typical OAuth2 error codes
+        // (`invalid_scope`, `unauthorized_client`, etc.) are well under
+        // this. Per @solaris007 review.
         if (hasText(errorBody.error)) {
-          errorMessage += ` - ${errorBody.error}`;
+          errorMessage += ` - ${errorBody.error.slice(0, 200)}`;
         } else if (hasText(errorBody.message)) {
-          errorMessage += ` - ${errorBody.message}`;
+          errorMessage += ` - ${errorBody.message.slice(0, 200)}`;
         }
       } catch (e) {
         // Response body is not JSON or cannot be parsed, ignore
--- a/packages/spacecat-shared-ims-client/src/clients/ims-client.js
+++ b/packages/spacecat-shared-ims-client/src/clients/ims-client.js
@@ -178,7 +178,18 @@ export default class ImsClient extends ImsBaseClient {
     );
 
     if (!tokenResponse.ok) {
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-ims-client/src/clients/ims-client.js
```diff
@@ -178,7 +178,18 @@ export default class ImsClient extends ImsBaseClient {
...
+      try {
+        const errorBody = await tokenResponse.json();
+        if (hasText(errorBody.error)) {
+          errorMessage += ` - ${errorBody.error}`;
+        } else if (hasText(errorBody.message)) {
+          errorMessage += ` - ${errorBody.message}`;
+        }
+      } catch (e) {
+        // Response body is not JSON or cannot be parsed, ignore
+      }
```
MysticatBot, line 195: issue (blocking): The same 11-line try/catch/hasText block is now copy-pasted three times in this file (four counting `getAccountCluster` at lines 430-441). This is a maintenance trap: when the extraction logic needs to evolve (e.g., adding `error_description`, truncating long bodies), all copies must change in lockstep.

The PR description says "zero new abstractions" - understood, but a private helper is not a new abstraction, it is DRY on code this PR introduces. A minimal extraction:

```js
async #extractErrorDetail(response) {
  try {
    const body = await response.json();
    if (hasText(body.error)) return ` - ${body.error}`;
    if (hasText(body.message)) return ` - ${body.message}`;
  } catch {
    // not JSON, ignore
  }
  return '';
}
```

Then each call site becomes:

```js
throw new Error(`IMS getServiceAccessToken request failed with status: ${tokenResponse.status}${await this.#extractErrorDetail(tokenResponse)}`);
```

The existing `getAccountCluster` site is out of scope for this PR but becomes a natural follow-up candidate.

packages/spacecat-shared-ims-client/test/clients/ims-client.test.js
```diff
@@ -152,6 +152,31 @@ describe('ImsClient', () => {
...
+
+    it('v2 mint failure includes message field from response body when error field absent', async () => {
+      nock(`https://${DUMMY_HOST}`)
+        .post('/ims/token/v4')
+        .query(true)
+        .reply(401, {
+          message: 'Client credentials rejected',
+        });
+
+      await expect(client.getImsOrganizationDetails('123456@AdobeOrg'))
```
MysticatBot, line 176: nit: v3 and service-principal each have 4 error-body tests (error field, message field, both-prefers-error, non-JSON fallback). The v2 path has only 2 (error field, message field). The existing `.reply(500)` test implicitly hits the non-JSON catch path, but the "both fields prefers error" case has no v2 coverage at all. Adding the missing two for symmetry is low-urgency since all three methods share identical logic.

MysticatBot (changes requested):
Hey @GeezerPelletier,

**Verdict: Request changes** - one structural concern on the tripled logic block.
_Changes: Enriches error messages in three IMS token-mint methods to include the response body's `error` or `message` field for faster on-call debugging (2 files)._

## Must fix before merge
1. [Important] Identical 11-line error-extraction block repeated 3 times (4 with `getAccountCluster`) - `packages/spacecat-shared-ims-client/src/clients/ims-client.js:180` (details inline)

<details>

- nit: v2 tests cover 2 of 4 branches while v3 and service-principal cover all 4 (missing "both fields prefers error" and explicit non-JSON fallback for v2) - `packages/spacecat-shared-ims-client/test/clients/ims-client.test.js:155`
- suggestion: Consider appending `error_description` alongside `error` for richer on-call context (e.g., `invalid_grant: Authorization code expired` instead of just `invalid_grant`)
- suggestion: Consider truncating the appended detail to a reasonable max length (e.g., 500 chars) as a defensive measure against unexpectedly verbose upstream responses
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 51s | <b>Cost:</b> $3.31 | <b>Commit:</b> <code>2240df84c9535ebee978c4f4beace460665c527c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

solaris007 (approved):
Hey @GeezerPelletier,

**Verdict: Approve** - clean, well-scoped diagnostic improvement; no secret-leak risk, only optional polish.
_Changes: Enriches three IMS token-mint methods to append the response body's `error`/`message` to the thrown error for faster on-call debugging (2 files)._

Security check (the obvious concern with "include the response body"): no leak. The body is parsed only on `!ok`, where no token exists; the success-path token read is separate and unchanged. Request credentials live in the request, never the response. Appending `error`/`message` while omitting `error_description` is the correct safe choice - keep it that way.

<details>
<summary>Non-blocking (3): minor suggestions</summary>

- suggestion: I read the "identical 11-line block repeated 3x" point as non-blocking - the 4-call-site YAGNI tradeoff is justified in the PR body, and the block is small, identical, and fully tested. Extract a shared helper only if a 5th site appears.
- nit: test-branch asymmetry - v2 covers 2 of 4 branches while v3 and service-principal cover all 4. Add the "both fields prefers error" and non-JSON fallback cases for v2 to match - `packages/spacecat-shared-ims-client/test/clients/ims-client.test.js:155`
- suggestion: optional defense-in-depth - cap the appended detail length (e.g. first ~200 chars) to bound any future IMS-side surprise in `message`. Do not add `error_description` for "richer context" - that field reflects input and is the one worth keeping out.
</details>

MysticatBot (approved):
Hey @GeezerPelletier,

**Verdict: Approve** - prior feedback addressed cleanly, no new concerns.
_Changes: Enriches error messages in three IMS token-mint methods to include the response body's `error` or `message` field (capped at 200 chars) for faster on-call debugging (2 files)._

- 200-char truncation cap applied consistently across all four error-body extraction sites (three new + existing `getAccountCluster`)
- v2 test branch coverage now matches v3 and service-principal (all four branches covered per method)
- Truncation regression test added on the v3 path to pin the 200-char bound

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 42s | <b>Cost:</b> $3.36 | <b>Commit:</b> <code>232e44258f55eb108a3ab2352beb0838584d6635</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1694, opened 2026-06-19
fix: sync package-lock.json with released workspace versions

rainer-friederich:
## Problem

`npm ci` fails on every branch off `main` during CI setup:

```
npm error `npm ci` can only install packages when your package.json and
package-lock.json or npm-shrinkwrap.json are in sync.
npm error Missing: @adobe/spacecat-shared-ims-client@1.13.3 from lock file
```

This blocks the `Test` job (and therefore release) on unrelated PRs — it is a pre-existing drift on `main`, not introduced by any single feature branch.

## Root cause

- `spacecat-shared-cloud-manager-client` pins `@adobe/spacecat-shared-ims-client` at the **exact** version `1.13.3` (introduced in PR 1687 to drop the electrodb/jsonschema transitive deps). While the workspace `ims-client` package was itself `1.13.3`, the lockfile recorded that dependency as a workspace link and `npm ci` was satisfied.
- The `ims-client` `1.14.0` release (PR 1689) then moved the workspace package past the exact pin. semantic-release bumps `package.json` + `CHANGELOG` but does **not** run `npm install`, so the root `package-lock.json` was never regenerated: no registry node for `1.13.3` was added, and several recorded workspace `version` fields went stale relative to the `package.json` values `main` already ships.

`brand-client` / `gpt-client` pin `1.11.10` and stay green only because the lock already contains registry nodes for that version.

## Fix

Regenerate the lockfile only (`npm install --package-lock-only`):

- Adds the nested registry resolution for `@adobe/spacecat-shared-ims-client@1.13.3` under `cloud-manager-client/node_modules`, satisfying its exact pin.
- Syncs the recorded workspace `version` fields (`data-access`, `ims-client`, `utils`, `cloud-manager-client`) to the values already in each package's `package.json` on `main`.

No `package.json` or source changes — lockfile-only.

## Follow-up (not in this PR)

The deeper trigger is exact-pinning an internal workspace dependency (`cloud-manager-client` → `ims-client: "1.13.3"`): it re-breaks the lock every time `ims-client` releases. A separate change could move it to the workspace/caret range so releases don't leave the lock stale.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25942,7 +25942,7 @@
     },
     "packages/spacecat-shared-cloud-manager-client": {
       "name": "@adobe/spacecat-shared-cloud-manager-client",
-      "version": "1.4.3",
+      "version": "1.4.4",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "4.3.0",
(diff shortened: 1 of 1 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean lockfile resync with a clear root-cause explanation and correct minimal fix.
_Changes: regenerates package-lock.json to add the nested registry resolution for @adobe/spacecat-shared-ims-client@1.13.3 and sync stale workspace version fields (1 file)._

No issues found. The resolved URL points to the expected public npm registry under the @adobe scope, the integrity hash is present and well-formed, and no new dependencies are introduced. The PR description thoroughly documents the root cause and the follow-up work (bumping the exact pin in cloud-manager-client).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 46s | <b>Cost:</b> $1.68 | <b>Commit:</b> <code>d6fe3d972d528762bb81e1871b12f157d4ea56e0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1695, opened 2026-06-19
ci: prevent and detect package-lock.json drift on release

rainer-friederich:
## Why

PR #1694 hot-fixed a `main`-wide outage: every branch failed `npm ci` with `Missing @adobe/spacecat-shared-ims-client@1.13.3 from lock file`. This PR fixes the **mechanism** so it cannot recur.

### Root cause

semantic-release bumps each package's `package.json` + `CHANGELOG` on release, but the `@semantic-release/git` `assets` list in every `.releaserc.cjs` is `["package.json", "CHANGELOG.md"]` — it **never regenerates the root `package-lock.json`**. So the lock drifts on every release. When a released workspace version later crosses a consumer's **exact internal pin** (e.g. `cloud-manager-client` → `spacecat-shared-ims-client: "1.13.3"`, satisfied by a workspace link until `ims-client` released `1.14.0`), the lock loses that resolution and `npm ci` breaks on every branch — first surfacing on unrelated PRs.

Internal deps are **uniformly exact-pinned** across the monorepo (no carets anywhere), so de-pinning would be a convention change. This PR instead keeps the lock authoritative, preserving the exact-pin convention.

## What

Two complementary layers in `.github/workflows/main.yaml`:

- **Prevention — release job (`Sync package-lock.json after release`):** after `Semantic Release`, regenerate the root lock against the just-published versions and, if it changed, commit it back to `main` with a skip-ci marker (GitHub natively skips workflow runs for such commits — no new CI/release). `continue-on-error: true` so it can **never** fail a release; the packages are already published and the detection gate is the backstop.
- **Detection — test job (`Verify package-lock.json is in sync`):** regenerate the lock with the same pinned npm (`11.13.0`, from the `setup-node-npm` composite) and fail if it differs, so residual drift is caught on the PR that introduces it instead of silently breaking `main`.

Plus `RELEASE-RUNBOOK.md` **Failure mode 7** documenting recovery if the sync step ever fails (a lockfile-only PR, exactly what #1694 did).

## Safety / layering

- On a healthy `main` the detection regeneration is a **no-op** (verified locally: `npm install --package-lock-only` on current `main` produces no diff, post-#1694).
- npm/node versions match the repo pins (`11.13.0` / `24.16.0`), so no lockfile-format false-diffs.
- If the release auto-sync loses a push race (serialized concurrency makes this unlikely; one rebase-retry is built in), it is `continue-on-error`, and the next PR's detection gate surfaces the residual drift → manual lockfile-only PR per Failure mode 7.

## Reviewer note (SRE)

The prevention step **writes to `main` from CI** (mirrors how semantic-release already pushes its release commits with `ADOBE_BOT_GITHUB_TOKEN`). If you'd prefer not to auto-commit, the **detection gate stands on its own** — drop the `Sync package-lock.json after release` step and recovery is the manual lockfile PR (Failure mode 7) whenever the gate goes red after a release.

> Note: the prose here intentionally avoids the literal bracketed skip-ci token so a squash-merge message derived from it does not skip the post-merge Build.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -271,13 +271,28 @@ jobs:
           # skips this commit (belt-and-suspenders alongside the native [skip ci]
           # skip), matching how that guard is documented.
           git commit -m "chore(release): sync package-lock.json after release [skip ci]"
-          remote="https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
+          # Authenticate via http.extraHeader (base64 Basic auth) rather than
+          # embedding the token in the remote URL. git echoes the remote URL in
+          # push/pull error output, so a URL-embedded token can leak into logs if
+          # Actions' secret masking misses the exact substring; the header value
+          # is never echoed. Mirrors how actions/checkout injects credentials.
+          # Pre-mask the base64 form (which is NOT auto-masked, being a distinct
+          # string from the raw secret) before it can reach any log line.
+          auth_b64="$(printf 'x-access-token:%s' "${GH_TOKEN}" | base64 | tr -d '\n')"
+          echo "::add-mask::${auth_b64}"
+          remote="https://github.com/${GITHUB_REPOSITORY}.git"
+          git_auth() { git -c "http.extraheader=AUTHORIZATION: basic ${auth_b64}" "$@"; }
           # Releases are serialized (concurrency: npm-publish-main), so a race is
-          # unlikely, but retry once against the latest main just in case.
-          if ! git push "${remote}" HEAD:main; then
+          # unlikely, but retry once against the latest main just in case. KNOWN
+          # LIMITATION: a SECOND concurrent push landing within this retry window
+          # still loses — by design we do not loop. continue-on-error keeps it
+          # from failing the release, the ERR trap surfaces a ::warning::, and the
+          # residual drift is recovered via the next PR's drift guard +
+          # RELEASE-RUNBOOK Failure mode 7 (the manual lockfile-only PR).
+          if ! git_auth push "${remote}" HEAD:main; then
             echo "push rejected; rebasing on latest main and retrying once."
-            git pull --rebase "${remote}" main
-            git push "${remote}" HEAD:main
+            git_auth pull --rebase "${remote}" main
+            git_auth push "${remote}" HEAD:main
           fi
           echo "Pushed package-lock.json sync to main."
 
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -248,7 +248,16 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.ADOBE_BOT_GITHUB_TOKEN }}
         run: |
-          set -uo pipefail
+          set -euo pipefail
+          # continue-on-error keeps this step from failing a release, which also
+          # means a failure is otherwise invisible. Surface ANY failure (regen,
+          # commit, or the push+rebase retry) as a workflow ::warning:: + run
+          # summary line so the operator sees it without digging through logs;
+          # the next PR's drift guard is still the backstop (RELEASE-RUNBOOK
+          # Failure mode 7). errexit makes a failed `npm install` an honest red
+          # mark instead of a false "already in sync" — the `if ! git push`
+          # retry below is errexit-safe (the negated condition suppresses it).
+          trap 'echo "::warning title=package-lock sync after release failed::Lockfile sync did not complete; see RELEASE-RUNBOOK Failure mode 7. The next PR drift guard will surface residual drift."; echo "package-lock.json sync after release failed — see RELEASE-RUNBOOK Failure mode 7." >> "$GITHUB_STEP_SUMMARY"' ERR
           npm install --package-lock-only --ignore-scripts
           if git diff --quiet -- package-lock.json; then
             echo "OK: package-lock.json already in sync; nothing to commit."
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yaml
```diff
@@ -201,6 +229,46 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.ADOBE_BOT_GITHUB_TOKEN }}
 
+      - name: Sync package-lock.json after release
+        # semantic-release bumps the released package versions but does not
+        # regenerate the root package-lock.json (see the Test job's drift
+        # guard). Left unsynced, the lock's resolutions for internal exact pins
+        # go stale the moment a released workspace version crosses a consumer's
+        # pin, breaking `npm ci` on every branch. Regenerate the lock against
```
MysticatBot, line 237: issue (blocking): The sync step uses `set -uo pipefail` (no `-e`), while the detection step uses `set -euo pipefail`. Without `-e`, if `npm install --package-lock-only` fails (registry timeout, corrupt cache, network issue), the script continues silently. `git diff --quiet` then finds no diff (npm never modified the file), and the step exits with "OK: package-lock.json already in sync" - a false negative that masks a failure to sync.

Adding `-e` is safe here: the `if ! git push` retry pattern is errexit-safe (the `if` suppresses errexit on the negated command), so all existing control flow still works. With `-e`, a failed `npm install` exits the script, the step reports failure, and `continue-on-error: true` absorbs it at the job level - producing an honest red mark instead of a misleading green check.

**Fix:** Change `set -uo pipefail` to `set -euo pipefail`.

MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - one low-effort fix to close a silent false-negative in the sync step.
_Changes: Adds a two-layer lockfile drift guard to CI - a prevention step that syncs package-lock.json after release, and a detection step that fails PRs when the lockfile is out of sync (2 files)._

## Must fix before merge
1. [Important] Missing `set -e` allows silent false-negative on npm failure - `.github/workflows/main.yaml` sync step (details inline)

<details>

- suggestion: Align sync commit prefix to `chore(release):` so the existing release-job skip guard catches it as belt-and-suspenders alongside `[skip ci]` - `.github/workflows/main.yaml`
- suggestion: Add a `::warning::` annotation when the sync step fails its retry, so operators see the failure in the workflow summary without inspecting logs - `.github/workflows/main.yaml`
- nit: Bot commit email uses a third-party domain (`semantic-release-bot@martynus.net`) - consider an org-owned noreply address for auditability - `.github/workflows/main.yaml`
- nit: Failure mode 7 in the runbook could include detection-timing and owner fields to match the pattern established by earlier failure modes - `docs/RELEASE-RUNBOOK.md`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 51s | <b>Cost:</b> $3.63 | <b>Commit:</b> <code>e0bece44de44f7249a08e012380664fa04c980b4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

solaris007 (approved):
🙌🤌👍

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior blocking finding resolved, no new issues.
_Changes: Adds a two-layer lockfile drift guard to CI - a prevention step that syncs package-lock.json after release, and a detection step that fails PRs when the lockfile is out of sync (2 files)._

<details>

- nit: Token embedded in git remote URL could appear in verbose error output if log masking misses it - consider `git -c http.extraheader` or a credential-helper approach for defense-in-depth - `.github/workflows/main.yaml:254`
- nit: Single-retry on push race means a second concurrent push causes silent failure - mitigated by `continue-on-error` + FM-7 recovery, so no action needed - `.github/workflows/main.yaml:258`
</details>

- `set -euo pipefail` now guards both shell steps (no more silent false-negative on npm failure)
- Sync commit prefix aligned to `chore(release):` for belt-and-suspenders skip-guard
- ERR trap added with `::warning::` annotation and `$GITHUB_STEP_SUMMARY` line for operator visibility

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 53s | <b>Cost:</b> $3.09 | <b>Commit:</b> <code>9c21680f37531b65139eb8d7995b814cde36cbbb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior nits resolved, no new issues.
_Changes: Adds a two-layer lockfile drift guard to CI - a prevention step that syncs package-lock.json after release, and a detection step that fails PRs when the lockfile is out of sync (2 files)._

<details>
<summary>Non-blocking (0): no issues</summary>

No findings. The incremental changes cleanly address both prior nits: token is now injected via `http.extraHeader` with `::add-mask::` on the base64 form (matching `actions/checkout` internals), and the single-retry push-race limitation is documented in RELEASE-RUNBOOK Failure mode 7.
</details>

- Token no longer embedded in remote URL - uses `http.extraheader` with pre-masked base64 auth
- Single-retry push-race limitation documented as known in RELEASE-RUNBOOK FM-7

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 37s | <b>Cost:</b> $3.37 | <b>Commit:</b> <code>4317bd193ccc730214fa5d4b899be058af05ebd8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1705, opened 2026-06-23
feat(user-manager-client): add typed client wrapper + IMS Bearer auth + retry; vendor package

rainer-friederich:
## 1. Abstract
Adds the typed client wrapper `createSerenityUserManagerApiClient` to `@adobe/spacecat-shared-user-manager-client`, and vendors the package (removes `private`) so services can depend on it.

## 2. Reasoning
The user-manager-client package already shipped the Semrush User Manager spec, generation pipeline, spec-correction overlay, generated TS/Pydantic types, and a Counterfact mock — but no actual client. The sibling Project Engine client (LLMO-5461) established the wrapper pattern for exactly this kind of Semrush integration; this PR ports that pattern onto the User Manager API so callers have a thin, typed, auth- and retry-aware client instead of hand-rolling `fetch` against the generated `paths`.

- New `createSerenityUserManagerApiClient(options)` — an `openapi-fetch` client fully typed over the generated `paths`. It owns the fixed `/enterprise/users/api` prefix (callers pass only an origin), normalises the base URL to `protocol//host` (dropping any leftover path or embedded credentials), and fails fast on a non-http(s) or malformed base URL.
- Auth: forwards the caller's IMS JWT verbatim as `Authorization: Bearer <token>` — no minting or exchange, since Semrush accepts the IMS bearer directly. Accepts either a static token or a (sync/async) getter resolved once per logical request, so the header reflects the current caller. Fails fast on an empty token rather than sending `Bearer undefined`, and propagates a throwing getter (e.g. IMS outage).
- Retry: `429` retried for any method; retryable `5xx` and network errors retried only for idempotent methods, with jittered exponential backoff that honours `Retry-After` (clamped to a ceiling). Bodied requests are cloned per attempt; an optional best-effort `onRetry` hook is available for observability and can never break the request.
- Package made publishable: `private` removed and `main`/`types`/`exports`/`files` added, so the `@adobe`-scoped package is released to public npm by semantic-release and consumers depend on it as a normal npm dependency. This is a **deliberate divergence** from the sibling Project Engine client, which is kept `private: true` and vendored (copied) into `spacecat-api-service/third-party/` rather than published. Adds the `openapi-fetch` runtime dependency.
- The auth model needed no new spec work — the package's existing overlay correction already removes the legacy `Auth-Data-Jwt` header and expresses `Authorization: Bearer`, so the generated types already match the client's behaviour.

## 4. Required information
- Jira / issue: LLMO-5558 — https://jira.corp.adobe.com/browse/LLMO-5558

- spacecat-api-service — intended future runtime consumer (contract). No code change here; wiring the client into the api-service transport is deferred (tracked separately). Unlike the Project Engine client (vendored into `third-party/`), this package will be consumed as a published npm dependency.

## 7. Test plan
- (a) Local: the client is covered by unit suites that stub `fetch` and assert the auth header, base-URL normalisation (origin + `/enterprise/users/api`), credential stripping, retry-on-5xx, token-getter resolution, and fail-fast paths. For an integration check, run the bundled Counterfact mock (`npm run mock`, serves `:4010` under `/enterprise/users/api`) and point the client at `http://localhost:4010`, e.g. `GET /v1/countries`.
- (b) Per-env: this is a shared library with no standalone deploy. It takes effect only when a consuming service bumps the dependency; verify there at that time. No environment-specific action is required for this PR on its own.


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-structured greenfield client with thorough tests and careful error handling; no blocking issues.
_Changes: adds a typed openapi-fetch client wrapper for the Semrush User Manager API with IMS Bearer auth, jittered exponential-backoff retry, and base-URL normalization, and makes the package publishable (11 files)._

<details>

- nit: misleading "idempotent prefix" comment in test - `test/client.test.js:46` says "a base URL that already carries the prefix normalises to the same target (idempotent)" but the code simply strips all paths; the test proves path-stripping, not prefix deduplication. Rename or remove the comment.
- suggestion: update CLAUDE.md "API Client Pattern" section to acknowledge the openapi-fetch wrapper variant - two packages (`project-engine-client`, `user-manager-client`) now diverge from the documented class+factory pattern. A one-line note ("openapi-fetch-based clients use a thin factory wrapper instead") prevents confusion for future contributors.
- nit: 429 retry is method-agnostic (including POST) - the JSDoc at `src/internal.js:64-78` documents the assumption that Semrush rate-limits at the edge before processing writes. Consider adding an opt-in `retryPostOn429` flag (default true) so future write-heavy callers can disable it without forking the client.
- suggestion: add tests for missing-options error paths - `createSerenityUserManagerApiClient()` (no arg), `createSerenityUserManagerApiClient({})` (missing baseUrl), and `{ baseUrl: 'https://x' }` (missing authToken) are real misconfiguration paths that should produce clear diagnostics and be documented via tests.
</details>

Out of scope, worth tracking: `src/internal.js` is near-identical to `spacecat-shared-project-engine-client/src/internal.js` (retry, jitter, Retry-After parsing, token-getter, auth middleware). Before a third Semrush client arrives, extract the shared retry/auth layer into a common package to avoid O(n) maintenance of identical copies.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 54s | <b>Cost:</b> $5.24 | <b>Commit:</b> <code>a6ef68535cde4fba25d1a59b36ff824244ef2c30</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior feedback addressed cleanly, no new issues introduced.
_Changes: addresses prior review feedback with a test comment clarification, two new misconfiguration-path tests, and a CLAUDE.md note documenting the openapi-fetch wrapper variant (2 files)._

- Misleading "idempotent prefix" test comment reworded to describe what `resolveBaseUrl` actually does (drops path, re-appends prefix)
- CLAUDE.md "API Client Pattern" section now acknowledges the openapi-fetch wrapper variant used by Semrush clients
- Missing-options error path tests added for `baseUrl` and `authToken` absence with clear diagnostics

The declined `retryPostOn429` flag (no current write-heavy caller, would diverge from sibling client, worth revisiting with shared-layer extraction) is reasonable - dropping it.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 38s | <b>Cost:</b> $2.02 | <b>Commit:</b> <code>5faf17febe4bfbe78a3362f87675cba13bad9b57</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1706, opened 2026-06-23
fix(project-engine-client): correct workspace-id param location (CR4) + publish to npm

rainer-friederich:
## 1. Abstract
Corrects the workspace `id` parameter location in the generated Semrush Project Engine types (overlay correction CR4) and removes the `private` flag so `@adobe/spacecat-shared-project-engine-client` publishes to npm.

## 2. Reasoning
The vendored Semrush swagger declares the workspace `id` as a query parameter on 11 operations where `{id}` is in fact a path-template variable. openapi-typescript faithfully reproduced that, so a consumer passing the runtime-correct `path: { id }` fails to type-check while still working at runtime. Separately, spacecat-api-service needs to consume this typed client from the registry to back its Semrush transport, but the package was foundation-gated with `private: true` and therefore never published, despite already carrying public publish config, a release config, and git tags to v1.1.0.

- Adds overlay correction CR4 to `spec/overlays/corrections.yaml`: relocates every `id` parameter to `in: path`. This corrects the 11 mis-located operations and is a no-op for the 87 that were already correct (verified: all 11 have `{id}` in their path template, and `id` is a workspace path identifier everywhere it appears). The generated `types.ts` is regenerated from the overlaid spec, so the affected operations now type the workspace id as a path parameter.
- Adds a foundation guard test pinning the corrected shape, so a future Semrush spec refresh that silently drops the overlay fails loudly (matching the existing CR1/CR2 guards).
- Removes `private: true` so the next release publishes the package to npm with the CR4 fix.
- Runtime behaviour is unchanged for callers — the fix is type-only; openapi-fetch already substituted the path variable correctly at runtime.

## 4. Required information
- Other: consumer adoption — https://github.com/adobe/spacecat-api-service/pull/2643

- spacecat-api-service — consumes this published typed client (contract): its Semrush Project Engine transport is backed by `createSerenityProjectEngineApiClient`, and it pins the version this release publishes.

- Verified the defect two-sided: traced the generated types back to the vendored swagger, where the workspace `id` is declared `in: query` on the affected operations, and confirmed every one of those 11 operations carries `{id}` in its path template — so the relocation to `in: path` matches both the URL contract and the live API (which the deployed transport already calls successfully with the path-substituted id).
- After applying the overlay and regenerating, confirmed the affected operations type `id` under `path` (and no longer under `query`).

## 7. Test plan
- (a) Local: the consuming spacecat-api-service Semrush transport was run against this package's source and behaves identically to the prior vendored copy — expected, since the correction is type-only and the runtime URL was already correct.
- (b) Per-env: this is a library package with no service runtime. After merge, confirm the semantic-release pipeline publishes the package to npm, then confirm spacecat-api-service can install the published version and its transport/serenity suites pass against it.

- https://github.com/adobe/spacecat-api-service/pull/2643 — related / blocks: it switches its Semrush transport from a vendored copy to this published package, pinned to the version this release publishes; its `npm install` cannot resolve that version until this publishes.
- Sequence: merge this PR → semantic-release publishes the package → update/verify the spacecat-api-service lockfile against the published version → merge the api-service PR.

### DIFF
--- a/packages/spacecat-shared-project-engine-client/package.json
+++ b/packages/spacecat-shared-project-engine-client/package.json
@@ -3,7 +3,6 @@
   "version": "1.1.0",
   "description": "Shared modules of the Spacecat Services - Semrush Project Engine client and generated types",
   "type": "module",
-  "private": true,
   "main": "src/index.js",
   "types": "src/index.d.ts",
   "exports": {
--- a/packages/spacecat-shared-project-engine-client/spec/overlays/corrections.yaml
+++ b/packages/spacecat-shared-project-engine-client/spec/overlays/corrections.yaml
@@ -109,6 +109,11 @@ actions:
   # rest. The single-equality overlay filter cannot express "name == 'id' && in ==
   # 'query'", but the blanket flip is equivalent and safe here. deep-merge sets
   # `in` to `path` (scalar overwrite); `required: true` is already present.
+  # CAUTION (for future spec-refresh reviewers): this is a BLANKET filter on the
+  # param name alone. If a future spec revision adds a legitimately query-bound
+  # parameter literally named `id` on some new operation, this overlay would
+  # silently relocate it to `in: path`. If that ever happens, narrow the filter
+  # (or special-case that operation) rather than dropping CR4.
   - target: "$.paths.*.*.parameters[?(@.name == 'id')]"
     description: "CR4: workspace `id` is a path parameter on every operation (not query)"
     update:
--- a/packages/spacecat-shared-project-engine-client/spec/overlays/corrections.yaml
+++ b/packages/spacecat-shared-project-engine-client/spec/overlays/corrections.yaml
@@ -13,6 +13,12 @@ info:
 #      Remove when Semrush fixes securityDefinitions upstream.
 # CR3  Correct GET /v2/.../projects summary+description — it also returns initial
 #      (never-yet-published) drafts, not only published projects.
+# CR4  Relocate the workspace `id` parameter from query → path. 11 operations
+#      declare `id` as `in: query` though `{id}` is a path-template variable in
+#      every one of them; openapi-typescript then emits an unsatisfiable `{id}`
+#      path key + a `query: { id }`, so a caller passing the runtime-correct
+#      `path: { id }` fails to type-check. Remove when Semrush fixes the param
+#      location upstream.
 actions:
   # ── CR1: Add GET /v1/ai_models ──────────────────────────────────────────
   # Live: 200 -> { page:int, total:int, items:[AIModelResponse] }.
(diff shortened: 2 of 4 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean, well-scoped type-only fix with appropriate regression guard and safe publish ungate.
_Changes: Corrects the workspace `id` parameter location from query to path in the generated Semrush Project Engine types via overlay CR4 and removes the `private` flag to enable npm publishing (4 files)._

<details>

- nit: The test boundary `types.indexOf('requestBody', start)` is fragile - if a future spec refresh removes the `requestBody` from the target operation, `indexOf` returns -1 and `slice(start, -1)` captures nearly the entire file, causing assertions to pass vacuously rather than fail. Adding `expect(endIdx).to.be.greaterThan(start)` before slicing would close this gap. - `test/foundation.test.js:77`
- suggestion: The blanket JSONPath filter `$.paths.*.*.parameters[?(@.name == 'id')]` forces every `id` param to `in: path`. The inline comment already documents why this is safe today (87 are already `in: path`), but if a future spec revision introduces a legitimate query param named `id` on a new operation, the overlay would silently relocate it. Consider a `// CAUTION:` annotation noting this coupling for future spec-refresh reviewers. - `spec/overlays/corrections.yaml:108`
- suggestion: Add a brief note on why `aio-create-prompts-with-tags` was chosen as the representative operation - specifically, it is the only one where `id` was the sole query param, making `query?: never` a uniquely strong regression signal. - `test/foundation.test.js:69`

</details>

Note: CI test check is currently pending.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 0s | <b>Cost:</b> $3.50 | <b>Commit:</b> <code>3a761a943cc15e932cb7388f713e6f89ced5c63e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean follow-up commit addressing all prior non-blocking items; no new concerns.
_Changes: Adds boundary guards, rationale comments, and a CAUTION annotation to the CR4 overlay and its foundation test (2 files)._

- Fragile `requestBody` slice boundary now guarded with an explicit `greaterThan(start)` assertion.
- Blanket `id` filter in `corrections.yaml` now carries a CAUTION block for future spec-refresh reviewers.
- Representative operation rationale documented in the test comments.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 33s | <b>Cost:</b> $2.30 | <b>Commit:</b> <code>6e16106c480b89b43f67dcbe945f7e8526d023b0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1711, opened 2026-06-23
feat(mac-giver-client): add MacGiver FACS permission client

ravverma:
Introduces the new `@adobe/spacecat-shared-mac-giver-client` package — a thin client over the MacGiver `/api/facs/permissions/check` endpoint, consumed by `spacecat-auth-service` at login to mint the `facs_permissions` JWT claim.

## API
- `checkListOfPermission({ userId, imsOrgId, permissions })` — evaluate an explicit list; returns the allowed subset.
- `checkAllPermission({ userId, imsOrgId, namespaces })` — evaluate every permission in the given namespaces; returns the allowed subset.
- `macGiverClientWrapper` attaches the client to `context`.

## Notable correctness details
- Request body uses canonical FACS shapes: `subject { type: 'user' }` and `object { type: 'org' }` (not `'organization'`, which fails relation resolution → 500); `permissions` is sent only when non-empty.
- Reads the bearer from `getServiceAccessToken().access_token` (throws if missing); does **not** send `X-User-Token` (FACS uses it for logging only).
- Non-2xx / transport failure **throws** (tagged `warn`) so callers fail-safe; evaluated-but-empty returns `[]`.

## Tests
100% coverage (lines/branches/functions).

Part of the MAC / FACS state-layer rollout. Pairs with a follow-up `http-utils` PR (AuthInfo FACS methods + constants + IMS-handler RBAC block) to unblock the auth-service PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1712, opened 2026-06-23
feat(http-utils): FACS AuthInfo accessors, constants, and IMS-handler RBAC block

ravverma:
Adds the `@adobe/spacecat-shared-http-utils` pieces the MAC / FACS rollout needs — additive only, no facsWrapper in this slice. Unblocks the `spacecat-auth-service` PR.

## Changes
- **AuthInfo**: `getFacsPermissions()` and `hasFacsPermission(permission)` — read the `facs_permissions` JWT claim.
- **constants**: `FT_MAC_FACS_PERMISSIONS` (per-product LaunchDarkly flag table) and `X_PRODUCT_HEADER` — single source of truth consumed by both auth-service (login-time lookup) and the request-time gate.
- **IMS handler**: `isOrgRbacEnabled()` blocks IMS auth for orgs with the product's FACS flag enabled, forcing them down the JWT/auth-service path (which mints `facs_permissions`). Mirrors the existing read-only-org gate — product-scoped via `x-product`, first-org-only, fail-open on LD errors.

## Tests
Full package suite green; coverage gate met (the added code is fully covered).

Part of the MAC / FACS state-layer rollout. Pairs with the mac-giver-client PR (#1711); together they give the published versions that the auth-service PR will bump to.


MysticatBot (changes requested):
Hey @ravverma,

**Verdict: Request changes** - one authorization-relevant type-safety gap to address.
**Complexity: HIGH** - Auth/IMS risk surface on a medium diff.
_Changes: Adds FACS permission accessors to AuthInfo, product-scoped LaunchDarkly flag constants, and an RBAC enforcement gate to the IMS auth handler (5 files)._

## Must fix before merge

1. [Important] `getFacsPermissions()` trusts the JWT claim type without validation - a non-array truthy value (e.g. string) causes `hasFacsPermission` to do substring matching, silently producing false positives on authorization checks - `src/auth/auth-info.js:95` (details inline)

<details>

- nit: Missing blank line between `isOrgRbacEnabled` closing brace and the `@deprecated` JSDoc block - `src/auth/handlers/ims.js:230`
- nit: Both gates throw identical `new Error('Unauthorized')` making it impossible to distinguish read-only block from RBAC block without correlating log lines - `src/auth/handlers/ims.js:309`
- suggestion: Freeze `FT_MAC_FACS_PERMISSIONS` with `Object.freeze()` to prevent accidental mutation by consumers - `src/auth/constants.js:40`
</details>

packages/spacecat-shared-http-utils/src/auth/auth-info.js
```diff
@@ -92,6 +92,10 @@ export default class AuthInfo {
 
   isReadOnlyAdmin() { return this.profile?.is_read_only_admin; }
 
+  getFacsPermissions() { return this.profile?.facs_permissions || []; }
```
MysticatBot: issue (blocking): `getFacsPermissions()` returns `this.profile?.facs_permissions || []` without validating the claim is actually an array. If `facs_permissions` is present but is a non-array truthy value (e.g. a string), the `|| []` fallback does not activate, and `hasFacsPermission` calls `.includes()` on the string. String `.includes()` does substring matching, not element matching - so `hasFacsPermission("can")` would return `true` on a string like `"llmo/can_read"`, silently granting access for partial matches.

This is an authorization boundary where fail-safe matters. The existing `isAdmin()`/`isReadOnlyAdmin()` accessors return boolean-like values so type confusion is benign there, but here the return value is iterated for authorization decisions.

**Fix:**
```js
getFacsPermissions() {
  const perms = this.profile?.facs_permissions;
  return Array.isArray(perms) ? perms : [];
}
```

This preserves the fail-closed contract (unknown shape returns empty permissions) while preventing substring-match false positives.

MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - all prior findings addressed, no new blocking issues.
**Complexity: HIGH** - Auth/IMS risk surface on a medium diff.
_Changes: Adds FACS permission accessors to AuthInfo, product-scoped LaunchDarkly flag constants, and an RBAC enforcement gate to the IMS auth handler (5 files)._

<details>

- suggestion: Add a warn/debug log inside the bare `catch {}` in `isOrgRbacEnabled` so LD failures are observable in production (mirrors `isOrgBlockedFromImsAuth` which has the same gap, but new code is a good place to set the better standard) - `src/auth/handlers/ims.js:224`
- nit: LD stub in RBAC test uses `.withArgs('FF_LLMO-3026')` without matching the second arg (org ID), making the happy-path stub less strict than the assertion that follows it - `test/auth/handlers/ims.test.js:659`
</details>

- `getFacsPermissions()` now fails closed on unexpected claim shapes via `Array.isArray` guard
- `FT_MAC_FACS_PERMISSIONS` frozen with `Object.freeze()`
- Blank line added before `@deprecated` JSDoc
- Identical error text nit - author's reasoning accepted (distinct warn log lines disambiguate, error is internal-only)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 12s | <b>Cost:</b> $4.29 | <b>Commit:</b> <code>3c83e003f80088e03432017c9ebb32ec2f004f78</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1717, opened 2026-06-24
feat(http-utils): add facsWrapper for hybrid FACS authorization

ravverma:
## What

Adds the **FACS authorization wrapper** (`facsWrapper`) to `spacecat-shared-http-utils`, plus its supporting modules. This is the piece intentionally deferred from the minimal-scope shared PRs (#1711 / #1712) — it builds on the `AuthInfo` FACS helpers and constants already on `main`.

## Changes (net-new, additive only)

- `src/auth/facs-wrapper.js` — per-route FACS enforcement for external customer users; gated by a LaunchDarkly flag. Internal identities and Adobe internal orgs bypass.
- `src/auth/facs-resource-resolver.js` — resolves the ReBAC resource for a request; defers to the controller when no resource is resolvable.
- `src/auth/facs-state-layer.js` — `normalizeImsOrgId` (canonicalises bare ident → `<ident>@<authSrc>`) and `findFacsResourceBinding`.
- `src/auth/route-utils.js` — route-pattern matching helpers.
- `src/index.js` / `src/index.d.ts` — barrel exports + `facsWrapper` type.

No changes to `auth-info.js`, `constants.js`, or `handlers/ims.js` — those shipped in #1712 and `main`'s versions are preserved.

## Testing

- `npm test -w packages/spacecat-shared-http-utils` → **474 passing**
- Coverage: 100% lines / 99.12% branches (gate: 100% / 97%); all FACS files at 100%.
- Lint clean.


packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js
```diff
@@ -0,0 +1,478 @@
...
+      || authInfo?.isS2SConsumer?.()
+      || authInfo?.isReadOnlyAdmin?.()
+      || authType === 'legacyApiKey'
+      || authType === 'scopedApiKey'
+    ) {
+      log.info({
+        tag: 'facs',
+        bypass: 'internal-identity',
+        method,
+        suffix,
```
MysticatBot, line 230: **issue (blocking):** The LD flag evaluation hardcodes `` `${orgId}@AdobeOrg` `` via template literal. If `getTenantIds()` returns a pre-suffixed value (e.g. `ACME@AdobeOrg`), this produces `ACME@AdobeOrg@AdobeOrg` - the flag evaluates against a non-existent org, returns false, and FACS enforcement is silently bypassed.

Fix: `normalizeImsOrgId(orgId)` is already imported and used in step 7. Use it here too:
```js
isFacsEnabled = await ldClient.isFlagEnabledForIMSOrg(flagKey, normalizeImsOrgId(orgId));
```

packages/spacecat-shared-http-utils/src/index.d.ts
```diff
@@ -35,6 +35,20 @@ export declare function unauthorized(message?: string, headers?: object): Respon
...
+ * FACS authorization wrapper for the helix-shared-wrap `.with()` chain.
+ * Enforces FACS permissions for external customer users per route, gated by a LaunchDarkly
+ * feature flag. Internal identities and Adobe internal orgs always bypass.
+ *
+ * @param fn - The handler to wrap.
+ * @param opts - Options containing a routeFacsCapabilities map of route patterns to FACS actions.
+ * @returns A wrapped handler.
+ */
+export function facsWrapper(
+  fn: Function,
```
MysticatBot: **issue (blocking):** `Record<string, string>` does not match the actual shape. The real config is deeply nested (`PRODUCTS_ROUTES`, `PRODUCTS_FACS_RESOURCE_PARAM_ALIASES`, `INTERNAL_ROUTES`). TS consumers get no compile-time guidance - violates the repo rule that `.d.ts` files accurately describe public APIs.

MysticatBot (changes requested):

MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - both prior findings addressed; no new blocking issues.
**Complexity: HIGH** - large diff; auth/IMS signal.
_Changes: adds facsWrapper implementing a hybrid FACS permission model (JWT union state-layer grants) with per-product route governance, resource resolution, and LaunchDarkly flag gating (10 files)._

- Double-suffix on LD flag key now prevented by `normalizeImsOrgId` (idempotent); regression test confirms pre-canonicalized orgs stay single-suffixed.
- Weak `.d.ts` types replaced with proper `FacsRouteCapabilities` interface with correct nested shapes.

<details>
<summary>Non-blocking (6): minor issues and suggestions</summary>

- suggestion: Parallelize the user-scoped and org-scoped `findFacsResourceBinding` calls with `Promise.all` - they are independent point reads and the sequential await doubles latency on the state-layer path - `facs-wrapper.js:377-388`
- suggestion: Lift `parseFacsExceptionInternalOrgs(context.env)` into a lazily-cached closure variable - it re-parses the same env var on every request, unlike `aliasLookupsPerProduct` which is built once at construction - `facs-wrapper.js:320`
- suggestion: Consider an explicit `DEFERRED_ROUTES` allowlist so routes that legitimately defer to the controller are a deliberate config choice rather than a silent fallback when resource resolution returns null - `facs-wrapper.js:376`
- nit: `FACS_NON_RESOURCE_PARAMS` declared in the TypeScript interface but not consumed by any code in this PR - creates a false expectation that setting it has runtime effect - `index.d.ts:58`
- nit: Local `forbidden()` / `serviceUnavailable()` / `internalServerError()` helpers duplicate the barrel-exported response factories with slightly different serialization - `facs-wrapper.js:24-36`
- suggestion: Add a test where the first `findFacsResourceBinding` resolves null and the second rejects - validates fail-closed behavior across the full two-lookup sequence under partial failure - `facs-wrapper.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 42s | <b>Cost:</b> $8.95 | <b>Commit:</b> <code>6eaafb1615da6cb902f6cadc3e98dae40b15bdc2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1727, opened 2026-06-25
feat: add schema-less URL support to pathname helpers

dhavkuma:
## Summary

- \`toPathname\` now calls \`prependSchema\` before parsing, so schema-less inputs like \`example.com/path\` correctly resolve to \`/path\` instead of falling back to the raw string
- Added a leading-slash guard (\`url.startsWith('/')\`) to short-circuit before \`prependSchema\` and return the path as-is, preventing the WHATWG URL parser from mangling paths like \`/some/PATH\` into a bogus hostname
- Added \`hasSamePathname\` and \`allHaveSamePathname\` exports for normalized, case-insensitive pathname comparison across domain-based URLs
- Updated and extended tests: fixed the \`BULK.COM/page\` fallback test, added schema-less coverage for all three helpers, and added an explicit catch-branch test

## Test plan

- [ ] \`npm test -w packages/spacecat-shared-utils\` passes with 100% coverage
- [ ] \`toPathname('example.com/path')\` → \`/path\`
- [ ] \`toPathname('/some/PATH')\` → \`/some/PATH\` (passed through as-is)
- [ ] \`hasSamePathname('a.com/page', 'https://b.com/page')\` → \`true\`
- [ ] \`allHaveSamePathname(['a.com/page', 'https://b.com/page'], 'c.com/page')\` → \`true\`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -525,7 +525,10 @@ export function toPathname(url) {
     return '';
   }
   try {
-    const { pathname } = new URL(url);
+    if (url.startsWith('/')) {
+      return url.toLowerCase();
+    }
+    const { pathname } = new URL(prependSchema(url));
     return pathname === '/' ? pathname : pathname.replace(/\/$/, '').toLowerCase();
   } catch {
     return url.toLowerCase();
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -527,8 +527,7 @@ export function toPathname(url) {
   }
   try {
     if (url.startsWith('/')) {
-      const lower = url.toLowerCase();
-      return lower === '/' ? lower : lower.replace(/\/$/, '');
+      return url.toLowerCase();
     }
     const { pathname } = new URL(prependSchema(url));
     return pathname === '/' ? pathname : pathname.replace(/\/$/, '').toLowerCase();
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -513,13 +513,13 @@ function filterBySiteScope(urls, siteBaseUrl) {
 }
 
 /**
- * Extracts the pathname from a URL string. Expects a domain-based URL, with or without schema
- * (e.g. 'https://example.com/path' or 'example.com/path'). Trailing slashes are stripped on
- * non-root paths and the result is lowercased. Leading-slash inputs (already a pathname) are
- * returned as-is. Returns '' for non-string or empty input.
+ * Extracts the pathname from a domain-based URL string (e.g. 'https://example.com/path' or
+ * 'example.com/path'). Trailing slashes are stripped on non-root paths and the result is
+ * lowercased. Inputs that already start with '/' are returned unchanged. Returns '' for
+ * non-string or empty input.
  *
- * @param {string} url
- * @returns {string} pathname, or the input as-is for leading-slash paths, or '' for non-strings
+ * @param {string} url - Domain-based URL, with or without schema.
+ * @returns {string} Normalized pathname.
  */
 export function toPathname(url) {
   if (!url || typeof url !== 'string') {
@@ -537,10 +537,11 @@ export function toPathname(url) {
 }
 
 /**
- * Checks whether two URLs share the same normalized pathname.
- * Comparison is case-insensitive and ignores trailing slashes on non-root paths.
- * @param {string} url - URL to compare.
- * @param {string} referenceUrl - URL to compare against.
+ * Checks whether two domain-based URLs share the same normalized pathname.
+ * Both arguments must be domain-based URLs (with or without schema). Comparison is
+ * case-insensitive and ignores trailing slashes on non-root paths.
+ * @param {string} url - Domain-based URL to compare.
+ * @param {string} referenceUrl - Domain-based URL to compare against.
  * @returns {boolean} True if both URLs resolve to the same pathname.
  */
 export function hasSamePathname(url, referenceUrl) {
@@ -549,8 +550,9 @@ export function hasSamePathname(url, referenceUrl) {
 
 /**
  * Checks whether every URL in an array shares the same normalized pathname as a reference URL.
- * @param {string[]} urls - Array of URLs to check.
- * @param {string} referenceUrl - URL whose pathname all entries must match.
+ * All entries and the reference must be domain-based URLs (with or without schema).
+ * @param {string[]} urls - Array of domain-based URLs to check.
+ * @param {string} referenceUrl - Domain-based URL whose pathname all entries must match.
  * @returns {boolean} True if every URL in the array has the same pathname as referenceUrl.
  */
 export function allHaveSamePathname(urls, referenceUrl) {
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -514,8 +514,9 @@ function filterBySiteScope(urls, siteBaseUrl) {
 
 /**
  * Extracts the pathname from a URL string, stripping trailing slashes on non-root paths.
- * The result is always lowercased. Falls back to the raw string (also lowercased) when
- * the URL is not parseable (e.g. invalid or relative). Returns '' for non-string input.
+ * The result is always lowercased. Leading-slash inputs are returned directly (lowercased,
+ * trailing slash stripped). Falls back to the raw lowercased string when the URL is not
+ * parseable (e.g. invalid). Returns '' for non-string input.
  *
  * @param {string} url
  * @returns {string} pathname, or the original string on parse failure, or '' for non-strings
@@ -526,7 +527,8 @@ export function toPathname(url) {
   }
   try {
     if (url.startsWith('/')) {
-      return url.toLowerCase();
+      const lower = url.toLowerCase();
+      return lower === '/' ? lower : lower.replace(/\/$/, '');
     }
     const { pathname } = new URL(prependSchema(url));
     return pathname === '/' ? pathname : pathname.replace(/\/$/, '').toLowerCase();
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -527,7 +527,7 @@ export function toPathname(url) {
   }
   try {
     if (url.startsWith('/')) {
-      return url.toLowerCase();
+      return url;
     }
     const { pathname } = new URL(prependSchema(url));
     return pathname === '/' ? pathname : pathname.replace(/\/$/, '').toLowerCase();
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -515,11 +515,11 @@ function filterBySiteScope(urls, siteBaseUrl) {
 /**
  * Extracts the pathname from a domain-based URL string (e.g. 'https://example.com/path' or
  * 'example.com/path'). Trailing slashes are stripped on non-root paths and the result is
- * lowercased. Inputs that already start with '/' are returned unchanged. Returns '' for
+ * lowercased. Inputs that already start with '/' are passed through as-is. Returns '' for
  * non-string or empty input.
  *
- * @param {string} url - Domain-based URL, with or without schema.
- * @returns {string} Normalized pathname.
+ * @param {string} url - Domain-based URL, with or without schema, or a leading-slash path.
+ * @returns {string} Normalized pathname, or the input unchanged for leading-slash paths.
  */
 export function toPathname(url) {
   if (!url || typeof url !== 'string') {
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -513,13 +513,13 @@ function filterBySiteScope(urls, siteBaseUrl) {
 }
 
 /**
- * Extracts the pathname from a URL string, stripping trailing slashes on non-root paths.
- * The result is always lowercased. Leading-slash inputs are returned directly (lowercased,
- * trailing slash stripped). Falls back to the raw lowercased string when the URL is not
- * parseable (e.g. invalid). Returns '' for non-string input.
+ * Extracts the pathname from a URL string. Expects a domain-based URL, with or without schema
+ * (e.g. 'https://example.com/path' or 'example.com/path'). Trailing slashes are stripped on
+ * non-root paths and the result is lowercased. Leading-slash inputs (already a pathname) are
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -525,7 +525,10 @@ export function toPathname(url) {
     return '';
   }
   try {
-    const { pathname } = new URL(url);
+    if (url.startsWith('/')) {
+      return url.toLowerCase();
```
MysticatBot: issue (blocking): The early return `url.toLowerCase()` bypasses the trailing-slash stripping logic that the main branch applies (`pathname.replace(/\/$/, '')`).

This creates an inconsistency:
- `toPathname('https://example.com/path/')` returns `'/path'` (trailing slash stripped)
- `toPathname('/path/')` returns `'/path/'` (trailing slash preserved)

This will cause `hasSamePathname('/path/', 'example.com/path')` to return `false` when both refer to the same resource.

Fix:
```js
if (url.startsWith('/')) {
  const lower = url.toLowerCase();
  return lower === '/' ? lower : lower.replace(/\/$/, '');
}
```

Also add a test: `expect(toPathname('/some/path/')).to.equal('/some/path')`.

MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - the leading-slash short-circuit path has an inconsistency that will produce incorrect `hasSamePathname` comparisons.
**Complexity: LOW** - small diff, single utility function.
_Changes: makes `toPathname` handle schema-less URLs by prepending `https://` before parsing, and short-circuits leading-slash paths (2 files)._

## Must fix before merge
1. [Important] Trailing-slash normalization bypassed on leading-slash short-circuit - `packages/spacecat-shared-utils/src/url-helpers.js:528` (details inline)
2. [Important] Behavioral change mislabeled as `refactor` in commit prefix - the return value of `toPathname('BULK.COM/page')` changes from `'bulk.com/page'` to `'/page'`. This is a semantic change in a shared library consumed by multiple services. The `refactor` conventional commit prefix signals no behavioral change and produces a patch bump via semantic-release. Downstream consumers will not review their assumptions. Change the prefix to `feat(utils)` so the release tooling produces a minor bump that communicates the contract change.

<details>

- nit: JSDoc on `toPathname` still says the catch block handles "relative" paths, but relative paths (leading-slash) now hit the early return, not the catch - `packages/spacecat-shared-utils/src/url-helpers.js:517`
- nit: Bare-word inputs without dots or slashes (e.g., `'not-a-domain'`) now return `'/'` instead of the lowercased input, since `prependSchema` makes them parseable hostnames - low risk but a silent behavioral change for that input class - `packages/spacecat-shared-utils/src/url-helpers.js:530`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 58s | <b>Cost:</b> $3.17 | <b>Commit:</b> <code>aa4207c6a84becba49888b0833b9de59c1d0e4fc</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -513,19 +513,23 @@ function filterBySiteScope(urls, siteBaseUrl) {
...
- * @returns {string} pathname, or the original string on parse failure, or '' for non-strings
+ * @returns {string} pathname, or the input as-is for leading-slash paths, or '' for non-strings
  */
 export function toPathname(url) {
   if (!url || typeof url !== 'string') {
     return '';
   }
   try {
-    const { pathname } = new URL(url);
+    if (url.startsWith('/')) {
```
MysticatBot, line 529: issue (blocking): The early return `return url;` bypasses both trailing-slash stripping (`pathname.replace(/\/$/, '')`) and lowercasing that the URL branch applies. This creates observable inconsistency:

- `toPathname('https://example.com/path/')` returns `'/path'` (trailing slash stripped)
- `toPathname('/path/')` returns `'/path/'` (trailing slash preserved)
- `toPathname('example.com/PATH')` returns `'/path'` (lowercased)
- `toPathname('/PATH')` returns `'/PATH'` (case preserved)

This directly breaks `hasSamePathname('/path/', 'example.com/path')` (returns `false`) and `hasSamePathname('/PATH', 'example.com/path')` (returns `false`). The prior review raised trailing-slash specifically; you initially fixed it then reverted in commits c5e46bd/c87dedf without explaining why.

Fix:
```js
if (url.startsWith('/')) {
  return url === '/' ? url : url.replace(/\/$/, '').toLowerCase();
}
```

If the verbatim pass-through is intentional for a use case the review is not seeing, explain the reasoning in the PR discussion so reviewers can evaluate the trade-off.

packages/spacecat-shared-utils/test/url-helpers.test.js
```diff
@@ -1375,6 +1387,10 @@ describe('URL Utility Functions', () => {
     it('does not crash when url is null', () => {
       expect(hasSamePathname(null, 'https://b.com/page')).to.be.false;
     });
+
+    it('matches schema-less URL against an absolute URL', () => {
```
MysticatBot, line 1391: issue (blocking): The new `hasSamePathname` and `allHaveSamePathname` tests only cover matching-case, no-trailing-slash inputs. The normalization asymmetry means these comparisons silently fail when formats are mixed with case or trailing-slash differences:

- `hasSamePathname('/Page', 'example.com/page')` - expected true, currently false
- `hasSamePathname('/path/', 'example.com/path')` - expected true, currently false
- `hasSamePathname('/path', '/PATH')` - expected true, currently false

Add tests documenting the expected behavior for these cross-format scenarios. If the normalization fix is applied (consistent lowercasing + trailing-slash strip), these should all assert `true`.

MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - the leading-slash short-circuit bypasses normalization that `hasSamePathname` relies on, producing incorrect comparison results.
**Complexity: LOW** - small diff, single utility function.
_Changes: makes `toPathname` handle schema-less URLs via `prependSchema` and short-circuits leading-slash inputs as pass-through (2 files)._

## Must fix before merge
1. [Critical] Leading-slash path bypass skips trailing-slash stripping and lowercasing, breaking `hasSamePathname` - `packages/spacecat-shared-utils/src/url-helpers.js:529` (details inline)
2. [Important] No test coverage for the cross-format comparison failure cases the asymmetry introduces - `packages/spacecat-shared-utils/test/url-helpers.test.js` (details inline)

<details>

- nit: JSDoc says "the result is lowercased" but leading-slash paths preserve case; update to note the exception - `packages/spacecat-shared-utils/src/url-helpers.js:516`
- nit: PR title uses `fix:` but the change is a feature (new schema-less URL support). Semantic-release already handles the bump via the `feat(utils)` commit, but the title is misleading about the change's nature.
- suggestion: Document expected behavior for bare-word inputs without dots or slashes (e.g. `'not-a-domain'` now returns `'/'` instead of `'not-a-domain'`) - `packages/spacecat-shared-utils/src/url-helpers.js:530`
- suggestion: Catch-branch test should use mixed-case input (e.g. `'[Invalid-URL'`) to actually validate lowercasing behavior - `packages/spacecat-shared-utils/test/url-helpers.test.js:1351`
</details>

- Commit prefix changed to `feat(utils):` so semantic-release produces a minor bump signaling the contract change.
- JSDoc updated to describe the leading-slash short-circuit behavior.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 46s | <b>Cost:</b> $3.56 | <b>Commit:</b> <code>701638a88fb2879b55a33dba78f2ddfb26f8a2ce</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -513,30 +513,35 @@ function filterBySiteScope(urls, siteBaseUrl) {
...
+ * @returns {string} Normalized pathname.
  */
 export function toPathname(url) {
   if (!url || typeof url !== 'string') {
     return '';
   }
   try {
-    const { pathname } = new URL(url);
+    if (url.startsWith('/')) {
+      return url;
```
MysticatBot, line 530: issue (blocking): The early return `return url;` bypasses both trailing-slash stripping (`pathname.replace(/\/$/, '')`) and lowercasing (`.toLowerCase()`) that the URL-parsed branch applies. This creates observable inconsistency that breaks the comparison functions:

- `toPathname('/Page')` returns `'/Page'` (case preserved)
- `toPathname('example.com/page')` returns `'/page'` (lowercased)
- `hasSamePathname('/Page', 'example.com/page')` returns `false` (should be `true`)
- `hasSamePathname('/path/', 'example.com/path')` returns `false` (should be `true`)

The JSDoc for `hasSamePathname` promises "case-insensitive" comparison, which is violated when one argument starts with `/`. This was raised in both prior reviews. The author commented it was fixed but then reverted the fix in commits c5e46bd/c87dedf without explaining why.

Fix:
```js
if (url.startsWith('/')) {
  return url === '/' ? url : url.replace(/\/$/, '').toLowerCase();
}
```

packages/spacecat-shared-utils/test/url-helpers.test.js
```diff
@@ -1330,12 +1330,24 @@ describe('URL Utility Functions', () => {
...
+    it('extracts pathname from a schema-less URL with trailing slash', () => {
+      expect(toPathname('example.com/')).to.equal('/');
+    });
+
+    it('extracts pathname from a schema-less URL with a deep path', () => {
+      expect(toPathname('example.com/path/page')).to.equal('/path/page');
+    });
+
+    it('handles leading-slash paths as-is', () => {
+      expect(toPathname('/some/PATH')).to.equal('/some/PATH');
```
MysticatBot, line 1346: issue (blocking): This test encodes the normalization bypass as expected behavior (`'/some/PATH'` instead of `'/some/path'`). Once the normalization is fixed, update to:

```js
expect(toPathname('/some/PATH')).to.equal('/some/path');
expect(toPathname('/some/path/')).to.equal('/some/path');
```

Additionally, the `hasSamePathname` and `allHaveSamePathname` suites are missing cross-format comparison scenarios:
- `hasSamePathname('/page', 'example.com/page')` (leading-slash vs domain)
- `hasSamePathname('/PAGE', 'example.com/page')` (case mismatch across formats)
- `hasSamePathname('/page/', 'example.com/page')` (trailing slash vs clean)

These would all return `false` today due to the normalization bypass - adding them documents the expected contract and guards against regression.

MysticatBot (dismissed):
Hey @dhavkuma,

**Verdict: Request changes** - the leading-slash short-circuit still bypasses normalization that `hasSamePathname` relies on, producing incorrect cross-format comparisons.
**Complexity: LOW** - small diff, single utility function.
_Changes: updates `toPathname` to handle schema-less URLs via `prependSchema` and short-circuits leading-slash inputs as verbatim pass-through (2 files)._

## Must fix before merge
1. [Critical] Leading-slash early return bypasses lowercasing and trailing-slash stripping, breaking `hasSamePathname` for mixed-format inputs - `packages/spacecat-shared-utils/src/url-helpers.js:530` (details inline)
2. [Important] Tests encode the broken behavior and omit cross-format comparison scenarios that would catch the regression - `packages/spacecat-shared-utils/test/url-helpers.test.js:1346` (details inline)

<details>

- nit: JSDoc says "the result is lowercased" (line 517) while also saying "Inputs that already start with '/' are returned unchanged" (line 518) - these contradict each other; after fixing the normalization, remove the "returned unchanged" clause - `packages/spacecat-shared-utils/src/url-helpers.js:517`
</details>

- Commit prefix changed to `feat(utils):` so semantic-release produces a minor bump signaling the contract change.
- JSDoc reworded to describe the domain-based URL contract.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 55s | <b>Cost:</b> $2.74 | <b>Commit:</b> <code>3db1314c5a2b866bdcd026783948bd07060000ae</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-utils/src/url-helpers.js
dhavkuma: urls values should have domain names. If not return them as is.

packages/spacecat-shared-utils/test/url-helpers.test.js
dhavkuma: urls passed should have domain name. otherwise return as is.

packages/spacecat-shared-utils/src/url-helpers.js
dhavkuma: urls expected to contain domain

packages/spacecat-shared-utils/test/url-helpers.test.js
dhavkuma: urls should contain domains

MysticatBot (approved):
Hey @dhavkuma,

**Verdict: Approve** - the contract narrowing to domain-based URLs is documented, tested, and internally consistent.
**Complexity: LOW** - small diff, single utility function.
_Changes: makes `toPathname` handle schema-less URLs via `prependSchema` and short-circuits leading-slash inputs as verbatim pass-through (2 files)._

<details>

- nit: Leading-slash early return preserves case (`/some/PATH` stays `/some/PATH`) while the domain path lowercases (`example.com/PATH` yields `/path`) and the catch fallback also lowercases. This asymmetry is documented ("passed through as-is") but creates a subtle footgun if callers ever mix input formats in comparisons. Consider lowercasing the leading-slash branch too for uniformity, or adding a brief inline comment explaining why verbatim is intentional. - `packages/spacecat-shared-utils/src/url-helpers.js:529`
- nit: The fallback test uses an already-lowercase input (`'[invalid-url'`), so it does not actually verify that the catch branch lowercases. Use mixed-case input (e.g. `'[Invalid-URL'`) and assert `'[invalid-url'` to prove the behavior. - `packages/spacecat-shared-utils/test/url-helpers.test.js:1350`
</details>

- Leading-slash normalization concern: author documented the domain-based URL contract explicitly in JSDoc; leading-slash is an intentional defensive pass-through, not a primary use case.
- Cross-format comparison test coverage: tests correctly encode the documented contract; `hasSamePathname` JSDoc now states "both arguments must be domain-based URLs."

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 56s | <b>Cost:</b> $2.93 | <b>Commit:</b> <code>8231044e6c9e4bfbdaa76e0a89c7d42a70b2e392</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1733, opened 2026-06-26
test(project-engine-client): gate overlay corrections against the vendored swagger

rainer-friederich:
## 1. Abstract
Adds an offline CI freshness gate to `spacecat-shared-project-engine-client` that asserts every overlay correction in `spec/overlays/corrections.yaml` still applies and is still necessary against the vendored Semrush swagger, and hardens the overlay build script to fail loudly on a stale correction.

## 2. Reasoning
The package mocks the Semrush Project Engine API by applying overlay corrections (CR1–CR10) on top of a vendored swagger that is never hand-edited. Each correction patches a divergence between that vendored spec and the real live API. Until now nothing detected when a correction silently drifted — became a dead no-op because its target moved, or became redundant because the vendored swagger caught up. That only surfaced when someone re-ran the manual live-capture recipe. The build script already computed a `staleRemove` signal but merely warned and exited zero, so drift could land unnoticed.

Behaviour delta:

- **New CI gate (offline).** A test reproduces the `spec:convert` step (vendored swagger 2.0 → OAS3) and then applies each overlay action against that converted doc, asserting per correction that it (a) still matches more than zero nodes — *still applies* — and (b) actually mutates the spec — *still necessary*. A re-vendored swagger that makes any correction stale (target moved/removed) or redundant (upstream now declares it) now fails CI, instead of drifting silently until the next manual pass. No live API, credentials, or scheduled job are involved.
- **Build script now fails loudly.** `scripts/apply-overlay.mjs` `main()` previously printed a warning and exited zero on a 0-match (stale) `remove`. It now returns a process exit code that is non-zero when any correction is stale, so `npm run spec:overlay` — and the `npm run generate` chain — stops rather than producing types/models from a quietly-degraded overlay. `main()` is exported and accepts injectable spec/overlay paths and a logger so the exit-code behaviour is directly testable.
- **Docs.** Added the missing CR10 entry to the corrections summary comment and a short note describing the gate; corrected a stale "CR1–CR3" reference in the script docblock to "CR1–CR10".

Scope note: detecting an *insufficient* correction (upstream added a field the consumer uses that the mock does not model) is intentionally out of scope — that needs live-API / consumer knowledge, which this offline gate deliberately avoids.

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-shared/issues/1731

Ran the real overlay pipeline against the committed sources this session — `npm run spec:convert` followed by `npm run spec:overlay` — and observed it apply all 21 overlay actions and exit zero (no stale corrections today), confirming the hardened CLI's success path on the real default paths. `build/openapi3.json` is gitignored, so the gate reproduces the convert step from the vendored YAML rather than reading a committed artefact.

## 7. Test plan
- Local: the freshness gate runs as part of the package's normal test suite (`npm test -w packages/spacecat-shared-project-engine-client`) — fully offline, no setup. It is also exercised by the real `spec:convert` + `spec:overlay` run described in section 6.
- Per-environment: none. This change touches only the package's test suite and a build-time script (`scripts/`), neither of which is published (`files: ["src"]`) or deployed. There is no eph/dev/stage/prod verification step.

### DIFF
--- a/packages/spacecat-shared-project-engine-client/scripts/apply-overlay.mjs
+++ b/packages/spacecat-shared-project-engine-client/scripts/apply-overlay.mjs
@@ -14,7 +14,7 @@
  * Applies the OpenAPI Overlay in spec/overlays/corrections.yaml to the
  * swagger2openapi output build/openapi3.json, in place. The overlay is the single
  * source of truth for the corrections that align the vendored swagger with the
- * live API's actual behaviour (see that file for CR1-CR10); this script just
+ * live API's actual behaviour (see that file for the full CR list); this script just
  * executes it. The vendored spec/projectengine_swagger_public.yaml is never modified.
  *
  * It implements the subset of the Overlay spec the overlay uses — `update`
@@ -218,8 +218,10 @@ export function applyOverlay(spec, overlay) {
  * Returns a process exit code so the build fails loudly when a correction has gone stale: a
  * 0-match `remove` (`staleRemove`) means upstream may have caught up and the correction should be
  * pruned, so we exit 1. (A 0-match `update` throws from {@link applyAction} and aborts before any
- * write — also a hard failure.) Paths and the logger are injectable so the behaviour is testable
- * without spawning a subprocess; the defaults target this package's real build artefact.
+ * write — also a hard failure.) The overlaid spec is written back BEFORE the stale check, so a
+ * failed run still leaves the produced artefact on disk for local inspection. Paths and the logger
+ * are injectable so the behaviour is testable without spawning a subprocess; the defaults target
+ * this package's real build artefact.
  *
  * @param {object} [opts]
  * @param {string} [opts.specPath] OAS3 doc to overlay (default: build/openapi3.json)
--- a/packages/spacecat-shared-project-engine-client/scripts/apply-overlay.mjs
+++ b/packages/spacecat-shared-project-engine-client/scripts/apply-overlay.mjs
@@ -14,7 +14,7 @@
  * Applies the OpenAPI Overlay in spec/overlays/corrections.yaml to the
  * swagger2openapi output build/openapi3.json, in place. The overlay is the single
  * source of truth for the corrections that align the vendored swagger with the
- * live API's actual behaviour (see that file for CR1-CR3); this script just
+ * live API's actual behaviour (see that file for CR1-CR10); this script just
  * executes it. The vendored spec/projectengine_swagger_public.yaml is never modified.
  *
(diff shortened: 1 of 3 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean, well-scoped change that delivers exactly what issue 1731 asks for.
**Complexity: LOW** - small diff, single package, test + build script only.
_Changes: Adds an offline CI freshness gate that asserts every overlay correction still applies and is still necessary against the vendored swagger, and hardens the build script to fail on staleness (3 files)._
Note: Recommend a human read before merge - possible divergence from the documented design (heuristic, flagged by the conventions reviewer: CR10 adds fields not present in the vendored swagger). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `writeFileSync` executes before the stale check, so exit 1 leaves a partially-overlaid file on disk - consider a one-line JSDoc note that the write-then-fail is deliberate (allows local inspection of the produced artefact) - `scripts/apply-overlay.mjs:237`
- suggestion: the docblock reference "CR1-CR10" will need updating with each new correction; consider "see that file for the full CR list" to avoid the maintenance tick - `scripts/apply-overlay.mjs:17`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 9s | <b>Cost:</b> $4.92 | <b>Commit:</b> <code>a9ec839b1c7996ed8093f311d3fd371898e13c66</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - docblock-only follow-up cleanly addresses both prior nits, no functional changes.
**Complexity: MEDIUM** - medium diff, single package, test + build-script only.
_Changes: Adds an offline CI freshness gate for overlay corrections and hardens the build script to fail on staleness (3 files)._

- "CR1-CR10" replaced with "the full CR list" to avoid maintenance burden on each new correction.
- Write-before-stale-check behavior now documented in main() JSDoc, making the deliberate ordering explicit.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 23s | <b>Cost:</b> $2.52 | <b>Commit:</b> <code>5aaad0e62172225865a1b09efa4bdfd4408afae8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1734, opened 2026-06-26
feat(user-manager-client): stateful Counterfact mock for the sub-workspace lifecycle gateway (LLMO-5616)

rainer-friederich:
## 1. Abstract
A stateful Counterfact mock of the Semrush User Manager gateway (`/enterprise/users/api`) for the `spacecat-shared-user-manager-client` package — modelling the sub-workspace lifecycle surface the api-service serenity transport calls — plus the typed-client/overlay scaffolding around it (spec overlay, generated types, freshness gate, package contract doc).

## 2. Reasoning
The api-service serenity transport drives a multi-step sub-workspace lifecycle (create child → poll status → list family → transfer resources → delete) against the live User Manager gateway. There was no way to exercise that consumer path in tests or local dev without hitting real Semrush, and the vendored swagger diverges from the gateway's real responses in several places. This brings the User Manager client to parity with the project-engine-client mock that already shipped, so both Semrush gateways have a faithful, offline, stateful test double.

Adds a `mock/` tree (not published — `files: ["src"]`) that boots a serve-only Counterfact instance under the gateway prefix and answers exactly the five lifecycle routes plus auth-exempt `__*` control routes; every other path 404s.

Behaviour modelled:
- **Stateful workspace tree** — a child create writes a workspace linked to its parent; status and family reads reflect it; delete cascades over descendants. Backed by an in-memory store with typed factories (no inline fixtures).
- **Deterministic async** — a `not ready → created` status budget reproduces the consumer's poll loop with no wall clock; settable via a seed or the `__status` control route.
- **Parent-pool quota metering** — a child create / resource transfer draws from the parent allocation; an over-draw returns the gateway's `422`. Default (no pool) is unlimited.
- **Bearer auth** — every real route requires `Authorization: Bearer`, else the live gateway's `401 { detail }`. The guard is injected at one materialization seam and fails closed.

Corrections to the vendored swagger are expressed as an OpenAPI overlay (`spec/overlays/corrections.yaml`, CR1–CR4) applied at generation time — the vendored swagger is never edited. CR1–CR4 are each live-pinned against the real gateway (Bearer scheme; `status` is an object not an array; always-present identity fields marked required; `resources/transfer` returns the updated workspace, not the spec's `WorkspaceResourcesV2`). An overlay-freshness gate fails CI if any correction goes stale against the upstream swagger.

Generated types (`src/generated/types.ts`) and the pydantic models are regenerated for the corrected schema. The published client surface is unchanged behaviourally.

## 4. Required information
- Jira / issue: LLMO-5616
- Other: project-engine-client mock (the reference implementation this mirrors) — https://github.com/adobe/spacecat-shared/pull/1665

- `spacecat-api-service` — runtime consumer of this gateway via the serenity transport (`createSerenityUserManagerApiClient`); this mock is the offline test double for that consumer path. Contract only — no api-service code changes here.
- `spacecat-shared-project-engine-client` — sibling package whose mock this mirrors; shared structure, no shared code.

Every modelled operation was replayed against the **real** User Manager gateway under a funded prod parent, on a throwaway child, with the parent's `family` listing asserted byte-identical before and after (zero residue). This is what pinned CR1–CR4 and the auth/error bodies, and surfaced the `resources/transfer` response-shape divergence (CR4).

The mock was then placed behind a local HTTPS front and driven by spacecat-api-service's **real** consumer logic (`createSerenityTransport` + the ensure/decommission lifecycle) in an A/B against live Semrush: the full activate → re-grant → deactivate sequence behaves identically against mock and live, with unit-level net-zero on the shared parent pool. Throwaway resources were status-polled to `created` before delete and verified fully reclaimed.

## 7. Test plan
- (a) Local end-to-end: a `MOCK_E2E` suite boots `mock/run.js` and drives the real typed client through every op, the deterministic `not ready → created` poll, the `401`, the pool-exhaustion `422`, and the `403`-after-delete. The mock-vs-live A/B against spacecat-api-service's consumer (section 6) passed identically.
- (b) Per-environment: none — the mock is an in-workspace dev dependency, not deployed to any environment. To exercise locally: `npm run generate -w @adobe/spacecat-shared-user-manager-client` then `npm run mock -w @adobe/spacecat-shared-user-manager-client` (serves on `:4010`), or run the e2e via `MOCK_E2E=1 npm run test:e2e -w @adobe/spacecat-shared-user-manager-client`. See `docs/mock-usage.md`.


rainer-friederich (commented):
Hey @rainer-friederich,

> ⚠ Degraded review - no spec document was found for this change. It deliberately mirrors the already-merged project-engine-client mock (the design-of-record is that sibling, not a separate spec), so confidence here is on code-level quality.

**Verdict: Request changes** - a solid, well-tested, live-pinned mirror of the merged sibling; five fidelity/test/hygiene items to fix first. (Posted as a COMMENT because GitHub blocks the PR author from formally requesting changes on their own PR.)
_Changes: adds a stateful Counterfact mock of the User Manager sub-workspace lifecycle gateway - store, ops, seeds, quota, auth, overlay CR1-CR4, freshness gate, e2e (48 files)._

## Must fix before merge
1. [Important] Public-repo doc hygiene - `packages/spacecat-shared-user-manager-client/docs/mock-statefulness.md:42-44` commits the live gateway host (`adobe-hackathon.semrush.com`) and a funded prod workspace UUID ("LLMO-Dev-2", `bb0f4e1c-...`). This repo is public; the value is non-authorizing but it names a concrete live target and internal topology. Redact to placeholders (the team already keeps these in `local/` handover docs); `docs/mock-usage.md` section 9 references the same path.
2. [Important] `mock/counterfact/routes/v2/workspaces/{id}/child.js:25` has no unknown-parent guard - it goes straight to `canAllocate` (true under the default unlimited pool), draws, and creates, while `status.js`, `transfer.js`, and `{id}.js` all return 403 on a missing workspace. A create under an unknown parent succeeds and yields an orphan child that `listFamily` never returns (the walk starts from a root that must exist) - an internally inconsistent state the live gateway would not produce. Add `if (!context.ops.workspaces.exists(path.id))` -> 403 before the quota check, and cover it in the e2e/stateful tests.
3. [Important] The e2e never pins the live 403/422 envelope strings. The 401 body is deep-equal asserted, but `invalid access attempt` (403) and `insufficient available units` (422) are asserted nowhere. The route handlers are coverage-excluded and their raw returns bypass Counterfact response validation, so either string can drift silently - which undercuts the package's live-fidelity thesis. Assert `(await res.json()).message` on one 403 and one 422 case in `test/e2e/user-manager-mock.e2e.js`.
4. [Important] The metering draw-down is never exercised end to end - the only e2e pool case pre-exhausts to `{projects:0,prompts:0}` then asserts 422, proving rejection but not that a successful create/transfer actually persists `quota.draw()`. Because the handlers are coverage-excluded, a handler that forgot the `draw()` call (or drew against the wrong owner) would pass every test. Add an e2e that sets a finite pool, creates a child (succeeds), then asserts the next create 422s, or asserts `GET /__quota` usage decremented.
5. [Important] No CI gate that the committed `src/generated/types.ts` is in sync with `corrections.yaml`. `.github/workflows/user-manager-mock-e2e.yaml` regenerates only the gitignored `build/openapi3.json`; `test:types` passes because the hand-written `src/index.d.ts` and the committed generated file move together. Someone who adds a `CRn` and runs only the documented `spec:convert && spec:overlay` shortcut ships a stale generated type surface. Add `generate:ts` + `git diff --exit-code packages/spacecat-shared-user-manager-client/src/generated/types.ts` to the workflow (it is pure Node, no Python needed).

<details>
<summary>Non-blocking (9): minor issues and suggestions</summary>

- suggestion: header comment says "CR1-CR3" but CR4 was added in this PR - `packages/spacecat-shared-user-manager-client/scripts/apply-overlay.mjs:17`
- suggestion: the `// @ts-check` exception is documented in `CLAUDE.md`, not `mock/` or `docs/mock-usage.md` - `eslint.config.js:84`
- suggestion: `keys()`/`size()` are unused and their JSDoc claims a quota-layer consumer that does not exist - delete them or fix the doc - `packages/spacecat-shared-user-manager-client/mock/store.js`
- nit: `main()` writes the spec back before returning 1 on a stale correction (harmless, gitignored artifact) - `packages/spacecat-shared-user-manager-client/scripts/apply-overlay.mjs`
- suggestion: cascade delete leaves an orphan `workspace_pool` record when a pool-owning parent is deleted - `packages/spacecat-shared-user-manager-client/mock/stateful.js`
- suggestion: bind the mock's control routes to loopback - the auth-exempt `__dump`/`__seed`/`__quota` allow full state read and replace - `packages/spacecat-shared-user-manager-client/mock/run.js`
- nit: workflow comment says `generate:ts` needs Python; only `generate:pydantic` does - `.github/workflows/user-manager-mock-e2e.yaml`
- nit: the `./mock/*` export resolves only via the in-workspace symlink, not the npm tarball (documented/intentional) - `packages/spacecat-shared-user-manager-client/package.json`
- nit: `pickPort` bind-then-release has a TOCTOU window (documented/acceptable for tests) - `packages/spacecat-shared-user-manager-client/test/e2e/user-manager-mock.e2e.js`
</details>

- The mock infrastructure (`store.js`, `run.js`, `inject-auth-guard.js`, `auth.js`) is now copy-pasted across this package and the merged project-engine-client sibling, and the two standalone e2e workflows are near-identical. Before a third client mock, extract a shared counterfact mock-kit (store + runner + auth-guard injection + overlay applier) and convert the workflows to a reusable `workflow_call`. Consciously a follow-up, not a blocker on this PR.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all five blocking findings from the prior review are addressed; the mock is solid.
**Complexity: HIGH** - large diff (48 files); API surface + dependencies + CI workflow.
_Changes: addresses all prior review findings - adds unknown-parent 403 guard, 403/422 envelope assertions, quota draw-down e2e, generated-types drift CI gate, and redacts live workspace ids from docs (1 commit, 9 files changed)._

- Unknown-parent create in `child.js` now returns 403 before quota check (mirrors status/transfer/delete)
- 403/422 envelope strings deep-equal asserted in e2e (pinning the live gateway bodies)
- Quota draw-down exercised end-to-end with a finite pool (proves `draw()` persists)
- Generated-types drift gate added to CI workflow (`generate:ts` + `git diff --exit-code`)
- Live workspace UUID redacted from `docs/mock-statefulness.md` (gateway hostname retained as documented base URL)
- Cascade delete now also drops the parent's `workspace_pool` record (non-blocking, addressed)
- `apply-overlay.mjs` header comment corrected to CR1-CR4

<details>

- suggestion: the `run.js` loopback-bind limitation (Counterfact 2.14.0 hardcodes listen with no host option) is worth a comment in `mock/run.js` so a future upgrader checks whether the option was added - `packages/spacecat-shared-user-manager-client/mock/run.js`
- nit: root CLAUDE.md package entry could mention quota metering and bearer auth alongside the overlay/factory patterns for discoverability - `CLAUDE.md:85`
</details>

byteclimber (approved):

PR #1732, opened 2026-06-26
feat(project-engine-client): publish stateful mock as an HTTPS Docker image (LLMO-5460)

rainer-friederich:
## 1. Abstract
Packages the stateful Project Engine (Semrush) Counterfact mock as a Docker image served over HTTPS, plus a release-tag-triggered workflow that builds and pushes it to GHCR version-locked to the published npm package, and a PR-time workflow that smoke-tests the image build.

## 2. Reasoning
The mock is the fidelity-checked test double for the Semrush Project Engine API, but the published client ships only `src/` (`files: ["src"]`), so consumers cannot get the mock off npm. A cross-repo e2e suite (spacecat-api-service) needs to boot it from its own repo/CI without vendoring the mock source or polluting its production dependency tree. Two hard constraints shaped the design: api-service's transport (`rest-transport.js` `baseUrl()`) throws `503` on any non-`https:` origin, and Counterfact only speaks plain HTTP. Distributing the mock as a runnable HTTPS image solves cross-repo consumption cleanly and is the Semrush vendor mock the cross-repo e2e harness spec depends on.

## 3. High-level overview of the changes
Before: the mock could only be run from inside the package via `npm run mock` (plain HTTP on 4010, in-monorepo). After: the same mock is distributable as `ghcr.io/adobe/spacecat-shared-project-engine-client-mock:<version>`, a self-contained HTTPS service any repo's e2e can `docker run`.

- **Dockerfile** (`node:24-alpine`): full `npm install` (counterfact is a devDep but required at runtime), bakes the gitignored `build/openapi3.json` via `spec:convert` + `spec:overlay` (no Python needed), and mints a self-signed multi-SAN cert (`localhost`, `127.0.0.1`, `project-engine-client-mock`) at build time — no private key is committed to this public repo.
- **Caddy** terminates TLS on `:8443` and reverse-proxies to the plain-HTTP mock on `127.0.0.1:4010`, so api-service's `https:` requirement is satisfied for real, not bypassed.
- **Entrypoint** runs the mock + Caddy and fails fast if either exits; it also waits for the mock to bind `:4010` (bounded ~10s, fail-loud on timeout) before starting Caddy, so proxied requests never hit a 502 in the startup gap. **docker-compose.yml** binds `127.0.0.1` only — the `__*` control routes are unauthenticated and `/__dump` returns the whole store, so the image is loopback-/ephemeral-CI-only by design.
[...]

PR #1735, opened 2026-06-26
feat(user-manager-client): publish stateful mock as an HTTPS Docker image

rainer-friederich:
## 1. Abstract
Packages the stateful User Manager (Semrush sub-workspace lifecycle) Counterfact mock as a Docker image served over HTTPS, plus a release-tag-triggered workflow that builds and pushes it to GHCR version-locked to the published npm package, and a PR-time workflow that smoke-tests the image build. Mirrors the project-engine-client mock Docker packaging for the user-manager-client mock.

## 2. Reasoning
The mock shipped in the base branch (`feat/user_manager`) is the offline test double for the api-service serenity transport's sub-workspace lifecycle, but the published client ships only `src/` (`files: ["src"]`), so consumers cannot get the mock off npm. A cross-repo e2e suite (spacecat-api-service) needs to boot it from its own repo/CI without vendoring the mock source. Two hard constraints shaped the design, identical to the project-engine case: api-service's transport (`rest-transport.js` `baseUrl()`) throws `503` on any non-`https:` origin, and Counterfact only speaks plain HTTP. Distributing the mock as a runnable HTTPS image solves cross-repo consumption cleanly and brings the User Manager gateway to parity with the already-packaged Project Engine gateway mock.

Before: the user-manager mock could only be run from inside the package via `npm run mock` (plain HTTP on 4010, in-monorepo). After: the same mock is distributable as `ghcr.io/adobe/spacecat-shared-user-manager-client-mock:<version>`, a self-contained HTTPS service any repo's e2e can `docker run`.

- **Dockerfile** (`node:24-alpine`): full `npm install` (counterfact is a devDep but required at runtime), bakes the gitignored `build/openapi3.json` via `spec:convert` + `spec:overlay` (no Python needed), and mints a self-signed multi-SAN cert (`localhost`, `127.0.0.1`, `user-manager-client-mock`) at build time — no private key is committed to this public repo.
- **Caddy** terminates TLS on `:8443` and reverse-proxies to the plain-HTTP mock on `127.0.0.1:4010`, so api-service's `https:` requirement is satisfied for real, not bypassed.
- **Entrypoint** runs the mock + Caddy and fails fast if either exits; it waits for the mock to bind `:4010` (bounded ~10s, fail-loud on timeout) before starting Caddy, so proxied requests never hit a 502 in the startup gap. **docker-compose.yml** binds `127.0.0.1` only — the `__*` control routes are unauthenticated and `/__dump` returns the whole store, so the image is loopback-/ephemeral-CI-only by design.
- **Publish workflow** (`user-manager-client-mock-image.yaml`): on each per-package release tag `@adobe/spacecat-shared-user-manager-client-v*` it builds and pushes `:<version>` + `:latest` to GHCR with the workflow's `GITHUB_TOKEN` (`packages: write`); deliberately decoupled from `main.yaml` so an image-build failure can never touch the npm release path. A `workflow_dispatch` (input: version) rebuilds a given version for the first publish / recovery.
- **Smoke workflow** (`user-manager-client-mock-image-smoke.yaml`): the publish workflow only builds at release, so this PR-gated workflow builds the image and smoke-tests a throwaway container end to end through Caddy — `__dump` 200 (TLS + proxy), `GET /v1/workspaces/{parent}/family` without a token → 401 (auth guard), with a token → 200 + the seeded child id (the stateful family walk serving real data, not a path echo) — catching a broken Dockerfile/Caddyfile/entrypoint at PR time instead of at release. Read-only permissions, no push.
- `docker:build` / `docker:run` package scripts and a `docs/mock-docker.md` usage doc (local + CI consumption, the `NODE_TLS_REJECT_UNAUTHORIZED=0` recipe, the loopback security note), plus `CLAUDE.md` / `mock-usage.md` cross-references.

The api-service serenity transport derives both Semrush gateways from the one `SEMRUSH_PROJECTS_BASE_URL` origin (host-only; each prefix is appended internally), so this image is consumed via the same env var as the project-engine mock — just a different image. No change to the published `src/` or any client behaviour — this is purely additive packaging.

## 4. Required information
- Jira / issue: LLMO-5616
- Other: project-engine-client mock Docker packaging this mirrors — https://github.com/adobe/spacecat-shared/pull/1732 ; user-manager mock core (base branch) — https://github.com/adobe/spacecat-shared/pull/1734

- **spacecat-api-service** — consumes the published image in its e2e (contract: `SEMRUSH_PROJECTS_BASE_URL=https://<host>:8443` + `NODE_TLS_REJECT_UNAUTHORIZED=0`, serenity user-manager transport). Not changed here; wiring is a downstream PR.
- **spacecat-shared-project-engine-client** — sibling package whose Docker packaging this mirrors; shared structure, no shared code.

Built and booted the image locally and verified end-to-end against a running container:
- HTTPS `__dump` (auth-exempt control route) returns 200 over TLS through Caddy → TLS terminates and the proxy reaches the mock.
- `GET /v1/workspaces/{parent}/family` without a bearer token returns 401 → the auth guard is wired end to end.
- The same route with a bearer token returns 200 and the body contains the seeded child workspace id → the stateful family walk serves real data through the proxy.
- Plain-HTTP request against `:8443` is rejected (400) → proves real TLS, not cleartext.
- The container's own HEALTHCHECK goes `healthy` within the start period.

## 7. Test plan
(a) Local e2e (done): `npm run docker:build` then booted the container; ran the HTTPS `__dump` / 401-without-token / 200+seeded-child / cleartext-rejection checks in section 6 — all passed.

(b) CI (this PR): `user-manager-client-mock-image-smoke.yaml` builds the image and smoke-tests a throwaway container (TLS + proxy + auth guard + stateful read) on every PR touching the image's inputs — so the build is exercised at PR time, not only at release.

(c) Verification after release (this is a registry artifact, not a deployed service — no dev/stage/prod service env applies):
- On release, `main.yaml` pushes the per-package tag, which triggers `user-manager-client-mock-image.yaml`; confirm `ghcr.io/adobe/spacecat-shared-user-manager-client-mock:<version>` appears and `docker run` boots it. The first image can also be produced immediately via `workflow_dispatch`.
- One-time after first publish: set the GHCR package visibility to public (so local devs and api-service CI pull with no auth); if org policy forbids a public package, keep it private and grant `spacecat-api-service` read on the package.

- Stacked on https://github.com/adobe/spacecat-shared/pull/1734 (base branch `feat/user_manager`, the user-manager mock core) — depends-on: this PR's Dockerfile bakes that mock and its spec/overlay.
- Sequence: merge #1734 to `main` first → rebase/merge this onto `main` → the release tag publishes the first image → a downstream spacecat-api-service PR wires its e2e to the published image.

### DIFF
--- a/.github/workflows/user-manager-client-mock-image-smoke.yaml
+++ b/.github/workflows/user-manager-client-mock-image-smoke.yaml
@@ -1,0 +1,126 @@
+name: User Manager Client Mock Image Smoke
+
+# PR-time guard that the mock Docker image actually BUILDS and BOOTS.
+#
+# The publish workflow (user-manager-client-mock-image.yaml) only runs on the release tag, so
+# without this a broken Dockerfile / Caddyfile / entrypoint would not surface until release — on a
+# tag, where it is hardest to fix. This builds the image and smoke-tests a throwaway container on
+# every PR that touches the package, so the same break is caught on the PR that introduces it.
+#
+# Path-gated (like user-manager-mock-e2e.yaml) so the Docker build cost is only paid when the
+# image's inputs change. Builds locally only — never pushes, no registry/OIDC auth (permissions
+# are read-only), independent of main.yaml's release path.
+
+permissions: {}
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'packages/spacecat-shared-user-manager-client/Dockerfile'
+      - 'packages/spacecat-shared-user-manager-client/Caddyfile'
+      - 'packages/spacecat-shared-user-manager-client/docker-entrypoint.sh'
+      - 'packages/spacecat-shared-user-manager-client/.dockerignore'
+      - 'packages/spacecat-shared-user-manager-client/mock/**'
+      - 'packages/spacecat-shared-user-manager-client/spec/**'
+      - 'packages/spacecat-shared-user-manager-client/scripts/**'
+      - '.github/workflows/user-manager-client-mock-image-smoke.yaml'
+  pull_request:
+    paths:
+      - 'packages/spacecat-shared-user-manager-client/Dockerfile'
+      - 'packages/spacecat-shared-user-manager-client/Caddyfile'
+      - 'packages/spacecat-shared-user-manager-client/docker-entrypoint.sh'
+      - 'packages/spacecat-shared-user-manager-client/.dockerignore'
+      - 'packages/spacecat-shared-user-manager-client/mock/**'
+      - 'packages/spacecat-shared-user-manager-client/spec/**'
+      - 'packages/spacecat-shared-user-manager-client/scripts/**'
+      - '.github/workflows/user-manager-client-mock-image-smoke.yaml'
+  workflow_dispatch:
+
+# One run per PR branch; cancel superseded runs.
+concurrency:
+  group: user-manager-client-mock-image-smoke-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  IMAGE: um-mock-smoke:ci
+  BASE: https://localhost:8443/enterprise/users/api
+  # Seeded by the default MOCK_SEED (parent-with-child): the parent's family listing includes this
+  # child id, so asserting it appears proves the stateful family walk served real data, not an echo
+  # of the path id. Both ids are from mock/seeds.js (SEED_IDS).
+  PARENT_ID: a1b2c3d4-e5f6-4a7b-8c9d-0e1f2a3b4c5d
+  CHILD_ID: b2c3d4e5-f6a7-4b8c-9d0e-1f2a3b4c5d6e
+
+jobs:
+  smoke:
+    name: Build & smoke-test mock image
+    runs-on: ubuntu-latest
+    # Bound a wedged build / a container that never goes healthy.
+    timeout-minutes: 15
+    permissions:
+      contents: read
+    steps:
+      - name: Check out
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: 'false'
+
+      - name: Build image
+        # Same context + Dockerfile the publish workflow uses, so a build break here is the build
+        # break that would otherwise only appear at release. Loads into the local daemon (no push).
+        run: docker build -t "$IMAGE" packages/spacecat-shared-user-manager-client
+
+      - name: Run throwaway container
+        run: |
+          docker run -d --name um-mock-smoke -p 127.0.0.1:8443:8443 "$IMAGE"
+
+      - name: Wait for healthy
+        # The image's own HEALTHCHECK is the readiness contract; poll it instead of re-implementing
+        # a probe. Fail fast (and dump logs) if it never goes healthy.
+        run: |
+          for i in $(seq 1 30); do
+            state=$(docker inspect -f '{{.State.Status}}' um-mock-smoke 2>/dev/null || echo missing)
+            health=$(docker inspect -f '{{.State.Health.Status}}' um-mock-smoke 2>/dev/null || echo missing)
+            echo "state[$i]: $state / health: $health"
+            [ "$health" = "healthy" ] && exit 0
+            # Fail fast on a crashed container instead of sleeping through all 30 iterations.
+            case "$state" in
+              exited|dead) echo "::error::container is $state before becoming healthy"; docker logs um-mock-smoke || true; exit 1 ;;
+            esac
+            [ "$health" = "unhealthy" ] && break
+            sleep 2
+          done
+          echo "::error::container did not become healthy"
+          docker logs um-mock-smoke || true
+          exit 1
+
+      - name: Smoke test — TLS, proxy path, auth guard
+        # Exercises the whole image end to end through Caddy on :8443:
+        #   1. __dump (auth-exempt control route) returns 200            -> TLS terminates + proxy reaches the mock
+        #   2. a real route WITHOUT a bearer token returns 401           -> the auth guard is wired end to end
+        #   3. the same route WITH a bearer token returns 200 + the      -> handler serves real stateful data through
+        #      seeded child id                                              the proxy (the family walk, not a path echo)
+        # -k: the cert is self-signed by design (see docs/mock-docker.md).
+        run: |
+          set -euo pipefail
+
+          echo "1) __dump (auth-exempt) -> 200"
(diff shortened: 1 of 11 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-structured packaging that mirrors the project-engine-client pattern; a few operational improvements worth addressing as follow-ups.
**Complexity: MEDIUM** - medium diff; CI/CD + infrastructure signals.
_Changes: Packages the stateful User Manager mock as a Docker image (Caddy TLS terminator + Counterfact) with a GHCR publish workflow and a PR-time smoke test (11 files)._

## Must fix before merge

No blocking findings.

<details>
<summary>Non-blocking (8): minor issues and suggestions</summary>

- suggestion: Container runs as root - add a non-root user (`USER node` or a custom `mockuser`) before ENTRYPOINT since this is a published image consumed cross-repo. Port 8443 does not require root. - `Dockerfile` (after line 56)
- suggestion: Pin the base image to a digest or versioned Alpine tag (`node:24-alpine@sha256:...`) for reproducible builds. Mutable tags mean rebuilds can silently produce different images. - `Dockerfile:1`
- suggestion: Pin GitHub Actions to full commit SHAs rather than mutable major-version tags (`@v6`, `@v3`) to prevent supply-chain risk in the publish workflow which has `packages: write`. - `.github/workflows/user-manager-client-mock-image.yaml:47-56`
- suggestion: The publish workflow uses `cancel-in-progress: false` but pushes `:latest` - if two tags arrive in quick succession, the slower build wins `:latest` regardless of version order. Either use `cancel-in-progress: true` or drop the `:latest` tag in favor of version-only tags. - `.github/workflows/user-manager-client-mock-image.yaml:19`
- suggestion: After `wait -n` returns, the surviving process is not explicitly killed (the trap only fires on INT/TERM). Adding `trap 'kill 0' EXIT` would ensure both children are reaped on any exit path. - `docker-entrypoint.sh:41`
- suggestion: The smoke workflow path-gate does not include `package.json` - a change to the `scripts` section (e.g. renaming `spec:convert`) could break the image build without triggering the smoke. - `.github/workflows/user-manager-client-mock-image-smoke.yaml:25-33`
- nit: `docker-compose.yml` healthcheck duplicates the Dockerfile HEALTHCHECK with identical config. Compose respects the image's built-in healthcheck by default - consider removing the override to avoid dual-maintenance. - `docker-compose.yml:22-27`
- nit: `provenance: false` disables SLSA attestation on the published image. If this is intentional for plain-manifest compatibility (the comment hints at this), a brief inline comment would clarify intent for future maintainers. - `.github/workflows/user-manager-client-mock-image.yaml:88`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 49s | <b>Cost:</b> $5.86 | <b>Commit:</b> <code>208786dbc6118b1e345ef5f0db76e4c5a04e4bc5</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

alinarublea (approved):

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - new commits cleanly fix the [skip ci] trigger bug and address prior feedback; no blocking issues remain.
**Complexity: HIGH** - large diff; infrastructure + CI/CD signals.
_Changes: Packages the stateful User Manager mock as a Docker image with a reworked release-event publish workflow and PR-time smoke test (11 files)._
Note: Recommend a human read before merge - this change modifies a package-level CLAUDE.md (architectural significance trigger). The bot review is a complement to, not a replacement for, a human read here.

No blocking findings.

<details>

- suggestion: Smoke workflow path-gate does not include `docker-compose.yml` - a broken compose file would not trigger the smoke. Low impact since the smoke builds via bare `docker build` + `docker run` (not compose), and compose is local-dev only. - `.github/workflows/user-manager-client-mock-image-smoke.yaml:25-49`
- nit: CLAUDE.md says "published by ... on each release tag" and the smoke comment (line 11) says "only runs on the release tag" - the workflow actually triggers on `release: published` events, not tag pushes. The distinction matters (the PR itself documents the `[skip ci]` suppression that kills tag-push triggers). Consider "on each release" or "when a release is published". - `packages/spacecat-shared-user-manager-client/CLAUDE.md:284`
- nit: Version extraction error message uses `${RELEASE_TAG:-$VERSION}` which on the `workflow_dispatch` path prints the raw user input with no label. A clearer form: `"FAIL: invalid version format: '$VERSION' (expected X.Y.Z)"`. - `.github/workflows/user-manager-client-mock-image.yaml:224`

</details>

- EXIT trap now reaps both processes on any exit path (c3a1c477)
- Smoke workflow path-gate now includes package.json (c3a1c477)
- Publish workflow trigger rewritten from `on: push: tags:` to `on: release: published`, fixing the latent `[skip ci]` suppression bug (123eaa24)
- `:latest` tag dropped in favor of immutable version-only tags (123eaa24)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 34s | <b>Cost:</b> $2.96 | <b>Commit:</b> <code>123eaa249ad6c3e4b03c91641a6ce21b14c47454</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - trivial doc-wording fix; no logic changes since last approval.
**Complexity: HIGH** - large full diff; infrastructure + CI/CD signals.
_Changes: Corrects "release tag" wording to "when a release is published" in the smoke workflow comment and package CLAUDE.md (2 files, 2 lines)._

No blocking findings. The incremental change since the last review is the doc-wording fix (`0a78d2a8`) addressing the prior nit, plus a merge from main. No new logic or configuration changes.

- Stale "release tag" wording corrected to "when a release is published" in both the smoke workflow comment and CLAUDE.md (0a78d2a8)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 57s | <b>Cost:</b> $1.75 | <b>Commit:</b> <code>bc37858f3fa0f891de101ade56bd3870174d50bc</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1738, opened 2026-06-26
feat(cloudflare-client): tag-based ownership check in deployWorkerScript

nit23uec:
## Summary

- Adds \`tags?: string[]\` option to \`deployWorkerScript\` — tags are written into the Cloudflare Workers upload metadata
- When \`overwrite\` is false (default), a two-step ownership check runs before deploy:
  1. **\`GET /workers/scripts-search?name=<name>&per_page=100\`** — finds the script by name (exact-match filtered client-side; \`name\` param returns partial matches)
  2. If found and \`tags\` provided: **\`GET /workers/scripts/<name>/script-settings\`** — fetches the script's tags and checks for overlap
- Three outcomes when \`overwrite: false\`:
  - Script **not found** → deploy (new script)
  - Script **found, tags match** → skip silently + log info, return \`null\` (idempotent redeploy)
  - Script **found, no tag match** (or no tags provided) → throw error
- \`overwrite: true\` bypasses all checks
- Removed: \`#listWorkers\`, \`#listAllWorkers\`, \`#workerExists\`, \`#isDeployBlocked\`, and the \`allowNotFound\` option in \`#cfFetch\` — no longer needed

## Motivation

Callers stamp their worker with a tag like \`createdBy=adobe\` on first deploy. Subsequent runs with the same tag are idempotent (skip, no error). Deploys from a different owner are blocked. This avoids clobbering another team's worker while still allowing safe redeployment by the same owner.

## API calls per scenario

| Scenario | Calls |
|---|---|
| \`overwrite: true\` | 0 |
| Script not found | 1 (search) |
| Script found, no tags | 1 (search → throw) |
| Script found, tags match | 2 (search + script-settings → skip) |
| Script found, tags differ | 2 (search + script-settings → throw) |

## Test plan

- [x] \`npm test -w packages/spacecat-shared-cloudflare-client\` — 51 tests pass, 100% coverage
- [x] Script not found → deploy proceeds
- [x] Script found, same-owner tag → \`null\` returned + info logged, no PUT
- [x] Script found, different-owner tag → throws
- [x] No tags provided, script exists → throws
- [x] \`overwrite: true\` → no existence check, PUT issued directly
- [x] API errors from search and script-settings propagated correctly

### DIFF
--- a/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
+++ b/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
@@ -113,6 +113,8 @@ export default class CloudflareClient {
    * @param {boolean} [opts.overwrite=false] - Allow replacing an existing script. When false a
    *   GET existence check is performed before upload. This guard is best-effort and not atomic —
    *   a concurrent deploy could create the script between the check and the PUT.
+   * @param {string[]} [opts.tags] - Tags to attach to the Worker script. When provided and the
+   *   script already exists, deploy is only blocked if no tag overlaps (ownership check).
    * @returns {Promise<object>}
    */
   async deployWorkerScript(accountId, scriptName, scriptContent, bindings = [], opts = {}) {
@@ -130,9 +132,10 @@ export default class CloudflareClient {
       compatibilityDate = '2025-01-01',
       observability = true,
       overwrite = false,
+      tags,
     } = opts;
 
-    if (!overwrite && await this.#workerExists(accountId, scriptName)) {
+    if (!overwrite && await this.#isDeployBlocked(accountId, scriptName, tags)) {
       throw new Error(`Worker script '${scriptName}' already exists in account ${accountId}. Set overwrite: true to replace it.`);
     }
 
@@ -141,6 +144,7 @@ export default class CloudflareClient {
       bindings,
       compatibility_date: compatibilityDate,
       ...(observability && { observability: { enabled: true } }),
+      ...(Array.isArray(tags) && tags.length > 0 && { tags }),
     };
 
     const form = new FormData();
@@ -188,6 +192,25 @@ export default class CloudflareClient {
       { method: 'GET', allowNotFound: true },
     );
     return result !== null;
+  }
+
+  async #listWorkers(accountId, { page = 1, perPage = 50 } = {}) {
+    return this.#cfFetch(`/accounts/${accountId}/workers/scripts?page=${page}&per_page=${perPage}`);
+  }
+
+  // Returns true when the deploy should be blocked.
+  // With tags: blocked only when the script exists AND shares none of the provided tags.
+  // Without tags: blocked whenever the script exists.
+  async #isDeployBlocked(accountId, scriptName, tags) {
+    if (Array.isArray(tags) && tags.length > 0) {
+      const workers = await this.#listWorkers(accountId);
+      const existing = workers.find((w) => w.id === scriptName);
+      if (!existing) {
+        return false;
+      }
+      return !existing.tags?.some((t) => tags.includes(t));
+    }
+    return this.#workerExists(accountId, scriptName);
   }
 
   /**
--- a/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
+++ b/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
@@ -193,7 +193,7 @@ export default class CloudflareClient {
   async #findWorker(accountId, scriptName) {
     // name param returns partial matches — filter to exact id.
     const workers = await this.#cfFetch(
-      `/accounts/${accountId}/workers/scripts-search?name=${encodeURIComponent(scriptName)}`,
+      `/accounts/${accountId}/workers/scripts-search?name=${encodeURIComponent(scriptName)}&per_page=100`,
     );
     return workers.find((w) => w.id === scriptName) ?? null;
   }
--- a/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
+++ b/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
@@ -182,7 +182,7 @@ export default class CloudflareClient {
 
   async #listWorkers(accountId, { page = 1, perPage = 50, tags } = {}) {
     const tagFilter = Array.isArray(tags) && tags.length > 0
-      ? `&tags=${tags.map((t) => `${encodeURIComponent(t)}:yes`).join(',')}`
+      ? `&tags=${tags.map((t) => `${encodeURIComponent(t)}:no`).join(',')}`
       : '';
     return this.#cfFetch(
       `/accounts/${accountId}/workers/scripts?page=${page}&per_page=${perPage}${tagFilter}`,
@@ -190,19 +190,16 @@ export default class CloudflareClient {
   }
 
   // Returns true when the deploy should be blocked.
-  // With tags: ask the API for workers matching our tags first — if the script is there,
-  // the same owner is redeploying → allow. If not, fall through to an unfiltered existence
-  // check; if it exists with no matching tag → block.
+  // With tags: query with tag:no — returns workers that do NOT have our tags. If the script
+  // appears there it exists under a different owner → block. If absent, it either has our tag
+  // or doesn't exist at all — both allow the deploy. Single API call.
   // Without tags: blocked whenever the script already exists.
   async #isDeployBlocked(accountId, scriptName, tags) {
-    if (Array.isArray(tags) && tags.length > 0) {
-      const ownedWorkers = await this.#listWorkers(accountId, { tags });
-      if (ownedWorkers.find((w) => w.id === scriptName)) {
-        return false;
-      }
-    }
-    const allWorkers = await this.#listWorkers(accountId);
-    return !!allWorkers.find((w) => w.id === scriptName);
+    const workers = await this.#listWorkers(
+      accountId,
+      Array.isArray(tags) && tags.length > 0 ? { tags } : {},
+    );
+    return !!workers.find((w) => w.id === scriptName);
   }
 
   /**
--- a/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
+++ b/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
@@ -132,17 +132,17 @@ export default class CloudflareClient {
 
     if (!overwrite) {
       if (Array.isArray(tags) && tags.length > 0) {
-        const ownedWorkers = await this.#listWorkers(accountId, { tags });
+        const ownedWorkers = await this.#listAllWorkers(accountId, { tags });
         if (ownedWorkers.find((w) => w.id === scriptName)) {
           this.log.info(`Worker script '${scriptName}' already deployed with a matching tag — skipping`);
           return null;
         }
-        const unownedWorkers = await this.#listWorkers(accountId, { tags, tagsAllowed: false });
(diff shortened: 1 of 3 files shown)


packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
```diff
@@ -181,13 +180,29 @@ export default class CloudflareClient {
...
-    return result !== null;
+  }
+
+  // Returns true when the deploy should be blocked.
+  // With tags: ask the API for workers matching our tags first — if the script is there,
+  // the same owner is redeploying → allow. If not, fall through to an unfiltered existence
+  // check; if it exists with no matching tag → block.
+  // Without tags: blocked whenever the script already exists.
+  async #isDeployBlocked(accountId, scriptName, tags) {
+    if (Array.isArray(tags) && tags.length > 0) {
```
MysticatBot: issue (blocking): `#listWorkers` fetches only page 1 (hardcoded `per_page=50`). `#isDeployBlocked` calls it once with defaults and then does `workers.find()`. If the account has more than 50 worker scripts and the target script is not on page 1, `find()` returns `undefined`, the method returns `false`, and the deploy proceeds - silently bypassing the ownership guard.

This is the exact scenario the feature targets (multi-team accounts with many workers), so the guard needs to work reliably there.

**Fix options (in order of preference):**
1. Use the existing single-script GET (`#workerExists`) to confirm existence first, then fetch the specific script's tags via the list endpoint with a name filter or targeted pagination. This avoids paginating the entire list and keeps the happy-path cost to one API call.
2. Paginate through all pages until the script is found or the list is exhausted (cap at a reasonable upper bound to avoid runaway loops).
3. At minimum, document the 50-script limitation prominently in the JSDoc so callers in large accounts are aware.

MysticatBot (changes requested):
Hey @nit23uec,

**Verdict: Request changes** - the ownership check silently fails for accounts with more than 50 worker scripts.
**Complexity: LOW** - small, single-service feature addition.
_Changes: Adds tag-based ownership checking to `deployWorkerScript` so the same owner can redeploy without `overwrite: true` while still blocking other teams (3 files)._

## Must fix before merge

1. [Important] `#listWorkers` fetches only page 1 (50 results); ownership check silently passes for accounts with >50 workers - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:197` (details inline)

<details>

- nit: Error message says "Set overwrite: true to replace it" even when the block was caused by tag mismatch, which encourages callers to bypass the safety mechanism - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:136`
- suggestion: Add a test case where the existing script has `tags: undefined` or `tags: []` to prove the fail-closed behavior is intentional, not accidental - `test/cloudflare-client.test.js`
- suggestion: Document the `overwrite: true` + `tags` interaction (overwrite skips ownership check; tags still written to metadata) - `src/cloudflare-client.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 47s | <b>Cost:</b> $3.58 | <b>Commit:</b> <code>8bbbf86b5805a3396105f0f9c2549f05a6ef35e9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
```diff
@@ -181,13 +190,15 @@ export default class CloudflareClient {
...
-    // Cloudflare Workers Scripts API supports GET/PUT on this path, not HEAD (returns 405).
-    const result = await this.#cfFetch(
-      `/accounts/${accountId}/workers/scripts/${scriptName}`,
-      { method: 'GET', allowNotFound: true },
+  async #findWorker(accountId, scriptName) {
+    const workers = await this.#cfFetch(
+      `/accounts/${accountId}/workers/scripts-search?search=${encodeURIComponent(scriptName)}`,
     );
-    return result !== null;
+    return workers.find((w) => w.id === scriptName) ?? null;
```
MysticatBot, line 198: issue (blocking): `#findWorker` relies on a single page of results from the `scripts-search` endpoint. If the account has many workers whose names match the search substring and the target falls outside page 1, `.find()` returns `null` and the code concludes the worker does not exist - bypassing the ownership check entirely and deploying over an existing script.

The prior review flagged the same class of problem on the old unfiltered list endpoint. While `scripts-search` narrows the result set significantly (substring match rather than all workers), Cloudflare search endpoints still paginate by default.

**How to fix (pick one):**
- Pass `per_page=200` (Cloudflare maximum) to reduce the chance of truncation, and check for a `result_info.total_count > per_page` condition to throw/warn when results are truncated.
- Fall back to a direct `GET /accounts/{id}/workers/scripts/{name}` (returns 200 or 404) to confirm existence deterministically, then use `script-settings` only for the tag comparison.
- Add an `Array.isArray` guard and a check that the result was not paginated (if the response includes `result_info`).

packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
```diff
@@ -130,17 +127,29 @@ export default class CloudflareClient {
...
+      const found = await this.#findWorker(accountId, scriptName);
+      if (found) {
+        if (Array.isArray(tags) && tags.length > 0) {
+          const settings = await this.#getWorkerSettings(accountId, scriptName);
+          if (tags.some((t) => settings.tags?.includes(t))) {
+            this.log.info(`Worker script '${scriptName}' already deployed with a matching tag — skipping`);
+            return null;
+          }
+        }
+        throw new Error(`Worker script '${scriptName}' already exists in account ${accountId}. Set overwrite: true to replace it.`);
```
MysticatBot, line 143: issue (blocking): When tags are provided but no tag matches (different owner), the error message says `"Set overwrite: true to replace it."` This suggests the caller should use `overwrite: true` as an escape hatch - but doing so would bypass the very ownership guard the tags are meant to enforce, potentially overwriting another team's worker.

Differentiate the two cases:
- No tags provided, script exists: current message is fine (caller owns it, just needs to opt in).
- Tags provided, no match: e.g. `"Worker script '...' already exists in account ... with a different tag set. Will not overwrite."`

The second message should not suggest `overwrite: true` since the tag mismatch signals a different owner.

MysticatBot (dismissed):
Hey @nit23uec,

**Verdict: Request changes** - the search-based existence check may silently bypass ownership on paginated results, and the error message can mislead callers into overwriting a different owner's worker.
**Complexity: MEDIUM** - medium diff, single-service feature addition.
_Changes: Adds tag-based ownership checking to `deployWorkerScript` using the Cloudflare `scripts-search` and `script-settings` endpoints, replacing the old GET-based existence check (3 files)._

## Must fix before merge

1. [Important] `#findWorker` does not handle pagination from `scripts-search` - if the target worker falls outside page 1, the code incorrectly concludes it does not exist and bypasses the ownership check - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:197` (details inline)
2. [Important] Error message on tag-mismatch block suggests `overwrite: true` which would bypass the ownership guard the tags are meant to enforce - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:143` (details inline)

<details>

- suggestion: Add a test for `settings.tags` being undefined/null to lock the fail-closed behavior via optional chaining - `packages/spacecat-shared-cloudflare-client/test/cloudflare-client.test.js`
- suggestion: Restore "best-effort, not atomic" caveat in the JSDoc since the check-then-act pattern still has a TOCTOU window - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:105`
- nit: `tags.some()` uses any-overlap semantics which is correct for the stated use case but the JSDoc should clarify this is intentional (any single matching tag = same owner) - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:138`
- suggestion: Add `if (!Array.isArray(workers)) return null;` guard in `#findWorker` to prevent a cryptic runtime error if the API response shape differs from expectations - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:194`

</details>

- Pagination on `#listWorkers` (50-result page 1 only) - resolved by switching to `scripts-search` with exact-match filter

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 11s | <b>Cost:</b> $6.68 | <b>Commit:</b> <code>bf32740a7b03ed72babc9e0d5b368daa195dc973</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
nit23uec: In both cases, overwrite should behave similarly. this is expected.

packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
nit23uec: script-name is already unique in account. it;'s a false positive.

MysticatBot (approved):
Hey @nit23uec,

**Verdict: Approve** - clean implementation with sound ownership semantics; prior concerns resolved.
**Complexity: MEDIUM** - medium diff, single-service feature addition.
_Changes: Adds tag-based ownership checking to `deployWorkerScript` using Cloudflare `scripts-search` and `script-settings` endpoints, replacing the old GET-based existence check (3 files)._

<details>

- nit: `#getWorkerSettings` failure produces a generic Cloudflare API error ("Cloudflare API returned 500 on ...") rather than a deploy-context message - callers can still distinguish it from "different owner" by the error text, but a wrapped message would be clearer in pipeline logs - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:137`
- nit: `scriptName` is URI-encoded in the query string (`encodeURIComponent` in `#findWorker`) but used raw in path segments (`#getWorkerSettings`, deploy PUT) - this matches the pre-existing pattern in the codebase and Cloudflare rejects malformed names server-side, but the inconsistency is worth noting - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:199`
- suggestion: Add a test where the `script-settings` endpoint returns 500 to lock the error-propagation behavior when tags are provided - `packages/spacecat-shared-cloudflare-client/test/cloudflare-client.test.js`
- suggestion: Add a test where `settings.tags` is undefined (settings response has no tags field) to prove the optional-chaining fail-closed path - `packages/spacecat-shared-cloudflare-client/test/cloudflare-client.test.js`
- suggestion: Add a test where `scripts-search` returns partial-name matches that differ from the exact `scriptName`, proving `.find(w => w.id === scriptName)` filters correctly - `packages/spacecat-shared-cloudflare-client/test/cloudflare-client.test.js`

</details>

- Pagination on `#listWorkers` (50-result page 1 only) - resolved by switching to `scripts-search` with exact-match filter; author correctly noted script names are unique per account
- Error message suggesting `overwrite: true` on tag mismatch - author correctly explained that `overwrite` is an intentional escape hatch and should behave consistently regardless of tag presence

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 40s | <b>Cost:</b> $6.42 | <b>Commit:</b> <code>f963f0e944155abf4d8a8689c95ec02c46ad5270</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1741, opened 2026-06-26
fix(cloudflare-client): filter scripts-search results by script_name not id

ssilare-adobe:
## Summary

- The `scripts-search` API response objects have two distinct fields: `id` (an internal UUID tag like `abb716cad07c40b4b4ddfd0907b3c9d5`) and `script_name` (the worker name used in URLs/routes)
- `#findWorker` was comparing `w.id === scriptName`, which always evaluates to `false` since `id` is never equal to the worker name
- This caused the existence check to always return `null`, so every second deploy silently overwrote the existing worker instead of returning a conflict error (propagated as 409 from the API service)

## Test plan

- [ ] `npm test -w packages/spacecat-shared-cloudflare-client` — all 52 tests pass, 100% coverage
- [ ] Verified against the live Cloudflare API: `scripts-search` response confirmed `id` = `"abb716cad07c40b4b4ddfd0907b3c9d5"` vs `script_name` = `"edge-optimize-router-frescopa-aem-screens-net"` for the same worker

### DIFF
--- a/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
+++ b/packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js
@@ -191,11 +191,13 @@ export default class CloudflareClient {
   }
 
   async #findWorker(accountId, scriptName) {
-    // name param returns partial matches — filter to exact id.
+    // name param returns partial matches — filter to exact script_name.
+    // Note: the result objects use `script_name` for the worker name (used in URLs/routes);
+    // `id` is the script tag identifier and is not the same value.
     const workers = await this.#cfFetch(
       `/accounts/${accountId}/workers/scripts-search?name=${encodeURIComponent(scriptName)}&per_page=100`,
     );
-    return workers.find((w) => w.id === scriptName) ?? null;
+    return (workers ?? []).find((w) => w.script_name === scriptName) ?? null;
   }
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @ssilare-adobe,

**Verdict: Approve** - clean, well-scoped bug fix with appropriate test coverage.
**Complexity: LOW** - small diff, single service.
_Changes: Fixes `#findWorker` to compare against `script_name` instead of `id` when filtering Cloudflare scripts-search results, preventing silent worker overwrites (2 files)._

<details>

- suggestion: Consider normalizing null/undefined list responses in `#cfFetch` itself rather than guarding with `(workers ?? [])` at each call site - prevents the same defensive pattern from spreading to other list-consuming methods. - `packages/spacecat-shared-cloudflare-client/src/cloudflare-client.js:197`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 17s | <b>Cost:</b> $2.34 | <b>Commit:</b> <code>9740c59c0e7a6dcf8b8d28eb04c8df2050400ca7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1743, opened 2026-06-29
build(mock-image): slim Semrush mock Docker images (~711MB to 416MB)

rainer-friederich:
## 1. Abstract
Slims the two Counterfact mock Docker images published from the Semrush client packages (`spacecat-shared-project-engine-client` and `spacecat-shared-user-manager-client`) from ~711 MB to ~416 MB on disk, with identical runtime behaviour.

## 2. Reasoning
The two mock images are consumed for API-level e2e tests in spacecat-api-service. Each was ~719 MB on disk, dominated by an `npm install` of the full build/test toolchain into the runtime image and an in-layer npm cache, even though the mock only needs the Counterfact engine at runtime. This is registry + disk hygiene — not urgent, just wasteful. Both Dockerfiles were identical, so both packages change together.

Each image is now a two-stage build:

- **builder** — full `npm install` (incl. devDeps), runs `spec:convert && spec:overlay` to bake `build/openapi3.json`, and mints the self-signed TLS cert with `openssl`.
- **runtime** — a clean base carrying only what the mock needs to run.

Behaviour deltas:

- **counterfact stays a devDependency** (no `package.json` change), so consumers of the *published* client (`files: ["src"]`) do not inherit it or its ~100 MB toolchain. The runtime stage installs it explicitly from a derived, throwaway runtime-only manifest, pinned to the exact version `package.json` declares.
- Runtime `node_modules` is prod-only (`--omit=dev`); the in-layer npm cache is cleaned and `*.map` / `*.md` / `*.d.ts` are stripped from the dep tree, all in the same layer.
- `openssl` (build-only) and `bash` are no longer installed — the entrypoint is POSIX `sh` and polls the two child PIDs instead of bash's `wait -n`.
- `spec/` and `scripts/` are absent from the runtime image (their only job ran in the builder).
- The base is pinned to the `node:24-alpine` multi-arch index digest for reproducibility.
- The container now runs as the unprivileged `node` user (uid 1000) instead of root.

Net: ~711 MB → ~416 MB (~42%) per image, with the exposed surface, TLS, auth guard, seeds, and control routes all unchanged. The dominant remaining weight is Counterfact itself, which transpiles the materialized handlers at runtime and so legitimately keeps `typescript`/`tsx`/`esbuild` — that is the floor without redesigning the mock.

## 4. Required information
- Jira / issue: https://github.com/adobe/spacecat-shared/issues/1739
- Other: consumed by the serenity API-level e2e suite in https://github.com/adobe/spacecat-api-service/pull/2709 ; original image plumbing in https://github.com/adobe/spacecat-shared/pull/1732 and https://github.com/adobe/spacecat-shared/pull/1735

- `spacecat-api-service` — consumes both GHCR mock images as CI `services:` for its serenity API-level e2e tests. Contract unchanged (same exposed `:8443` HTTPS surface, base URL, seeds, and control routes); this PR only changes how the image is built, so no consumer change is required.

Both images were built locally and validated end to end this session (not just unit tests):

- Built `project-engine-client` and `user-manager-client` images from the new Dockerfiles; both reach the image `HEALTHCHECK` `healthy` state.
- Ran each package's own smoke assertions against the running container over Caddy on `:8443`: `__dump` returns 200 (TLS terminates + proxy reaches the mock), a real route without a bearer token returns 401 (auth guard wired), and the same route with a token returns 200 with seeded data (handler serves real data — exercising Counterfact's runtime handler transpile, which confirms the `*.d.ts` strip is safe).
- Verified the container runs as `uid=1000(node)` and that `bash` / `openssl` are absent.
- Verified the POSIX-`sh` entrypoint exits the container when a child process dies (killed Caddy in-container; the container stopped rather than lingering).
- Measured the size delta with matched single-platform builds: old 711 MB → new 416 MB; the runtime npm layer dropped 314 MB → ~96 MB.

## 7. Test plan
- Local e2e: done (see section 6) — both images build, go healthy, and pass the full smoke flow.
- CI: the existing `*-mock-image-smoke.yaml` workflows build + boot + smoke-test each image on this PR (they path-trigger on the Dockerfile / entrypoint / `.dockerignore` / `mock` / `spec` / `scripts`). spacecat-api-service's serenity IT is the second end-to-end check.
- No environment (eph/dev/stage/prod) deploy applies — these are test-only mock images consumed in CI, never deployed to a runtime environment.

- Independent — no dependent PRs. Note: the slimmed images are republished to GHCR by the existing per-package image workflows on the next release of each client (or on demand via their `workflow_dispatch`); this PR's commit is `build:` and does not itself cut a release.

### DIFF
--- a/packages/spacecat-shared-project-engine-client/.dockerignore
+++ b/packages/spacecat-shared-project-engine-client/.dockerignore
@@ -11,3 +11,4 @@ docs
 python
 *.log
 test-results.xml
+.DS_Store
--- a/packages/spacecat-shared-project-engine-client/Dockerfile
+++ b/packages/spacecat-shared-project-engine-client/Dockerfile
@@ -8,35 +8,36 @@
 # on a non-`https:` scheme), and Counterfact only speaks plain HTTP, so this image runs Caddy as a
 # TLS terminator in front of the mock. TLS is satisfied for real, not bypassed.
 #
-# Build context = THIS package dir. The package has no in-monorepo dependencies (only openapi-fetch
-# + public devDeps incl. counterfact), so a standalone `npm install` resolves everything without
-# the monorepo. `npm run docker:build` invokes this from the package dir.
-FROM node:24-alpine
+# Build context = THIS package dir. The package has no in-monorepo dependencies, so a standalone
+# install resolves everything without the monorepo. `npm run docker:build` invokes this from the
+# package dir.
+#
+# Two stages keep the runtime image lean: the BUILDER carries the full dev toolchain (to convert +
+# overlay the vendored swagger and mint the cert); the RUNTIME stage ships only the mock engine
+# (counterfact), the built spec, the cert, and Caddy/curl — none of the build/test toolchain.
 
-# caddy   — TLS terminator in front of the plain-HTTP mock
-# openssl — mints the self-signed cert at BUILD time (no private key is committed to this public repo)
-# bash    — entrypoint uses `wait -n` (not available in busybox ash)
-# curl    — HEALTHCHECK probe
-RUN apk add --no-cache caddy openssl bash curl
+# Pinned to the multi-arch index digest of node:24-alpine for reproducibility (works on amd64 CI
+# and arm64 local alike). Bump the digest deliberately when upgrading the base.
+FROM node:24-alpine@sha256:a0b9bf06e4e6193cf7a0f58816cc935ff8c2a908f81e6f1a95432d679c54fbfd AS builder
+
+# openssl is BUILD-ONLY: it mints the self-signed cert below, which is then copied into the runtime
+# stage. The runtime stage never needs openssl.
+RUN apk add --no-cache openssl
 
 WORKDIR /app
 
-# Deps first for layer caching. counterfact is a devDependency but is REQUIRED at runtime
-# (mock/run.js spawns its CLI), so a full install — not --omit=dev — is intentional.
-# --legacy-peer-deps: a standalone install (no monorepo lockfile) otherwise ERESOLVE-fails on a
-# build-time-only devDep peer range (openapi-typescript wants typescript@^5, the package pins 6) that
-# the hoisted monorepo install tolerates. openapi-typescript/typescript are unused at runtime.
+# Full install (incl. devDeps) so spec:convert / spec:overlay can run. --legacy-peer-deps: a
+# standalone install (no monorepo lockfile) otherwise ERESOLVE-fails on a build-only devDep peer
+# range (openapi-typescript wants typescript@^5, the package pins 6) that the hoisted monorepo
+# install tolerates. openapi-typescript/typescript are unused at runtime.
 COPY package.json ./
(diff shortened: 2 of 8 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean build optimization with no blocking issues.
**Complexity: HIGH** - medium diff; infrastructure risk surface.
_Changes: Converts two Semrush Counterfact mock Docker images to multi-stage builds, dropping build-only tooling from the runtime layer and switching to unprivileged user execution (8 files)._

<details>

- nit: `trap 'kill 0' INT TERM EXIT` in the entrypoints can interfere with exit status propagation - the EXIT trap fires during `exit "$status"` and sends SIGTERM to the process group. Adding `trap - INT TERM EXIT` before the final `exit "$status"` would make exit code propagation deterministic. - `docker-entrypoint.sh:28`
- suggestion: Add a guard in the `node -e` manifest-derivation one-liner for missing `counterfact` in devDependencies (e.g. `if (!p.devDependencies?.counterfact) process.exit(1)`) - would surface a clear build error instead of a cryptic runtime failure if the field is ever removed. - `Dockerfile:94`
- nit: The inline comment says "LICENSE files are kept (3rd-party license compliance)" but the `find -delete` pattern includes `-name '*.md'` which also deletes `LICENSE.md` files - only extensionless `LICENSE` files actually survive. - `Dockerfile:97`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 45s | <b>Cost:</b> $4.94 | <b>Commit:</b> <code>62337259190c6fdfdfdd36fc47cf4fe81fab4c1a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior findings addressed, no new issues.
**Complexity: HIGH** - medium diff; infrastructure risk surface.
_Changes: Converts two Semrush Counterfact mock Docker images to multi-stage builds, dropping build-only tooling from the runtime layer and switching to unprivileged user execution (8 files)._

- Trap/exit-status propagation: `trap - INT TERM EXIT` now clears traps before the final `exit "$status"`, preventing the EXIT handler from corrupting the propagated exit code.
- Counterfact build guard: `if(!cf){console.error(...); process.exit(1);}` fails the build loudly if counterfact disappears from devDependencies.
- LICENSE.md preservation: `! -iname 'LICENSE*' ! -iname 'COPYING*'` excludes license files (any extension) from the find-delete pattern.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 54s | <b>Cost:</b> $3.69 | <b>Commit:</b> <code>ed00f760c2a00788cece199cc254d06712bce8db</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1744, opened 2026-06-29
fix(project-engine-client): bypass Counterfact 406 on empty-body 2xx mock acks

rainer-friederich:
## 1. Abstract
Makes the Project Engine Counterfact mock return its intended empty-body 2xx ack (publish, batch-delete, update-benchmark) when the request carries `Accept: application/json`, instead of `406 Not Acceptable`.

## 2. Reasoning
The mock 406'd on every empty-body 2xx handler whenever the caller sent `Accept: application/json` — which the serenity transport in spacecat-api-service does on every call. A bare `{ status, body }` return goes through Counterfact's response content-negotiation, and an empty-body `202` declares no `application/json` representation, so Counterfact answered `406` rather than the `202`. This is a mock-fidelity bug, not a production bug: the real Semrush gateway returns the empty ack regardless of `Accept`. The 406 blocked the serenity create/activate API-level IT increment — `createMarket` / `activate` create a draft then `publish`, and the publish step failed upstream with a `502 serenityUpstreamError`.

A new `emptyAck()` helper returns a raw `{ status, body: '', contentType }` envelope. The explicit content type is the signal that makes Counterfact skip response negotiation/validation — the same bypass the mock's auth 401 and quota 405 already rely on. The helper is wired onto the per-request Context as `context.emptyAck`, mirroring how the auth guard is exposed.

Behaviour delta:

- Before: the four empty-body `202` ack handlers (publish, benchmarks batch-delete, benchmark update, brand-urls batch-delete) returned `406` under `Accept: application/json`; `*/*` or no `Accept` got `202`.
- After: all four return `202` with an empty body regardless of `Accept`, matching the live gateway's `content-length: 0` ack.

The three `204 No Content` handlers (project delete, ai_models batch-delete, prompts batch-delete) are intentionally left as bare `{ status: 204 }`: Counterfact special-cases `204` and serves it without negotiating, so they never 406. A `204` carries no body and no `Content-Type`, which the bare return preserves — keeping them closer to the live response than routing them through the helper would.

## 4. Required information
- Jira / issue:  https://github.com/adobe/spacecat-shared/issues/1742
- Other:  blocked IT increment  https://github.com/adobe/spacecat-api-service/pull/2709  · PE mock image  https://github.com/adobe/spacecat-shared/pull/1732

- spacecat-api-service — consumer (contract). Its serenity API-level IT suite boots this mock as a GHCR image; the create/activate lifecycle increment is unblocked once the mock image carrying this fix is published and the tag is bumped there.

Booted the mock locally (`MOCK_SEED=workspace-with-data`) and replayed the issue's reproduction against the seeded workspace/project under `Accept: application/json`:

- publish, benchmarks batch-delete, benchmark PUT, brand-urls batch-delete — each now responds `202` (was `406`); the publish response body is `0` bytes, matching the live `content-length: 0` ack.
- The three `204` delete handlers still respond `204` under the same `Accept: application/json`, confirming they were correctly left untouched.

Validated against the real prod gateway with an IMS prod token (workspace `bfe3c7eb-469f-4331-9b10-7b1319bb6460`), probing the empty-2xx ack path non-destructively via a no-op batch delete (`{"ids": []}`):

- Prod returns `202` with `content-length: 0` and **no `Content-Type` header**.
- The mock matches prod on status and (empty) body, but adds one `Content-Type: application/json` header prod does not send. This is a known, unavoidable Counterfact limitation — a truthy content type is the only way to get a 0-byte body past Counterfact's negotiation; without it the response either `406`s or falls back to a `text/plain` "Accepted" 8-byte body (verified across both the raw-return and typed `$.response[202]` paths). The extra header is invisible to the consumer, which keys off `response.ok`/status. Documented inline in `mock/responses.js`.

## 7. Test plan
- Local: booted the mock and ran the issue's repro `curl` for all four ack endpoints with `Accept: application/json` (results in section 6); also ran the package e2e suite, which drives the real typed client against a live mock and now sends `Accept: application/json` on the raw-fetch ack assertions to pin the regression.
- To verify in the consuming repo: build/publish the mock image from this branch (or the released tag), bump the mock image tag in  https://github.com/adobe/spacecat-api-service/pull/2709 , and run the serenity create/activate IT lifecycle — `POST .../markets` (createMarket) and `POST .../activate` should reach `201` / `200` instead of `502`.

- Related:  https://github.com/adobe/spacecat-api-service/pull/2709  (related) — consumes the fixed mock image; its create/activate increment depends on this.
- Related:  https://github.com/adobe/spacecat-shared/pull/1732  (related) — the release-tag workflow that publishes the mock GHCR image.

Order: merge this PR → semantic-release publishes a new package version and the mock-image workflow publishes a new GHCR tag → bump that tag in the api-service PR to unblock its IT.

### DIFF
--- a/packages/spacecat-shared-project-engine-client/CLAUDE.md
+++ b/packages/spacecat-shared-project-engine-client/CLAUDE.md
@@ -70,10 +70,17 @@ entity through them — `context.factories.createBenchmarkMock({ brand_name, dom
 catalog handlers map their data rows through `createLanguageMock` / `createAiModelMock` /
 `createBrandTopicMock` — rather than emitting an inline literal that the untyped handler (the
 `@ts-check` exception) couldn't catch drifting. The factory is the single, tsc-checked source of
-truth for each shape. This holds for EVERY handler, including the trivial envelope shapes: the
-202 acks build a `createBasicResponseMock`, `getInitStatus` a `createInitStatusMock`, and
-`updateCiCompetitors` maps through `createCiCompetitorMock` — there are no inline-literal
-exceptions left.
+truth for each shape. This holds for EVERY handler that returns an entity, including the trivial
+envelope shapes: `getInitStatus` builds a `createInitStatusMock` and `updateCiCompetitors` maps
+through `createCiCompetitorMock` — there are no inline-literal entity exceptions left.
+
+The empty-body action acks (publish, batch-delete, update-benchmark) are NOT entities — they
+mirror the live gateway's `content-length: 0` ack, so they return `context.emptyAck(202)` from
+`mock/responses.js` rather than a factory. `emptyAck` is a raw `{ status, body: '', contentType }`
+envelope (like `auth.js`'s 401 and `quota.js`'s 405): the explicit `contentType` makes Counterfact
+skip response content-negotiation, which otherwise **406**s an empty-body 2xx under `Accept:
+application/json` (the header the real serenity transport always sends). A bare `{ status: 204 }`
+delete needs no helper — Counterfact serves 204 No Content without negotiating.
 
 ## Spec corrections: the overlay is the single source of truth
 
--- a/packages/spacecat-shared-project-engine-client/mock/context.js
+++ b/packages/spacecat-shared-project-engine-client/mock/context.js
@@ -41,6 +41,7 @@ import { InMemoryStore } from './store.js';
 import { createStatefulOps } from './stateful.js';
 import { createQuota } from './quota.js';
 import { authError } from './auth.js';
+import { emptyAck } from './responses.js';
 import * as factories from './factories.js';
 import { SEEDS, DEFAULT_SEED } from './seeds.js';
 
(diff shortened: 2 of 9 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean, focused mock-fidelity fix following established patterns.
**Complexity: MEDIUM** - medium diff size across 9 files; single-service, no risk signals.
_Changes: introduces `emptyAck()` helper to bypass Counterfact's 406 on empty-body 2xx acks under `Accept: application/json`, pinning regression in E2E tests (9 files)._
Note: Recommend a human read before merge - this change modifies `packages/spacecat-shared-project-engine-client/CLAUDE.md`, a project conventions document. The bot review is a complement to, not a replacement for, a human read here.
Note: CI "Test" check is still pending - verify it passes before merge.

<details>

- nit: second unit test ("carries the content type") is a strict subset of the first (`deep.equal` on the full object already covers `contentType`) - `test/mock/responses.test.js:28`
- suggestion: narrow the `@param` type from `number` to a `200|201|202` union (or add a dev-time assertion) to make the "2xx only" contract self-enforcing - `mock/responses.js:57`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 47s | <b>Cost:</b> $5.26 | <b>Commit:</b> <code>cc52f1eb69139605a0b3d3bc1a186a13b7b688af</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - no new issues; prior nits cleanly addressed.
**Complexity: MEDIUM** - medium diff size across 9 files; single-service, no risk signals.
_Changes: addresses both prior non-blocking items (narrowed JSDoc type union, removed redundant test) on the emptyAck mock-fidelity fix (9 files)._
Note: Recommend a human read before merge - this change modifies `packages/spacecat-shared-project-engine-client/CLAUDE.md`, a project conventions document. The bot review is a complement to, not a replacement for, a human read here.
Note: CI "Test" check is still pending - verify it passes before merge.

- Redundant "carries the content type" test removed; deep.equal on the full object already covers contentType.
- JSDoc @param type narrowed from `number` to `200 | 201 | 202` literal union, making the empty-body-2xx-only contract tsc-enforced.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 28s | <b>Cost:</b> $2.20 | <b>Commit:</b> <code>6e1c568fce3cb25c8fb3b330af0e7001209da5e8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1748, opened 2026-06-29
feat(data-access): status transition guard + deriveSuggestionStatus (SITES-47091)

alinarublea:
## What & why (SITES-47091)

Adds the JS-side enforcement primitives for the Suggestion ↔ FixEntity status lifecycle defined in ADR [adobe/mysticat-architecture#174](https://github.com/adobe/mysticat-architecture/blob/main/platform/decisions/design-suggestion-fix-entity-status-lifecycle.md). Today `spacecat-shared` validates `status` only as a closed enum — any value from any value, no state machine. This PR adds the transition table + guard + the shared bubble-up function, **shipped warn-only by default**.

## Changes (spacecat-shared-data-access only)

- **Transition tables + predicates** — `fix-entity.transitions.js` (canonical FixEntity table from the ADR), `suggestion.transitions.js` (intentionally **V1-permissive**; the warn logs inform tightening before enforce).
- **`util/status-transition-guard.js`** — `getEnforcementMode()` reads `STATUS_TRANSITION_ENFORCEMENT` ∈ `off | warn | enforce`, **default `warn`**. `guardTransition()`: allowed/no-op → pass; `warn` → log violation (`Entity id from -> to`) and proceed; `enforce` → throw `ValidationError`; `off` → skip.
- **`setStatus` override + `transitionStatus` alias** on FixEntity and Suggestion models — a model-defined `setStatus` blocks the auto-generated setter (`base.model.js:128`), so this is a single chokepoint capturing **every** JS writer (api-service single PATCH, `bulkUpdateStatus`, autofix-worker). Default `warn` is byte-equivalent to the old setter (`isReference` was always `false` for `status`).
- **`deriveSuggestionStatus`** — the **non-CWV 1:1** bubble-up only (DEPLOYED/PUBLISHED→FIXED, FAILED→ERROR, PENDING→IN_PROGRESS, ROLLED_BACK→SKIPPED; severity collapse for >1 fix).
- Barrel + `.d.ts` exports.

## Rollout / behavior

Default `warn` ⇒ **zero behavior change** on consumer bump. Consumers adopt and observe violation logs, then flip to `enforce` — tracked in **SITES-47286**. This guard covers only the two JS writers; mystique (Python direct-DB) bypasses it and needs a separate guard later.

## Deferred (agreed, split out)

The **CWV multi-issue bubble-up** is deferred to **SITES-47285**: the ADR's per-issue vocabulary (`cwvIssueStatus` = `PATCH_*`/`GUIDANCE_GENERATED`) exists only in mystique, written into the overloaded per-issue `status` field via direct-DB, while the JS `ISSUE_STATUSES` enum lists only the Suggestion vocabulary. `deriveSuggestionStatus` throws if a non-empty `issues[]` is passed until that's reconciled. `FixEntity.REJECTED` is likewise omitted (not in the enum yet).

## Tests

2441 data-access tests pass; all new files at 100% coverage; lint clean.

Tracking: SITES-47091 · parent SITES-47076 · follow-ups SITES-47285, SITES-47286.


MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - well-designed state machine guard with a safe warn-only rollout strategy.
**Complexity: HIGH** - large diff (977 lines, 16 files).
_Changes: Adds status transition tables, a tri-mode enforcement guard, setStatus overrides on FixEntity and Suggestion models, and deriveSuggestionStatus bubble-up function (16 files)._

<details>

- nit: Duplicate status constants across `fix-entity.transitions.js`, `suggestion.transitions.js`, and `derive-status.js` create a sync hazard when the enum grows - `packages/spacecat-shared-data-access/src/models/fix-entity/fix-entity.transitions.js:17`. The circular-import rationale is sound, but consider extracting to a leaf constants module before flipping to `enforce` mode.
- suggestion: `transitionStatus(to)` is a pure pass-through to `setStatus` with no distinct behavior or documented guidance on when to prefer it - `packages/spacecat-shared-data-access/src/models/fix-entity/fix-entity.model.js:63`. Either document the intended semantic split or defer the alias until behavior diverges.
- suggestion: `deriveSuggestionStatus` silently returns `null` when a fix entry is `null`/`undefined` or has an unrecognized status. Consider adding a test pinning this edge-case behavior (e.g. `deriveSuggestionStatus([null])` returns `null`) so the contract is explicit - `packages/spacecat-shared-data-access/src/models/suggestion/derive-status.js:90`.

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 37s | <b>Cost:</b> $3.91 | <b>Commit:</b> <code>44c7c12e0de4811f4d6828be9e1704812c5620ba</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-data-access/src/models/suggestion/index.js
```diff
@@ -28,3 +28,12 @@ export {
...
 } from './suggestion.data-schemas.js';
+
+// Suggestion status transition table + predicate, and the 1:1 bubble-up
+// (SITES-47091). CWV multi-issue bubble-up deferred (see derive-status.js).
+export {
+  SUGGESTION_TRANSITIONS,
+  SUGGESTION_CREATE,
+  isAllowedSuggestionTransition,
+} from './suggestion.transitions.js';
+export { deriveSuggestionStatus } from './derive-status.js';
```
MysticatBot, line 39: issue (blocking): The new exports here (`deriveSuggestionStatus`, `SUGGESTION_TRANSITIONS`, `SUGGESTION_CREATE`, `isAllowedSuggestionTransition`) have no corresponding type declarations in `index.d.ts`. Same gap exists in `fix-entity/index.d.ts` for `FIX_ENTITY_TRANSITIONS`, `FIX_ENTITY_CREATE`, `isAllowedFixTransition`.

TypeScript consumers importing these will get implicit `any` or module-resolution failures depending on their tsconfig strictness.

**Fix:** Add declarations for the exported functions and constants to both `.d.ts` files. Example for suggestion:

```typescript
export declare const SUGGESTION_CREATE: unique symbol;
export declare const SUGGESTION_TRANSITIONS: Record<string | symbol, string[]>;
export declare function isAllowedSuggestionTransition(from: string | null | undefined, to: string): boolean;
export declare function deriveSuggestionStatus(fixes: Array<object | string>, issues?: Array<object>): string | null;
```

packages/spacecat-shared-data-access/src/models/suggestion/derive-status.js
```diff
@@ -0,0 +1,103 @@
...
+  if (Array.isArray(issues) && issues.length > 0) {
+    throw new Error('deriveSuggestionStatus: CWV multi-issue bubble-up is not yet implemented (deferred to SITES-47076)');
+  }
+
+  if (!Array.isArray(fixes) || fixes.length === 0) {
+    return null;
+  }
+
+  const statuses = fixes.map(toStatus);
+
```
MysticatBot, line 97: issue (blocking): The `toStatus` helper returns `undefined` for null/undefined array entries or objects without a `status` property. When this happens, the resulting `statuses` array contains `undefined` entries that match none of the `includes()` checks and fail the `every()` check, causing the function to silently return `null`.

A caller passing `[validFix, null]` (e.g., from a sparse junction query result) gets `null` back with no indication anything went wrong - indistinguishable from "no fixes exist."

**Fix:** Filter out undefined/null after mapping:

```javascript
const statuses = fixes.map(toStatus).filter((s) => s != null);
if (statuses.length === 0) return null;
```

This degrades gracefully (consistent with the empty-array behavior) while preventing a single malformed entry from nullifying an otherwise well-defined derivation.

packages/spacecat-shared-data-access/src/util/status-transition-guard.js
```diff
@@ -0,0 +1,75 @@
...
+  if (from === to) {
+    return;
+  }
+  if (isAllowed(from, to)) {
+    return;
+  }
+
+  const mode = getEnforcementMode();
+  if (mode === ENFORCEMENT_MODES.OFF) {
+    return;
```
MysticatBot, line 60: issue (blocking): In `enforce` mode, the `ValidationError` message includes the entity ID and current status: `status transition violation: FixEntity <uuid> PENDING -> PUBLISHED`. If this propagates to the HTTP response (ValidationError typically maps to 400 in the API layer), it discloses the entity's existence and current state to the caller.

This is an information oracle - an attacker probing the PATCH endpoint with arbitrary IDs and status values learns whether an entity exists and what state it is in.

**Fix:** Keep the detailed message for the server-side log, throw a sanitized message:

```javascript
if (log) { log.warn(message); }
throw new ValidationError(`Invalid status transition for ${entityName}: cannot transition to ${to}`);
```

This preserves debuggability (the warn log fires regardless of mode) while keeping the thrown error safe for external consumption.

MysticatBot (changes requested):
Hey @alinarublea,

**Verdict: Request changes** - three fixable issues before this is ready for consumers to adopt.
**Complexity: HIGH** - large diff (977 lines, 17 files); state machine enforcement across model layer.
_Changes: Adds status transition tables, a tri-mode enforcement guard (off/warn/enforce), setStatus overrides on FixEntity and Suggestion models, and deriveSuggestionStatus bubble-up function (17 files)._

## Must fix before merge

1. [Important] Missing TypeScript declarations for newly exported public API - `packages/spacecat-shared-data-access/src/models/suggestion/index.d.ts:28` (details inline)
2. [Important] `deriveSuggestionStatus` silently returns null when fix entries contain undefined status - `packages/spacecat-shared-data-access/src/models/suggestion/derive-status.js:88` (details inline)
3. [Important] ValidationError in enforce mode leaks entity ID and current status to callers - `packages/spacecat-shared-data-access/src/util/status-transition-guard.js:62` (details inline)

<details>

- nit: `guardTransition` evaluates `isAllowed` before checking mode - when mode is `off`, the check is wasted work. Consider moving the mode check above the `isAllowed` call if `off` becomes a hot path. - `src/util/status-transition-guard.js:49`
- suggestion: Add a drift-detection test that asserts transition table keys cover all values in the model's STATUSES enum - catches sync drift at CI time rather than in production. - `src/models/fix-entity/fix-entity.transitions.js:44`
- suggestion: Consider structured log objects (`{ event, entityName, entityId, from, to }`) instead of template strings for better Splunk queryability during the warn period. - `src/util/status-transition-guard.js:60`
- nit: CLAUDE.md references SITES-47285 for CWV deferral, but the code JSDoc references SITES-47076 - reconcile the ticket numbers. - `packages/spacecat-shared-data-access/CLAUDE.md:33`

</details>

- Prior review approved with 3 non-blocking suggestions (duplicate constants, transitionStatus alias, null-return contract). All remain unresolved but were non-blocking; no regression.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 22s | <b>Cost:</b> $5.69 | <b>Commit:</b> <code>5352a7d9cdab2a3cb839d5d21ada2a9a8279a89d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - all prior blocking findings addressed; no new blocking issues.
**Complexity: HIGH** - large diff (977 lines, 17 files); API surface signal.
_Changes: Adds status transition tables, a tri-mode enforcement guard (off/warn/enforce), setStatus overrides on FixEntity and Suggestion models, and deriveSuggestionStatus bubble-up function (17 files)._
Note: Recommend a human read before merge - this change modifies project documentation (CLAUDE.md) that defines operational contracts for the package. The bot review is a complement to, not a replacement for, a human read here.

<details>

- suggestion: Add a drift-guard test for `deriveSuggestionStatus` that asserts every `FixEntity.STATUSES` value is either mapped or explicitly documented as unmapped, same pattern as the existing transition-table drift guards - `packages/spacecat-shared-data-access/src/models/suggestion/derive-status.js:109`
- suggestion: Pin the mixed `['ROLLED_BACK', 'UNKNOWN_STATUS']` scenario in `derive-status.test.js` to document the null-return behavior when unrecognized statuses prevent the `every()` check from passing - `packages/spacecat-shared-data-access/test/unit/models/suggestion/derive-status.test.js`
- nit: `to` parameter is reflected unsanitized in the thrown ValidationError message; low risk (not XSS, the API layer wraps errors), but truncating to 30 chars or stripping non-alphanumeric would harden it against log injection - `packages/spacecat-shared-data-access/src/util/status-transition-guard.js:67`
- suggestion: Consider structured log objects (`{ event: 'status_transition_violation', entityName, entityId, from, to }`) instead of template strings for better Splunk queryability during the warn rollout period - `packages/spacecat-shared-data-access/src/util/status-transition-guard.js:63`

</details>

- TypeScript declarations added for all newly exported public APIs (both index.d.ts files)
- `deriveSuggestionStatus` now filters null/undefined entries before deriving status
- ValidationError in enforce mode now throws a sanitized message (entity ID and current status only in server-side log.warn)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 28s | <b>Cost:</b> $5.47 | <b>Commit:</b> <code>36193aa09505196fc27a01715725253b009401d2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ramboz (approved):

tkotthakota-adobe (approved):
Added few comments.

packages/spacecat-shared-data-access/src/models/suggestion/derive-status.js
```diff
@@ -0,0 +1,109 @@
...
+ * @param {Array<object>} [issues] - CWV per-issue data; MUST be empty for now
+ * @returns {string|null} derived Suggestion status, or null if there are no fixes
+ * @throws {Error} if a non-empty `issues` array is supplied (CWV deferred)
+ */
+export const deriveSuggestionStatus = (fixes, issues = []) => {
+  if (Array.isArray(issues) && issues.length > 0) {
+    throw new Error('deriveSuggestionStatus: CWV multi-issue bubble-up is not yet implemented (deferred to SITES-47285)');
+  }
+
+  if (!Array.isArray(fixes) || fixes.length === 0) {
```
sandsinh, line 86: We can also pass the current status of the suggestion to deriveSuggestionStatus and if there are no fixes the current status of the suggestion becomes the derived suggestion status.

packages/spacecat-shared-data-access/src/models/suggestion/suggestion.transitions.js
```diff
@@ -0,0 +1,83 @@
...
+  // re-detection can reopen a fixed suggestion (-> NEW), retry (-> IN_PROGRESS), or outdate it.
+  [S.FIXED]: [S.NEW, S.IN_PROGRESS, S.ERROR, S.OUTDATED],
+  // errors are recoverable: re-attempt or reopen.
+  [S.ERROR]: [S.NEW, S.IN_PROGRESS, S.FIXED, S.SKIPPED, S.OUTDATED],
+  // un-skip / re-detect.
+  [S.SKIPPED]: [S.NEW, S.OUTDATED],
+  // re-detection reopens an outdated suggestion.
+  [S.OUTDATED]: [S.NEW],
+  // terminal.
+  [S.REJECTED]: [],
```
sandsinh: We can relax S.REJECTED being terminal to allow S.NEW, accounts for incorrect classification and reopen of the issues.

PR #1749, opened 2026-06-29
fix(tokowaka-client): real browser UA for verifyRouting human probe

ABHA61:
## What
The human leg of `verifyRouting` (CloudFront Edge Optimize verify) used a bare `Mozilla/5.0`. Bot-detection can misclassify that as a bot, making the human probe look bot-like and the verify step **falsely fail** (it requires the human leg to NOT be routed to Edge Optimize).

## Change
- Default the human probe to a realistic desktop Chrome User-Agent.
- Make it overridable via an optional `humanUa` option (`verifyRouting(url, { humanUa })`), default-preserving for existing callers.
- The bot probe (`chatgpt-user`) is unchanged.

## Tests
Added two cases (default browser UA is used for the human probe; custom `humanUa` is honored). `verifyRouting` suite green (11 passing); eslint clean.

## Consumer
`@adobe/spacecat-api-service` calls `verifyRouting(url)` and inherits the new default after this is published + the dependency is bumped.


MysticatBot (approved):
Hey @ABHA61,

**Verdict: Approve** - clean, well-scoped bug fix with tests; no blocking issues.
**Complexity: LOW** - small diff, single package, no risk signals.
_Changes: Replaces the bare `Mozilla/5.0` User-Agent in the human probe of `verifyRouting` with a realistic Chrome UA string and makes it overridable via an options parameter (2 files)._

<details>

- nit: Hardcoded `Chrome/124.0.0.0` will age silently - bot-detection systems that flag outdated UA versions could eventually trigger the same false-failure this PR fixes. Consider a comment noting the vintage (e.g. `// Last updated 2024-Q2 - bump periodically`) so future maintainers know when to refresh. - `packages/spacecat-shared-tokowaka-client/src/cdn/cloudfront/index.js:1202`
- suggestion: Both new tests rely on `fetchStub.secondCall` (positional). If the implementation ever reorders the `Promise.all` array or adds an intermediate fetch, both tests silently verify the wrong call. A more robust assertion would find the call whose `user-agent` header is NOT `chatgpt-user`. - `packages/spacecat-shared-tokowaka-client/test/cdn/cloudfront/index.test.js:2048`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 18s | <b>Cost:</b> $2.32 | <b>Commit:</b> <code>c4e7c40a4d6f4d3c03ff168466eddd55699031d8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-tokowaka-client/src/cdn/cloudfront/index.js
```diff
@@ -116,13 +160,16 @@ const delay = (ms) => new Promise((resolve) => {
  * @param {string} params.externalId - external ID baked into the connector role trust policy.
  * @param {string} [params.roleName] - connector role name (defaults to the standard name).
  * @param {string} [params.region] - STS region.
+ * @param {string} [params.operator] - operator identity (email); used as the RoleSessionName so the
+ *   customer's CloudTrail attributes each mutation to the person who ran it.
  * @returns {Promise<{roleArn: string, accountId: string, credentials: object}>}
  */
```
MysticatBot, line 166: issue (blocking): The empty catch swallows ALL exception classes (TypeError, credential expiration, throttling, SDK config errors) without any observable signal. If `TagResourceCommand` systematically fails (wrong Tags shape, missing permission boundary, invalid ARN), the ownership model silently degrades from "tag + Comment marker" to "Comment marker only" and operators have no way to detect it without manually inspecting CloudFront resource tags in the customer account.

The Comment marker IS a sufficient fallback for ownership decisions, so this is not a safety bug. But silent, invisible failure in production code running in customer accounts is an observability gap that should not ship without at minimum a log line.

**Fix:** Replace the empty catch with a warn/debug-level log:
```javascript
} catch (e) {
  log.debug(`best-effort tagging of ${functionArn} failed: ${e.message}`);
}
```
If no logger is in scope, at minimum narrow the catch to the expected `AccessDeniedException` and re-throw unexpected errors.

packages/spacecat-shared-tokowaka-client/src/cdn/cloudfront/index.js
```diff
@@ -116,13 +160,16 @@ const delay = (ms) => new Promise((resolve) => {
  * @param {string} params.externalId - external ID baked into the connector role trust policy.
```
MysticatBot, line 160: issue (blocking): The `assertOwnedOrThrow` for the Lambda function (further down at line ~978) runs AFTER the Pending/InProgress early-return. When a non-owned Lambda is still initializing (State=Pending or LastUpdateStatus=InProgress), the function returns `{ status: "provisioning" }` without ever checking ownership. On the next poll (once Active), it throws.

The caller sees "provisioning" (implying our deploy is in progress) then a hard ownership error on the next call. This is confusing UX and a potential false-positive in orchestration that treats "provisioning" as "we are making progress."

The IAM role IS correctly guarded before this point, so safety is preserved (we never publish code into a non-owned function). But the behavioral inconsistency should be resolved.

**Fix:** Move the ownership check (`assertOwnedOrThrow` on `cfg.Role`) above the Pending/InProgress early-return, or document explicitly in a comment why skipping the guard in that state is acceptable (e.g. the role check already ran, the Lambda cannot be published into during Pending).

MysticatBot (changes requested):
Hey @ABHA61,

**Verdict: Request changes** - two observability/consistency issues in the new ownership-guard logic.
**Complexity: HIGH** - large diff (941 lines); ownership infrastructure + operator threading + UA fix in one PR.
_Changes: Adds resource ownership tags and guards (ManagedBy/createdBy) across CloudFront, Lambda, and IAM resources to prevent reusing non-owned infrastructure, threads operator identity for CloudTrail attribution, and fixes the verifyRouting human probe User-Agent (2 files)._
_Scope note: title says "real browser UA for verifyRouting human probe" but the diff also delivers ownership tagging, resource reuse guards, and operator identity threading - `packages/spacecat-shared-tokowaka-client/src/cdn/cloudfront/index.js`._
Note: Recommend a human read before merge - possible divergence from the documented design (heuristic, flagged by the conventions reviewer). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

1. [Important] Silent catch block in best-effort tagging hides all errors without any log signal - `packages/spacecat-shared-tokowaka-client/src/cdn/cloudfront/index.js:522` (details inline)
2. [Important] Lambda ownership check is bypassed when function is in Pending/InProgress state, causing confusing provisioning-then-throw sequence for non-owned Lambdas - `packages/spacecat-shared-tokowaka-client/src/cdn/cloudfront/index.js:974` (details inline)

<details>

- nit: Duplicate consecutive `if (existing)` blocks in createOrigin should be merged into one - the guard and the self-heal logic share the same condition and separating them invites confused maintenance. - `packages/spacecat-shared-tokowaka-client/src/cdn/cloudfront/index.js:373`
- nit: Hardcoded `Chrome/124.0.0.0` in DEFAULT_HUMAN_UA will age silently. Bot-detection that flags outdated UAs could eventually trigger the same false-failure this PR fixes. The override escapes this, but consider a code comment noting when to refresh (e.g. annually). - `packages/spacecat-shared-tokowaka-client/src/cdn/cloudfront/index.js:1297`
- suggestion: `verifyRouting` tests rely on positional `fetchStub.secondCall` ordering. If the `Promise.all` array is ever reordered or a retry wrapper added, the tests verify the wrong probe. Find the call by UA header content instead. - `packages/spacecat-shared-tokowaka-client/test/cdn/cloudfront/index.test.js:2293`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 50s | <b>Cost:</b> $6.21 | <b>Commit:</b> <code>18f498e3ed85f1a0aa76b2e0dd0624b62b9837d0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1752, opened 2026-06-30
feat(project-engine-client): make project AIO tags a stateful mock resource

rainer-friederich:
## 1. Abstract
Makes project-level AIO **tags** a stateful resource in the Project Engine Counterfact mock (`spacecat-shared-project-engine-client`): `POST /aio/tags` now persists, and new `GET` and `DELETE` handlers let a standalone `category:<name>` tag be read back and removed.

## 2. Reasoning
The mock did not persist project tags — `POST /aio/tags` synthesized a `201` array without writing to the store, and there was no `GET` handler. So a standalone category created via `createProjectTags` (a category that has no prompts yet) could not be listed back and disappeared on refresh. This blocked end-to-end verification of the new Categories management surface in elmo, where the UI had to fall back to showing a freshly-added category optimistically because the mock couldn't round-trip it.

Adds `tags` as a per-project stateful resource, scoped `tags:{workspaceId}:{projectId}` so the same category registered across multiple market projects keeps an independent collection per project (the multi-market model: one category name is registered on N projects via N `createProjectTags` calls).

Behaviour deltas:
- **`POST /aio/tags`** — before: returned a synthesized array, wrote nothing. After: persists each tag idempotently (deterministic `tag-<name>` id, so re-creating the same name reuses the stored tag, matching the live reuse-by-name) and returns the live top-level `TreeNodeResponse` array.
- **`GET /aio/tags`** — new. Returns the project's stored tags as `AIOTagsListResponse`, so a 0-prompt category reads back. The spec-required `parent_id` / `search` query params are honoured (a request missing either `400`s, matching live); a non-empty `search` filters by case-insensitive name substring; `parent_id` is accepted but not used to filter (the mock's tags are a flat per-project collection).
- **`DELETE /aio/tags`** — new. Removes standalone tags by id so an orphaned 0-prompt project tag can be cleaned up. Prompts that carry a removed tag are a separate collection and stay fully intact — the tag delete never cascades to prompts.
- The deterministic tag-id derivation is now shared via `context.tagId` so the two routes that mint tag ids (`POST /aio/tags` and `POST /aio/prompts/tagged`) can't drift out of the cross-endpoint id contract that lets the Categories surface correlate a standalone tag with the same tag attached to a prompt.
- Spec overlay correction (CR5): `model.AIOTag.id` is marked required (a tag always has an id live), with the generated TypeScript types and the Python model regenerated to match. This is the only change to the published client surface.

Note: a mock GHCR image republish (on the next `@adobe/spacecat-shared-project-engine-client` release, per `project-engine-client-mock-image.yaml`) is required for the local-dev / integration-test stacks to pick this up.

## 4. Required information
- Jira / issue: adobe/spacecat-shared issue 1751 — https://github.com/adobe/spacecat-shared/issues/1751
- Other: related mock-fixture-fidelity issue — https://github.com/adobe/mysticat-data-service/issues/757

- `spacecat-api-service` — consumes the mock through the serenity transport `createProjectTags` (`POST /aio/tags`); the new `GET /aio/tags` is what unblocks end-to-end verification of the `POST /v2/orgs/:org/brands/:brand/serenity/tags` Categories endpoint. Contract: the persisted tag id and the id embedded in prompt tags now share one derivation.
- `mysticat-data-service` / local-dev + integration-test stacks — consume the containerized mock (GHCR image); they pick up this change only after the mock image is republished on the next package release.

## 7. Test plan
(a) Local end-to-end: exercised the change against a live Counterfact mock booted by the package runner, driving the real generated client through the full tag spine — create-then-`GET` round-trip of a 0-prompt category, idempotent re-create by name, `search` substring filtering, `__reset` lifecycle (created tags cleared back to the seed baseline), two-project multi-market isolation (a tag created on project A does not surface on project B in the same workspace), and `DELETE` of a project tag while the seeded prompt remains listable. All round-tripped as expected.

(b) Per-environment: this is a shared-library mock, not deployed infrastructure — there is no dev/stage/prod deploy for the mock itself. The only published-client delta is the regenerated types (`model.AIOTag.id` now required). For the local-dev / integration-test stacks to consume the new stateful behaviour, republish the mock GHCR image on the next `@adobe/spacecat-shared-project-engine-client` release (`project-engine-client-mock-image.yaml`), then verify a standalone category created via `createProjectTags` reads back through `GET /aio/tags` in the offline serenity local-dev stack.

### DIFF
--- a/packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/tags.js
+++ b/packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/tags.js
@@ -56,7 +56,9 @@ export function GET($) {
   const items = search
     ? stored.filter((t) => String(t.name).toLowerCase().includes(search))
     : stored;
-  return $.response[200].json({ items, page: query?.page ?? 1, total: items.length });
+  // `page` arrives as a query string (e.g. "2"); coerce so the response field stays the numeric
+  // type AIOTagsListResponse declares, regardless of whether the param was passed.
+  return $.response[200].json({ items, page: Number(query?.page ?? 1), total: items.length });
 }
 
 /** DELETE — remove standalone project tags by id (prompts are untouched) → 204 No Content. */
--- a/packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/tags.js
+++ b/packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/tags.js
@@ -11,21 +11,60 @@
  */
 
 /**
- * Handler for POST /v2/workspaces/{id}/projects/{project_id}/aio/tags — creates project-level
- * AIO tags (the consumer's `createProjectTags`). Request is `TreeNodeListRequest` `{ names }`.
- * The live API returns a TOP-LEVEL ARRAY of `TreeNodeResponse` `[{ id, name, children_count,
- * keyword_count }]` (verified 2026-06-25; overlay CR6 retypes the 201 as an array). A
- * deterministic id per name keeps a repeated create idempotent in the mock's view. Excluded
- * from coverage (materialized handler).
+ * Stateful handlers for /v2/workspaces/{id}/projects/{project_id}/aio/tags — the project-level AIO
+ * tag taxonomy (the Categories surface). The per-project `tags` collection (`tags:{ws}:{pid}`) is
+ * scoped so the same category registered across N market projects keeps N independent collections.
+ * Materialized into `.counterfact/routes/` by the mock runner; excluded from coverage.
+ *
+ * - POST (`createProjectTags`): request `TreeNodeListRequest` `{ names }`; persists each tag
+ *   idempotently (deterministic `tag-<name>` id, so a repeated create reuses the stored tag,
+ *   matching the live reuse-by-name) and returns the live shape — a TOP-LEVEL ARRAY of
+ *   `TreeNodeResponse` `[{ id, name, children_count, keyword_count }]` (verified 2026-06-25;
+ *   overlay CR6 retypes the 201 as an array).
+ * - GET (`aio-get-project-tags`): returns the project's stored tags as `AIOTagsListResponse`
+ *   `{ items, page, total }`, so a 0-prompt category created via POST reads back. The spec marks
+ *   `parent_id` + `search` as required query params (request validation 400s a request missing
+ *   either, matching live); the mock's tags are a flat per-project collection, so `parent_id` is
+ *   accepted but not used to filter, while a non-empty `search` filters by case-insensitive name
+ *   substring.
+ * - DELETE (`aio-delete-tags`): removes the body's tag ids (`BatchDeleteRequest` `{ ids }`) from
+ *   the standalone tag collection → 204. Prompts that carry a removed tag are a separate
+ *   collection and stay intact (the spec's `prompt_id` query param is accepted but not load-bearing
+ *   here — the mock's project-tag delete targets the standalone collection, never cascading to
+ *   prompts).
  */
 
-/** Deterministic tag id from a name so repeated creates under the same name share an id. */
-const tagId = (name) => `tag-${encodeURIComponent(name)}`;
-
-/** POST — create project tags → 201 array of TreeNodeResponse. */
+/** POST — create (persist, idempotent) project tags → 201 array of TreeNodeResponse. */
 export function POST($) {
-  const { body, context } = $;
+  const { path, body, context } = $;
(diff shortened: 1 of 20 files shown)


MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - one mock-fidelity gap to close before merge.
**Complexity: HIGH** - large diff (819 lines, 20 files).
_Changes: Makes project-level AIO tags a stateful resource in the Counterfact mock, adding POST persistence, GET listing, and DELETE support scoped per project (20 files)._
Note: CI checks are currently pending - confirm green before merge.

## Must fix before merge
1. [Important] No e2e test verifies GET `/aio/tags` returns 400 when required query params (`parent_id`/`search`) are omitted - `packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/tags.js` (details inline)

<details>

- nit: GET response `page` field is uncoerced query string ("2") vs numeric default (1) - inconsistent types depending on whether the param is passed. Consider `Number(query?.page ?? 1)`. - `packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/tags.js`
</details>

packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/tags.js
```diff
@@ -11,21 +11,60 @@
...
-  const tags = names.map((name) => context.factories.createTagNodeMock({ id: tagId(name), name }));
+  const stored = context.ops.tags.upsertMany(
+    scope,
+    names.map((name) => context.factories.createAIOTagMock({ id: context.tagId(name), name })),
+  );
+  const tags = stored.map((t) => context.factories.createTagNodeMock({ id: t.id, name: t.name }));
   return $.response[201].json(tags);
 }
+
+/** GET — list the project's stored tags (optional `search` name filter) → 200 list response. */
```
MysticatBot, line 50: issue (blocking): The PR description, JSDoc, and `docs/mock-usage.md` all state that omitting `parent_id` or `search` from the GET query returns 400, matching the live API. This contract is enforced by Counterfact's spec-level request validation (the vendored OpenAPI marks them `required`), but no e2e test verifies it works for this new endpoint.

The existing `getBrandTopics` endpoint has a dedicated 400 test (raw `fetch` bypassing the typed client). Adding the same pattern here would close the loop and catch a regression if the spec annotation is ever dropped.

Suggested fix: add a test using raw `fetch` against the endpoint without `parent_id`/`search` and assert the 400 status, mirroring the `getBrandTopics` 400 test pattern.

rainer-friederich: Fixed in `61fff9ff` — added `getProjectTags 400s when a required query param is missing`, mirroring the `getBrandTopics` 400 test (raw `fetch`, omit `parent_id`/`search`, assert 400 + the param name in the message).

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean implementation, prior findings addressed, no blocking issues.
**Complexity: HIGH** - large diff (838 lines, 20 files); API surface.
_Changes: Makes project-level AIO tags a stateful resource in the Counterfact mock with POST persistence, GET listing, and DELETE support scoped per project (20 files)._
Note: Recommend a human read before merge - this change modifies project-level documentation gathered as architectural context (package CLAUDE.md, docs/mock-statefulness.md). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: GET handler `total` field equals post-filter `items.length`, not the unfiltered collection size. If the live API returns an unfiltered total when `search` is applied (the common pagination pattern), the mock diverges in a way a consumer test might eventually notice. No current consumer keys off the difference. - `packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/tags.js`
- nit: DELETE returns raw `{ status: 204 }` while POST/GET use `$.response[201].json(...)` / `$.response[200].json(...)`. A bare `{ status: 204 }` is accepted by Counterfact for 204 No Content (documented in the package CLAUDE.md as needing no helper), so this is consistent with the project convention for empty-body deletes - just visually inconsistent within the file. - `packages/spacecat-shared-project-engine-client/mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/tags.js`
</details>

- GET /aio/tags 400 test for missing required query params now present (61fff9ff)
- GET page field coerced with `Number(query?.page ?? 1)` (61fff9ff)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 7m 30s | <b>Cost:</b> $6.07 | <b>Commit:</b> <code>61fff9ff1da2a85713d6c5f6114eea8051191e27</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1755, opened 2026-06-30
feat(http-utils): surface a ReBAC session flag on the no-resource defer path

ravverma:
## What

`facsWrapper` now sets `context.attributes.facs = { enabled, product, subjectId }` on the **no-resolvable-resource defer path** — the branch where the session is FACS-enrolled but the route carries no single ReBAC resource to enforce against (its path param is the org, not a resource — e.g. `GET /organizations/:id/sites`, `GET /v2/orgs/:id/brands`).

## Why

These are **collection endpoints**: there is no single resource for the wrapper to resolve and allow/deny. The wrapper already defers them to the controller. This change lets the controller know the session is ReBAC-enrolled so it can **filter the collection** to the resources the caller may view — without re-implementing LD-flag/enrollment logic in the consuming service.

## Semantics (who gets the flag)

Reaching the defer point means all of: not an internal identity, not an internal/exempt org, the per-product **LD flag is ON**, and the **JWT short-circuit did not fire** (caller lacks the org-wide capability). So:

| Session | Flag set? | Controller behavior |
|---|---|---|
| admin / S2S / internal-org / LD-off | no | full collection (unfiltered) |
| org-wide JWT capability holder | no (short-circuits at step 5) | full collection |
| enrolled, resource-scoped user | **yes** | filter to viewable resources |

Purely additive — no consumer reads the flag yet. `facs-wrapper.js` stays at 100% coverage (478 tests pass).

### DIFF
--- a/packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js
+++ b/packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js
@@ -378,6 +378,24 @@ export function facsWrapper(fn, { routeFacsCapabilities } = {}) {
     });
 
     if (!resource) {
+      // Reaching here means the session is FACS-enrolled (passed the internal,
+      // internal-org and LD-flag gates) and resource-scoped (the JWT short-circuit
+      // at step 5 did not fire, so the caller lacks the org-wide capability), yet
+      // there is no single ReBAC resource to enforce against. Surface that to the
+      // controller via `context.attributes.facs` so collection endpoints (whose
+      // path param is the org, not a resource — e.g. list-sites, list-brands) can
+      // ReBAC-filter their results to the resources the caller may view. Callers
+      // that bypass earlier (admin / internal org / LD-off) or short-circuit on an
+      // org-wide JWT capability never set this flag, so those sessions see the
+      // unfiltered collection — the controller treats an absent flag as "no filter".
+      // `context.attributes` is always present here: reaching this point past the
+      // per-product LD-flag gate required a tenant (`orgId`), read from `authInfo`
+      // on `context.attributes`.
+      context.attributes.facs = {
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - purely additive, well-scoped, tested on the path it modifies.
**Complexity: MEDIUM** - small diff; auth signal.
_Changes: Sets `context.attributes.facs` on the no-resolvable-resource defer path so collection-endpoint controllers can ReBAC-filter results without re-implementing enrollment logic (2 files)._

<details>

- suggestion: Freeze the surfaced object with `Object.freeze({ enabled, product, subjectId })` to communicate read-only intent and prevent downstream accidental mutation - `packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js:393`
- suggestion: Add a negative assertion (`expect(context.attributes.facs).to.be.undefined`) to at least one bypass/short-circuit test case (admin, internal-org, or org-wide capability path) to lock in the "absent flag means no filter" contract - `packages/spacecat-shared-http-utils/test/auth/facs-wrapper.test.js`
- nit: `subjectId` carries what appears to be an email address in tests; document whether this is expected to be a stable user GUID or a mutable claim like email, since consumers may later use it as a cache key or policy binding - `packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js:395`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 10s | <b>Cost:</b> $2.25 | <b>Commit:</b> <code>33534597460ca0a9004f825afa7b06d7fc4d8469</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1761, opened 2026-07-01
feat(utils): add isPathPatternWithinSiteScope url helper

dhavkuma:
## Summary
- Adds `isPathPatternWithinSiteScope(pathPattern, siteBaseUrl)` to `spacecat-shared-utils`, mirroring `isWithinSiteScope`'s `siteBaseUrl` argument, to determine whether a path pattern is confirmed to be scoped under a site's base path.
- A pattern is trusted only when it's a literal path (no regex metacharacters) that equals or is nested under the base path, or when it uses the `/*` pathname-prefix convention (matching `@adobe/spacecat-shared-tokowaka-client`'s `buildUrlMatcher`). Anything else fails closed, since such patterns are compiled as a regex and matched against the full URL downstream, making a literal `startsWith` check on the pattern text unsound.

## Test plan
- [x] `npm test -w packages/spacecat-shared-utils` — 1254 passing, 100% line/statement coverage
- [x] `npm run lint -w packages/spacecat-shared-utils`

### DIFF
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -595,6 +595,43 @@ export function allHaveSamePathname(urls, referenceUrl) {
   return urls.every((url) => hasSamePathname(url, referenceUrl));
 }
 
+/**
+ * A pattern is trusted as scoped to the site's base path only when it is an exact match or
+ * uses the `/*` pathname-prefix convention (see @adobe/spacecat-shared-tokowaka-client's
+ * buildUrlMatcher). Any other pattern is compiled as a regex and matched against the full
+ * URL downstream, so a literal `startsWith` check on the pattern text is not sound —
+ * e.g. `/kings/.*|/wolves/.*` starts with `/kings` but, once compiled, also matches
+ * `/wolves/...` via the `|` alternation. We fail closed for anything we cannot verify.
+ * @param {string} pathPattern - the pattern to check.
+ * @param {string} siteBaseUrl - the site's base URL defining the scope (e.g. "bulk.com/uk"),
+ * mirroring the siteBaseUrl argument of @adobe/spacecat-shared-utils's isWithinSiteScope.
+ * @returns {boolean} true if the pattern is confirmed to be within the site's scope.
+ */
+export function isPathPatternWithinSiteScope(pathPattern, siteBaseUrl) {
+  if (typeof pathPattern !== 'string') {
+    return false;
+  }
+  if (!siteBaseUrl) {
+    return true;
+  }
+
+  const siteBasePath = toPathname(siteBaseUrl);
+
+  if (siteBasePath === '/') {
+    return true;
+  }
+
+  if (pathPattern.endsWith('/*')) {
+    const prefix = pathPattern.slice(0, -2);
+    return prefix === siteBasePath || prefix.startsWith(`${siteBasePath}/`);
+  }
+  // eslint-disable-next-line no-useless-escape
+  if (/[|()\[\]^$+*?\\.]/.test(pathPattern)) {
+    return false;
+  }
+  return pathPattern === siteBasePath || pathPattern.startsWith(`${siteBasePath}/`);
+}
+
 export {
   ensureHttps,
   getSpacecatRequestHeaders,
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @dhavkuma,

**Verdict: Approve** - well-implemented security boundary with correct fail-closed design.
**Complexity: LOW** - small utility addition, single function.
_Changes: Adds `isPathPatternWithinSiteScope` helper that verifies path patterns are scoped to a site's base path, failing closed on unverifiable patterns (4 files)._

<details>

- suggestion: The `/*` branch returns before the metacharacter regex check runs, so a pattern like `/uk/x|bar/*` would have its prefix evaluated by `startsWith` alone. Since downstream `buildUrlMatcher` treats `/*` patterns as prefix matches (not regex), this is not a bypass today, but adding a metacharacter check on the prefix would be defense-in-depth against future routing changes. - `packages/spacecat-shared-utils/src/url-helpers.js:624`
- suggestion: The metacharacter regex does not include curly braces. A pattern like `/uk/page{1,3}` passes the check and is treated as literal. In practice this cannot cause a scope escape (quantifiers repeat existing atoms, not path prefixes), but including them would fully align with the "fail closed for anything we cannot verify" principle. - `packages/spacecat-shared-utils/src/url-helpers.js:628`
- nit: The `eslint-disable-next-line no-useless-escape` suppression exists because `\.` inside a character class does not need escaping. Removing the backslash (using `.` bare inside `[]`) eliminates the need for the suppression since `\[` and `\]` are required escapes that will not trigger the rule. - `packages/spacecat-shared-utils/src/url-helpers.js:627`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 49s | <b>Cost:</b> $3.85 | <b>Commit:</b> <code>dcdf60997f3b52126d0f5b45d5ab042671fe25cb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -595,6 +595,44 @@ export function allHaveSamePathname(urls, referenceUrl) {
...
+ */
+export function isPathPatternWithinSiteScope(pathPattern, siteBaseUrl) {
+  if (typeof pathPattern !== 'string') {
+    return false;
+  }
+  if (!siteBaseUrl) {
+    return true;
+  }
+
+  const siteBasePath = toPathname(siteBaseUrl);
```
dhavkuma, line 619: `toPathname` lower-cases its result, so `siteBasePath` here ends up case-folded — but `isWithinSiteScope`, the sibling function this one is meant to stay consistent with, is case-sensitive on the pathname:

```js
toPathname('Example.com/Kings')                                            // '/kings'
isWithinSiteScope('https://Example.com/kings/page', 'Example.com/Kings')   // false
```

Concrete effect: for a site scoped to `https://example.com/Kings` (mixed case), `isPathPatternWithinSiteScope('/kings/malicious/*', 'example.com/Kings')` returns `true` because the base path is lower-cased to `/kings` before comparing — even though `isWithinSiteScope` would reject an actual URL at that same casing for this site. The two checks disagree on the same input, which undermines the "fail closed for anything we cannot verify" contract this function documents for itself.

Suggest deriving `siteBasePath` the same case-preserving way `isWithinSiteScope` does internally (`new URL(prependSchema(siteBaseUrl)).pathname`, trailing slash stripped) instead of via `toPathname`, plus a test case with a mixed-case base path.

dhavkuma: Correction to my own comment above: case sensitivity isn't actually a requirement here, so the fix direction should go the other way.

Rather than making `isPathPatternWithinSiteScope`'s path comparison case-sensitive to match `isWithinSiteScope`, `isWithinSiteScope` itself should be relaxed to be case-insensitive on the path (matching `toPathname`'s existing lower-casing convention, which most of this module already follows — `hasSamePathname`/`allHaveSamePathname` are explicitly case-insensitive). URL paths being scoped by casing is not a real constraint for this guard, and keeping `isWithinSiteScope` case-sensitive is what creates the mismatch, not the other way around.

So: this PR's use of `toPathname(siteBaseUrl)` in `isPathPatternWithinSiteScope` is fine as-is. The actual follow-up is a separate change to `isWithinSiteScope` (in `url-helpers.js`) to lower-case/normalize both sides of its pathname comparison instead of comparing raw case-sensitive pathnames. Filing that as a follow-up rather than blocking this PR on it.

PR #1762, opened 2026-07-02
fix(facs): bypass FACS enforcement for IMS auth channel

ravverma:
## Problem

On a FACS-governed route, an **IMS-authenticated** caller (`authInfo.getType() === 'ims'`) hit the tenant gate at step (4):

```js
if (!orgId) {
  return forbidden('Tenant required for FACS evaluation');
}
```

Direct IMS tokens are a deprecating channel — they never carry `facs_permissions` and may not resolve a tenant `orgId` from `getTenantIds()`. So FACS cannot meaningfully evaluate them, yet the ladder was 403-ing **every org request** from an IMS session on a FACS route.

## Fix

Add an explicit **step 1b** bypass, right after the internal-identity bypass: when `authType === 'ims'`, skip the wrapper and call the handler unchanged.

Rationale: IMS orgs that ARE FACS/RBAC-enabled are gated ahead of this wrapper, so this bypass only admits the not-yet-enrolled IMS traffic that must keep working for now. JWT (the recommended channel, which carries `facs_permissions`) is unaffected and still fully enforced.

## Changes

- `facs-wrapper.js` — new `(1b)` IMS bypass block + updated bypass-rules doc comment.
- `facs-wrapper.test.js` — test asserting an IMS session with no tenant bypasses (before the tenant gate) and logs `bypass: 'ims-auth-channel'`.

## Verification

- `facs-wrapper.js` coverage: **100% / 100%**.
- Package suite: **479 passing**, lint clean.

## Follow-up

Once published, bump `@adobe/spacecat-shared-http-utils` in `spacecat-api-service` to pick this up.

### DIFF
--- a/packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js
+++ b/packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js
@@ -126,6 +126,12 @@ function routeMatchesAnyProductMap(context, productsRoutes) {
  *
  *   0. CORS preflight (`OPTIONS`) → bypass.
  *   1. Internal identities (admin, S2S, read-only admin, api-key) → bypass.
+ *   1b. IMS auth channel (`authInfo.getType() === 'ims'`) → bypass. Direct IMS
+ *      tokens are a deprecating channel that never carry `facs_permissions` and
+ *      may not resolve a tenant `orgId`, so FACS cannot evaluate them; forcing
+ *      them through the ladder would 403 every org request. IMS orgs that ARE
+ *      FACS/RBAC-enabled are gated ahead of this wrapper, so this bypass only
+ *      admits the not-yet-enrolled IMS traffic that must keep working for now.
  *   2. Adobe internal IMS orgs (`FACS_EXCEPTION_INTERNAL_ORGS`) → bypass.
  *   3. Route NOT in any product map → bypass.
  *      Route IS in some product map but `x-product` is missing / mismatched
@@ -234,6 +240,19 @@ export function facsWrapper(fn, { routeFacsCapabilities } = {}) {
         isS2SConsumer: !!authInfo?.isS2SConsumer?.(),
         isReadOnlyAdmin: !!authInfo?.isReadOnlyAdmin?.(),
       }, 'FACS bypass: internal identity');
+      return fn(request, context);
+    }
+
+    // (1b) IMS auth channel bypass. Direct IMS tokens are a deprecating channel
+    // that never carry `facs_permissions` and may not resolve a tenant `orgId`,
+    // so FACS cannot meaningfully evaluate them — pushing them through the ladder
+    // would 403 every org request at the tenant gate. IMS orgs that ARE
+    // FACS/RBAC-enabled are gated ahead of this wrapper; this bypass only admits
+    // the not-yet-enrolled IMS traffic that must keep working for now.
+    if (authType === 'ims') {
+      log.info({
+        tag: 'facs', bypass: 'ims-auth-channel', method, suffix, authType,
+      }, 'FACS bypass: IMS auth channel');
       return fn(request, context);
     }
 
--- a/packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js
+++ b/packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js
@@ -126,12 +126,15 @@ function routeMatchesAnyProductMap(context, productsRoutes) {
  *
  *   0. CORS preflight (`OPTIONS`) → bypass.
  *   1. Internal identities (admin, S2S, read-only admin, api-key) → bypass.
- *   1b. IMS auth channel (`authInfo.getType() === 'ims'`) → bypass. Direct IMS
- *      tokens are a deprecating channel that never carry `facs_permissions` and
- *      may not resolve a tenant `orgId`, so FACS cannot evaluate them; forcing
- *      them through the ladder would 403 every org request. IMS orgs that ARE
- *      FACS/RBAC-enabled are gated ahead of this wrapper, so this bypass only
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js
```diff
@@ -237,6 +243,19 @@ export function facsWrapper(fn, { routeFacsCapabilities } = {}) {
       return fn(request, context);
     }
 
+    // (1b) IMS auth channel bypass. Direct IMS tokens are a deprecating channel
+    // that never carry `facs_permissions` and may not resolve a tenant `orgId`,
+    // so FACS cannot meaningfully evaluate them — pushing them through the ladder
+    // would 403 every org request at the tenant gate. IMS orgs that ARE
```
MysticatBot, line 252: issue (blocking): The bypass fires on `authType === 'ims'` with no secondary signal confirming the caller genuinely cannot be FACS-evaluated. The upstream `isOrgRbacEnabled` gate (IMS handler) is fail-open by design - it returns `false` on LD client errors or when the `x-product` header is missing/unrecognized. During an LD outage, an enrolled IMS caller with a resolvable tenant would previously have continued through the FACS evaluation ladder (steps 2-7). After this change, step 1b fires first and skips the entire ladder.

This is a narrow regression path (LD outage + enrolled IMS org + FACS-governed route), but it does widen the permission surface compared to pre-PR behavior in that scenario.

**Suggested fix:** Add a defensive guard so the bypass only fires when FACS truly cannot evaluate the session:
```js
if (authType === 'ims' && !authInfo?.getFacsPermissions?.()?.length) {
```
This keeps the bypass scoped to sessions that carry no FACS claims (the stated intent) and makes it self-contained rather than dependent on the upstream gate holding.

packages/spacecat-shared-http-utils/test/auth/facs-wrapper.test.js
```diff
@@ -235,6 +235,23 @@ describe('facsWrapper', () => {
...
+  describe('IMS auth channel bypass', () => {
+    it('bypasses for ims auth type before the tenant gate', async () => {
+      // No tenant on the IMS session — without the bypass this would 403 at the
+      // tenant gate for every org request.
+      context.attributes.authInfo = makeAuthInfo({
+        getType: () => 'ims',
+        getTenantIds: () => [],
+      });
+      const wrapped = facsWrapper(handler, { routeFacsCapabilities });
+      await wrapped({}, context);
```
MysticatBot, line 248: suggestion: The test exercises the motivating scenario (empty tenant list), but the bypass is unconditional on `authType === 'ims'`. A second case with a populated tenant would document that intent explicitly and catch accidental reordering if someone moves this check below the tenant gate:
```js
it('bypasses for ims auth type even when a tenant is present', async () => {
  context.attributes.authInfo = makeAuthInfo({
    getType: () => 'ims',
    getTenantIds: () => ['SOME-ORG@AdobeOrg'],
  });
  const wrapped = facsWrapper(handler, { routeFacsCapabilities });
  await wrapped({}, context);
  expect(handler.calledOnce).to.be.true;
});
```

MysticatBot (changes requested):
Hey @ravverma,

**Verdict: Request changes** - the bypass relies on an upstream invariant that is fail-open during LD outages, creating a narrow regression path.
**Complexity: MEDIUM** - small diff; auth signal.
_Changes: Adds an IMS auth-channel bypass to the FACS wrapper so direct IMS tokens (which cannot carry `facs_permissions`) are no longer 403'd at the tenant gate (2 files)._

## Must fix before merge

1. [Important] Unconditional bypass depends on fail-open upstream gate; enrolled IMS callers skip FACS during LD outages - `packages/spacecat-shared-http-utils/src/auth/facs-wrapper.js:249` (details inline)

<details>

- suggestion: Add a test case where `getTenantIds` returns a populated array to document that the bypass is unconditional and guard against reordering regressions - `packages/spacecat-shared-http-utils/test/auth/facs-wrapper.test.js:247`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 10m 47s | <b>Cost:</b> $4.44 | <b>Commit:</b> <code>763724a22d8a0df566275dd4de78002b8fd51c14</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - prior blocking finding addressed cleanly; the guarded bypass is correct and well-tested.
**Complexity: MEDIUM** - small diff; auth signal.
_Changes: Adds a FACS-claims guard to the IMS auth-channel bypass so only sessions with no `facs_permissions` skip the FACS ladder, addressing the prior fail-open regression concern (2 files)._

- Unconditional IMS bypass depending on fail-open upstream gate - now guarded with `!authInfo?.getFacsPermissions?.()?.length`
- Missing populated-tenant test case - added, plus a negative test for sessions carrying FACS claims

<details>

- suggestion: The negative test ("does NOT take the ims bypass when the session carries FACS claims") asserts the bypass log was absent but not the terminal outcome. Adding one assertion for what *did* happen (handler called, or 403 returned) would make the test self-documenting about end-to-end behavior of a FACS-claimed IMS session - `packages/spacecat-shared-http-utils/test/auth/facs-wrapper.test.js:285`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 6s | <b>Cost:</b> $4.01 | <b>Commit:</b> <code>70af9d36d74fee29ce744d9ee41975bb8932857a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #765, opened 2025-05-26
feat: Support fetching only unbranded keywords from ahrefs

davidaurelio:
Adds an `excludeBranded` option to `AhrefsAPIClient#getOrganicKeywords` to exclusively retreive unbranded keywords from Ahrefs.

This is required for Project Elmo

PR #1763, opened 2026-07-02
feat(data-access): mapping-table site_id/deletedAt + Brand workspace-id rename

rainer-friederich:
## 1. Abstract

Two changes to the data-access layer: (a) adds `siteId`/`deletedAt` attributes to the `BrandSemrushProject` model and a new `allByOrganizationId` accessor to its collection; (b) flips the `Brand` model's write-of-record for the Semrush sub-workspace id from `semrushWorkspaceId` to `semrushSubWorkspaceId`, keeping the old attribute as read-only for backward compatibility.

## 2. Reasoning

`brand_to_semrush_projects` is being kept and actively maintained in Semrush sub-workspace mode rather than retired, and its schema is gaining `site_id`/`deleted_at` columns (mysticat-data-service, see "Required information" below) so a mapping row can carry a soft-delete tombstone and a link to its market-mirror Site. This package's model layer needs to expose both new columns, and cross-team consumers (starting with the Brand Presence / URL Inspector Elements API wrappers) need a single per-organization read path instead of fanning out one `allByBrandId` call per brand.

Separately, in the same review thread: `brands.semrush_workspace_id` actually holds the brand's **sub-workspace** id (the parent workspace lives on `organizations.semrush_workspace_id`), and the shared name confuses every new reader. The companion mysticat-data-service migration (#765) flips which underlying column is write-of-record via a DB trigger; this PR makes the model layer match — `semrushSubWorkspaceId` becomes the writable attribute, `semrushWorkspaceId` becomes read-only. The old attribute is *not* removed: it's kept indefinitely so BC readers of the deprecated field keep working, per the maintainer's explicit "keep the old column" call.

**`BrandSemrushProject` (mapping table):**
- Gains `siteId` (nullable UUID) and `deletedAt` (nullable ISO date, soft-delete tombstone — modeled the way `ApiKey.deletedAt` already is, since this package has no package-level soft-delete scope) attributes, both auto-mapping to their snake_case DB columns.
- New collection method `allByOrganizationId(organizationId, { includeDeleted })`: one PostgREST embedded-join query (INNER JOIN on the `brands` FK) per organization, filtering tombstones by default. Returns plain frozen identity DTOs (brand/project/slice/site plus the embedded sub-workspace id under the corrected `semrushSubWorkspaceId` name), not model instances — deliberately, so a consumer can only read identity through it.
- Tombstone contract, made explicit in the collection's class doc: every other accessor on this collection (including the auto-generated `allByBrandId` / `findBySemrushProjectId`) still returns tombstoned rows alongside live ones. Only `allByOrganizationId` filters by default.

**`Brand` (write-of-record flip):**
- `semrushSubWorkspaceId` now carries the `hasText` validator and is the writable attribute (setter generated); `semrushWorkspaceId` is now `readOnly: true` at the schema level — matches the reversed DB trigger direction in the companion migration.
- To keep this fully backward compatible for any external caller of the old setter (schema-level `readOnly` alone would be a semver-breaking API removal), `brand.model.js` defines a manual, deprecated `setSemrushWorkspaceId()` that delegates to `setSemrushSubWorkspaceId()`. No consumer-visible method disappears.
- `index.d.ts` updated: `getSemrushWorkspaceId()` documented as deprecated, `setSemrushSubWorkspaceId(value)` added, `setSemrushWorkspaceId(value)` kept and marked `@deprecated`.
- No behaviour change for existing readers of either attribute — both still resolve to the same (DB-trigger-mirrored) value; only which one accepts writes changed.

## 4. Required information

- Spec: https://github.com/adobe/serenity-docs/blob/docs/brand-semrush-mapping-maintenance/docs/discovery/brand-semrush-mapping-maintenance.md (§7.2 the sanctioned accessor, §10 the companion rename decision)
- Implementation plan: https://github.com/adobe/serenity-docs/blob/docs/brand-semrush-mapping-maintenance/docs/discovery/brand-semrush-mapping-maintenance-implementation.md
- Other: https://github.com/adobe/mysticat-data-service/pull/765 (companion schema migration), https://github.com/adobe/serenity-docs/pull/23

- `mysticat-data-service` — contract: this PR's model attributes and PostgREST query assume the `site_id`/`deleted_at`/`semrush_sub_workspace_id` columns and their grants exist, and that the sync trigger mirrors `semrush_sub_workspace_id → semrush_workspace_id` (adobe/mysticat-data-service#765).
- `spacecat-api-service` — contract: the sub-workspace market lifecycle handlers and a new reconcile endpoint (next PRs in this rollout) will call `BrandSemrushProject.create({ upsert: true, ... })` with the new attributes and read through `allByOrganizationId`; internal call sites cut over from `getSemrushWorkspaceId`/`setSemrushWorkspaceId` to the sub-workspace equivalents (adobe/spacecat-api-service#2739).

## 7. Test plan

Unit tests were run scoped to the touched entities (`test/unit/models/brand-semrush-project/`, `test/unit/models/brand/`) with coverage checked against this package's own 100%/97% thresholds — full coverage on all changed lines. `npx eslint` and the full `npx mocha` run on the three touched Brand test files (36 passing) were run locally. The package's integration test suite (`npm run test:it`, Docker + PostgREST against a pinned `mysticat-data-service` image) was not run this session: it needs a data-service image built from the companion migration (#765), which has not released yet, so the new columns aren't in that pinned schema. Before this merges, the integration test image pin (`packages/spacecat-shared-data-access/test/it/postgrest/docker-compose.yml`) should be bumped to a data-service release containing #765's migrations, and `npm run test:it` run to confirm the new columns, the `allByOrganizationId` join, and the flipped write-of-record against a real PostgREST schema.

- Depends on: https://github.com/adobe/mysticat-data-service/pull/765 — the `site_id`/`deleted_at`/`semrush_sub_workspace_id` columns and the reversed sync trigger this PR's model and query assume must exist in the schema this package's integration tests run against, and (for a real deploy) in the environment this package is used against.
- Blocks: the `spacecat-api-service` write-path PRs in the same rollout (they will bump their `@adobe/spacecat-shared-data-access` dependency to the release containing this PR).

Sequence: mysticat-data-service#765 merges and releases first; then this PR's integration tests are re-run against the updated pin and this PR merges/releases; then the api-service PRs bump their dependency.


packages/spacecat-shared-data-access/src/models/brand/brand.schema.js
```diff
@@ -35,14 +35,27 @@ const schema = new SchemaBuilder(Brand, BrandCollection)
     type: Brand.STATUSES,
     validate: (value) => value == null || Brand.STATUSES.includes(value),
   })
+  // DEPRECATED (serenity-docs brand-semrush-mapping-maintenance.md §10
+  // rename, write-of-record cutover): read-only mirror of
+  // semrushSubWorkspaceId below, maintained entirely by the
+  // mysticat-data-service brands_sync_semrush_workspace_id trigger
+  // (migration 20260702094229). No setter — kept for backward compatibility
+  // with any consumer still reading this attribute directly; app code must
```
MysticatBot: issue (blocking): Removing `setSemrushWorkspaceId` (via `readOnly: true`) is a **semver-breaking change**. Any consumer currently calling `brand.setSemrushWorkspaceId(value)` will get a runtime error on upgrade. Since the commit message is `feat(...)`, semantic-release will produce a minor bump - consumers auto-upgrading within the minor range will break silently.

**Why it matters:** `spacecat-shared` is an npm library consumed by multiple services. Even if the coordinated spacecat-api-service#2739 cuts over at the same time, any other consumer that bumps this dependency gets the break.

**How to fix (pick one):**
1. Add a `BREAKING CHANGE:` footer to the commit message so semantic-release produces a major bump with a clear migration note.
2. Keep a deprecated `setSemrushWorkspaceId` that delegates to `setSemrushSubWorkspaceId` internally (remove in a future major).

packages/spacecat-shared-data-access/src/models/brand-semrush-project/brand-semrush-project.collection.js
```diff
@@ -34,6 +49,94 @@ class BrandSemrushProjectCollection extends BaseCollection {
...
+   * @returns {Promise<Array<{
+   *   brandId: string,
+   *   semrushProjectId: string,
+   *   geoTargetId: number,
+   *   languageCode: string,
+   *   siteId: string|null,
+   *   organizationId: string,
+   *   semrushSubWorkspaceId: string|null,
+   * }>>}
+   */
```
MysticatBot, line 86: issue (blocking): The embedded-join filter `.eq("brands.organization_id", organizationId)` relies on PostgREST v11+ behavior where filtering on an embedded resource implicitly converts the join to INNER (excluding parent rows that do not match). On older PostgREST versions this is a LEFT JOIN - the query returns ALL mapping rows with `brands: null` on non-matching parents.

**Why it matters:** The behavior is version-dependent and non-obvious. Unit tests mask this because the mock only returns matching rows. The integration test gate (noted in your test plan) will catch it, but the code should be explicit.

**How to fix:** Use the `!inner` modifier in the select to make the join type explicit:
```js
const select = `brand_id, semrush_project_id, semrush_location_id, language, site_id, brands!${BRAND_FK}!inner(organization_id, semrush_workspace_id)`;
```
This makes INNER JOIN semantics explicit and removes the version dependency.

MysticatBot (changes requested):
Hey @rainer-friederich,

**Verdict: Request changes** - two issues need attention before this ships safely as a minor release.
**Complexity: HIGH** - large diff (690 lines, 12 files); API surface change with cross-service contract.
_Changes: adds siteId/deletedAt attributes to BrandSemrushProject, a new allByOrganizationId collection accessor with embedded-join pagination, and flips the Brand write-of-record from semrushWorkspaceId to semrushSubWorkspaceId (12 files)._

## Must fix before merge
1. [Important] Removing `setSemrushWorkspaceId` is a semver-breaking change without a BREAKING CHANGE footer - `brand.schema.js` (details inline)
2. [Important] PostgREST embedded-join filter semantics are version-dependent; use explicit `!inner` modifier - `brand-semrush-project.collection.js:77` (details inline)

<details>

- suggestion: Add a deterministic secondary sort (e.g. `.order('semrush_project_id')`) to pagination - one brand has many projects, so `brand_id` alone does not guarantee stable page boundaries. - `brand-semrush-project.collection.js:96`
- suggestion: Add a configurable max-row safety limit (e.g. 50,000) to `#fetchOrgRows` to prevent unbounded memory growth for unexpectedly large orgs. The base collection per-FK queries are naturally bounded; this org-level fan-out is not. - `brand-semrush-project.collection.js:78`
- nit: The select reads `brands.semrush_workspace_id` (the deprecated column) and maps it to `semrushSubWorkspaceId`. This works via the DB trigger mirror, but couples the accessor to a deprecated artifact. Consider switching to `semrush_sub_workspace_id` once the companion migration lands, or add a comment noting the temporary indirection. - `brand-semrush-project.collection.js:77`
- nit: `BrandSemrushProjectOrgRow.organizationId` is typed `string | null` but will never be null when the join succeeds (INNER JOIN excludes non-matching parents). Consider `string` for a tighter consumer contract. - `index.d.ts:45`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 10m 57s | <b>Cost:</b> $7.18 | <b>Commit:</b> <code>34bf9eda8eb4789808dc80988e5a37d9a1a7329d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-data-access/src/models/brand/brand.schema.js
rainer-friederich: Fixed in `ce9b3340` — restored a manual, deprecated `setSemrushWorkspaceId()` delegate on the model (delegates to `setSemrushSubWorkspaceId`) rather than a hard readOnly flip, so removing the schema-generated setter is not a semver-breaking change for any existing external caller. No BREAKING CHANGE footer / major bump needed since nothing is actually removed from the public API surface.

packages/spacecat-shared-data-access/src/models/brand-semrush-project/brand-semrush-project.collection.js
rainer-friederich: Fixed in `ce9b3340` — added the explicit `!inner` modifier to the embedded join select, removing the dependency on PostgREST-version-specific implicit-INNER-on-filter behavior. Also switched the embed to read the write-of-record `semrush_sub_workspace_id` column (the deprecated-column nit) and added a deterministic secondary sort + a MAX_ORG_ROWS safety cap in the same pass.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior blocking findings are addressed; remaining items are non-blocking nits.
**Complexity: HIGH** - large diff (690 lines, 13 files); API surface change with cross-service contract.
_Changes: adds siteId/deletedAt attributes to BrandSemrushProject, a paginated allByOrganizationId accessor with embedded-join and soft-delete filtering, and flips the Brand write-of-record from semrushWorkspaceId to semrushSubWorkspaceId with a BC delegate (13 files)._

<details>

- nit: Optional chaining on `brands` fields (`row.brands?.organization_id ?? null`) is unreachable under the `!inner` join guarantee and creates a type-contract gap - TS declares `organizationId: string` but runtime could produce `null` on a dead path. The corresponding test exercising `brands: null` validates an impossible PostgREST response. - `brand-semrush-project.collection.js:148`
- suggestion: JSDoc says "plain frozen identity DTOs" but returned objects are not `Object.freeze()`-d - consider adding freeze to enforce the documented contract (cheap at sub-50k rows). - `brand-semrush-project.collection.js:142`
- suggestion: Input validation on `organizationId` only checks truthiness (`if (!organizationId)`) - a non-UUID string passes the guard and returns empty results silently. Adding `isValidUUID(organizationId)` would fail fast with a clear error. - `brand-semrush-project.collection.js:76`
</details>

- setSemrushWorkspaceId restored as manual deprecated delegate (not a semver break)
- Explicit `!inner` modifier added to embedded join (no version dependency)
- Deterministic secondary sort on `semrush_project_id` added
- MAX_ORG_ROWS = 50,000 safety cap added
- Select switched from deprecated column to write-of-record `semrush_sub_workspace_id`
- TS type for `organizationId` corrected to `string` (reflects !inner guarantee)

Note: Security review did not complete; the attack surface is minimal (data-access library, parameterized PostgREST queries, no HTTP endpoints, no secrets).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 23s | <b>Cost:</b> $4.53 | <b>Commit:</b> <code>ce9b33409d48dc2c307d96c6232d197e56ea1149</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

vivesing (approved):

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - all prior blocking findings are addressed and the new commits are clean.
**Complexity: HIGH** - large diff (849 lines, 14 files); API surface change with cross-service contract.
_Changes: adds siteId/deletedAt attributes to BrandSemrushProject, a paginated allByOrganizationId accessor with embedded-join and soft-delete filtering, flips the Brand write-of-record from semrushWorkspaceId to semrushSubWorkspaceId with a BC delegate, and bumps the IT image to v5.57.0 (13 files)._

<details>

- nit: The pagination loop in `#fetchOrgRows` couples to the PostgREST builder being immutable (`.range()` returns a new builder rather than mutating in place). This is true for the current supabase-js lineage, but a one-line comment noting the assumption would help a future maintainer who encounters a different client fork. - `brand-semrush-project.collection.js:84`
- suggestion: The `allByOrganizationId` accessor implements its own offset-based pagination rather than reusing the existing cursor-based utility in `postgrest.utils.js`. This is justified (frozen DTOs, not model instances), but a brief comment noting why would prevent a future "should we refactor this" loop. - `brand-semrush-project.collection.js:70`
- suggestion: Consider importing `DEFAULT_PAGE_SIZE` in the pagination unit tests rather than re-declaring `const DEFAULT_PAGE_SIZE = 1000` locally. If the production constant ever changes, the test's hardcoded value would silently produce false-green results. - `brand-semrush-project.collection.test.js:518`
</details>

- setSemrushWorkspaceId restored as manual deprecated delegate (not a semver break)
- Explicit `!inner` modifier added to embedded join (no version dependency)
- Deterministic secondary sort on `semrush_project_id` added
- MAX_ORG_ROWS = 50,000 safety cap added
- Select switched from deprecated column to write-of-record `semrush_sub_workspace_id`
- Object.freeze applied to returned DTOs
- UUID validation via isValidUUID added with DataAccessError on failure
- IT docker image bumped to v5.57.0 (companion migration columns)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 31s | <b>Cost:</b> $5.58 | <b>Commit:</b> <code>374d293aea93a555091cea28e3a31c43f06c63d4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - the single new commit cleanly addresses all prior non-blocking suggestions.
**Complexity: HIGH** - large diff (861 lines, 14 files); API surface change with cross-service contract.
_Changes: adds siteId/deletedAt attributes to BrandSemrushProject, a paginated allByOrganizationId accessor with embedded-join and soft-delete filtering, flips the Brand write-of-record from semrushWorkspaceId to semrushSubWorkspaceId with a BC delegate, and bumps the IT image to v5.57.0 (14 files)._

No new issues found. The incremental commit adds two clarifying JSDoc comments and imports `DEFAULT_PAGE_SIZE` from the production module in tests rather than re-declaring a local constant.

- Pagination rationale documented (why offset loop instead of cursor-based postgrest.utils helper)
- `.range()` immutability assumption noted inline
- Tests import the real `DEFAULT_PAGE_SIZE` constant instead of re-declaring 1000 locally

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 26s | <b>Cost:</b> $1.50 | <b>Commit:</b> <code>e85c870e96121367ff4b9d98bfef562eca696a4c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - the single new commit is an unrelated lockfile version bump; no code changes since the last approval.
**Complexity: HIGH** - large diff (874 lines, 15 files); API surface change with cross-service contract.
_Changes: adds siteId/deletedAt attributes to BrandSemrushProject, a paginated allByOrganizationId accessor with embedded-join and soft-delete filtering, flips the Brand write-of-record from semrushWorkspaceId to semrushSubWorkspaceId with a BC delegate, and bumps the IT image to v5.57.0 (15 files)._

No new issues found. The only change since the prior approval is a `package-lock.json` version bump for the unrelated sibling package `spacecat-shared-project-engine-client` (1.8.0 to 1.9.0).

- setSemrushWorkspaceId restored as manual deprecated delegate (not a semver break)
- Explicit `!inner` modifier added to embedded join (no version dependency)
- Deterministic secondary sort on `semrush_project_id` added
- MAX_ORG_ROWS = 50,000 safety cap added
- Select switched from deprecated column to write-of-record `semrush_sub_workspace_id`
- Object.freeze applied to returned DTOs
- UUID validation via isValidUUID added with DataAccessError on failure
- IT docker image bumped to v5.57.0 (companion migration columns)
- Pagination rationale documented (why offset loop instead of cursor-based postgrest.utils helper)
- `.range()` immutability assumption noted inline
- Tests import the real `DEFAULT_PAGE_SIZE` constant instead of re-declaring locally

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 53s | <b>Cost:</b> $1.94 | <b>Commit:</b> <code>ee3e969988c7d28a59eb76c5cdf181c51b348073</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1769, opened 2026-07-02
feat(akamai-client): add spacecat-shared-akamai-client package

nit23uec:
## Summary
- New `@adobe/spacecat-shared-akamai-client` package: a thin client for Akamai's Property Manager API (PAPI), authenticated with EdgeGrid (EG1-HMAC-SHA256), based on the `akamai-api` POC.
- Exposes `findPropertiesByDomain`, `getLatestVersion`, `getRuleTree`, `createVersion`, `updateRuleTree`, `activate`, `getActivation`, `listActivations`, `latestActivation`.
- EdgeGrid signing is implemented directly (no `akamai-edgegrid` dependency, keeping this monorepo's `fetch`-based client pattern) and cross-validated in tests against the official `edgegrid-python` reference implementation for byte-identical signatures.
- Every public method validates required args, URL-encodes path segments before interpolation, and `searchBy`'s `key` is restricted to an allowlist (avoids a `{ [key]: value }` prototype-pollution-via-`__proto__` gotcha).

## Test plan
- [x] `npm test -w packages/spacecat-shared-akamai-client` — 61 passing, 100% statements/lines/functions, 99.15% branches
- [x] `npm run lint -w packages/spacecat-shared-akamai-client` — clean
- [x] EdgeGrid signature cross-validated against Python's `edgegrid-python` for both a GET (no body) and PUT (with JSON body) request, fixed timestamp/nonce, byte-for-byte match
- [ ] Review by a maintainer familiar with the PAPI integration patterns used elsewhere in this repo

### DIFF
--- a/packages/spacecat-shared-akamai-client/package.json
+++ b/packages/spacecat-shared-akamai-client/package.json
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "description": "Shared modules of the Spacecat Services - Akamai Client",
   "type": "module",
+  "private": true,
   "engines": {
     "node": ">=22.0.0 <25.0.0",
     "npm": ">=10.9.0 <12.0.0"
--- a/packages/spacecat-shared-akamai-client/package.json
+++ b/packages/spacecat-shared-akamai-client/package.json
@@ -1,0 +1,47 @@
+{
+  "name": "@adobe/spacecat-shared-akamai-client",
+  "version": "1.0.0",
+  "description": "Shared modules of the Spacecat Services - Akamai Client",
+  "type": "module",
+  "engines": {
+    "node": ">=22.0.0 <25.0.0",
+    "npm": ">=10.9.0 <12.0.0"
+  },
+  "main": "src/index.js",
+  "types": "src/index.d.ts",
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
+    "test": "c8 mocha",
+    "lint": "eslint .",
+    "clean": "rm -rf package-lock.json node_modules"
+  },
+  "mocha": {
+    "require": "test/setup-env.js",
+    "reporter": "mocha-multi-reporters",
+    "reporter-options": "configFile=.mocha-multi.json",
+    "spec": "test/**/*.test.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/adobe/spacecat-shared.git"
+  },
+  "author": "",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/adobe/spacecat-shared/issues"
+  },
+  "homepage": "https://github.com/adobe/spacecat-shared#readme",
+  "dependencies": {
+    "@adobe/spacecat-shared-utils": "1.122.1"
+  },
+  "devDependencies": {
+    "chai": "6.2.2",
+    "chai-as-promised": "8.0.2",
+    "nock": "14.0.15",
+    "sinon": "22.0.0",
+    "sinon-chai": "4.0.1",
+    "typescript": "6.0.3"
+  }
+}
--- a/packages/spacecat-shared-akamai-client/src/akamai-client.js
+++ b/packages/spacecat-shared-akamai-client/src/akamai-client.js
@@ -64,7 +64,8 @@ export default class AkamaiClient {
    * Creates an AkamaiClient from a Universal context. Reads EdgeGrid
    * credentials from context.env: AKAMAI_HOST, AKAMAI_CLIENT_TOKEN,
    * AKAMAI_CLIENT_SECRET, AKAMAI_ACCESS_TOKEN, and optionally
-   * AKAMAI_ACCOUNT_SWITCH_KEY.
+   * AKAMAI_ACCOUNT_SWITCH_KEY and AKAMAI_NOTIFY_EMAILS (comma-separated,
+   * required only if you intend to call activate()).
    *
    * @param {object} context - Universal function context
    * @returns {AkamaiClient}
@@ -77,9 +78,13 @@ export default class AkamaiClient {
       AKAMAI_CLIENT_SECRET: clientSecret,
       AKAMAI_ACCESS_TOKEN: accessToken,
       AKAMAI_ACCOUNT_SWITCH_KEY: accountSwitchKey,
+      AKAMAI_NOTIFY_EMAILS: notifyEmailsCsv,
     } = env;
+    const notifyEmails = hasText(notifyEmailsCsv)
+      ? notifyEmailsCsv.split(',').map((email) => email.trim()).filter(Boolean)
+      : undefined;
     return new AkamaiClient({
-      host, clientToken, clientSecret, accessToken, accountSwitchKey,
+      host, clientToken, clientSecret, accessToken, accountSwitchKey, notifyEmails,
     }, log);
   }
 
@@ -91,6 +96,8 @@ export default class AkamaiClient {
    * @param {string} config.clientSecret
    * @param {string} config.accessToken
    * @param {string} [config.accountSwitchKey]
+   * @param {string[]} [config.notifyEmails] - Required only to call activate();
+   *   emails PAPI notifies about activation progress.
    * @param {object} [log]
    */
   constructor(config = {}, log = console) {
@@ -104,6 +111,7 @@ export default class AkamaiClient {
     this.#clientSecret = config.clientSecret;
     this.#accessToken = config.accessToken;
     this.accountSwitchKey = config.accountSwitchKey;
+    this.notifyEmails = config.notifyEmails;
     this.log = log;
   }
 
@@ -229,8 +237,12 @@ export default class AkamaiClient {
       try {
         const items = await this.searchBy('hostname', host);
         items.forEach((item) => add(item, 'hostname', host));
-      } catch {
-        // no active version on that hostname, etc. — not a hard failure
+      } catch (e) {
+        // No active version on that hostname, auth failure, rate-limiting,
+        // etc. — not a hard failure (other candidates may still match), but
+        // log it so a systemic failure (e.g. bad credentials) is observable
+        // instead of silently returning an incomplete/empty result set.
+        this.log.warn(`searchBy('hostname', '${host}') failed, continuing: ${e.message}`);
       }
     }));
 
@@ -240,8 +252,8 @@ export default class AkamaiClient {
       try {
         const items = await this.searchBy('propertyName', nameQuery);
         items.forEach((item) => add(item, 'propertyName', nameQuery));
-      } catch {
-        // no match by that name — not a hard failure
+      } catch (e) {
+        this.log.warn(`searchBy('propertyName', '${nameQuery}') failed, continuing: ${e.message}`);
       }
     }));
 
@@ -379,12 +391,15 @@ export default class AkamaiClient {
   // --- Activation --------------------------------------------------------
 
   /**
+   * Activates a property version. Requires the client to have been
+   * constructed with a non-empty `notifyEmails` list (PAPI requires at least
+   * one address to notify about activation progress).
+   *
    * @param {string} propertyId
    * @param {number} version
    * @param {string} contractId
    * @param {string} groupId
    * @param {'STAGING'|'PRODUCTION'} network
-   * @param {string[]} notifyEmails
    * @param {string} [note]
    * @returns {Promise<string>} the activationLink returned by PAPI
    */
@@ -394,7 +409,6 @@ export default class AkamaiClient {
     contractId,
     groupId,
     network,
-    notifyEmails,
     note = 'Activated via spacecat-shared-akamai-client',
   ) {
     requireText('propertyId', propertyId);
@@ -403,6 +417,11 @@ export default class AkamaiClient {
     requireText('network', network);
     if (!Number.isInteger(version)) {
       throw new Error('version must be an integer');
+    }
+    if (!Array.isArray(this.notifyEmails) || this.notifyEmails.length === 0) {
+      throw new Error(
+        'AkamaiClient must be constructed with a non-empty notifyEmails array to call activate()',
+      );
     }
     const id = encodePathSegment(propertyId);
     const upperNetwork = network.toUpperCase();
@@ -413,7 +432,7 @@ export default class AkamaiClient {
         propertyVersion: version,
         network: upperNetwork,
         note,
-        notifyEmails,
+        notifyEmails: this.notifyEmails,
         acknowledgeAllWarnings: true,
       },
     });
--- a/packages/spacecat-shared-akamai-client/src/akamai-client.js
+++ b/packages/spacecat-shared-akamai-client/src/akamai-client.js
@@ -1,0 +1,504 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { hasText, tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
+
+import { signRequest } from './edgegrid-auth.js';
+
+const REQUIRED_CONFIG_KEYS = ['host', 'clientToken', 'clientSecret', 'accessToken'];
+const SEARCH_KEYS = ['hostname', 'edgeHostname', 'propertyName'];
+
+function requireText(name, value) {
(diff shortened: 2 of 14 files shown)


packages/spacecat-shared-akamai-client/src/akamai-client.js
```diff
@@ -0,0 +1,504 @@
...
+      const entry = results.get(propertyId);
+      entry.matchedOn.add(reason);
+      entry.matchedValues.add(value);
+    };
+
+    await Promise.all([...hostCandidates].map(async (host) => {
+      try {
+        const items = await this.searchBy('hostname', host);
+        items.forEach((item) => add(item, 'hostname', host));
+      } catch {
```
MysticatBot: issue (blocking): The empty `catch {}` blocks here (and line 245) swallow all errors from `searchBy` - including auth failures, rate-limiting (429), and network partitions - without any logging. The caller receives an incomplete result set with no indication that some lookups failed.

This violates the error-handling convention: broad catch blocks that swallow unexpected errors mask real operational failures.

**Fix:** Log at `warn` level to preserve graceful-degradation while making partial failures observable:
```javascript
} catch (e) {
  this.log.warn(`searchBy('hostname', '${host}') failed, continuing: ${e.message}`);
}
```

packages/spacecat-shared-akamai-client/src/akamai-client.js
```diff
@@ -0,0 +1,504 @@
...
+   * @param {string} propertyId
+   * @param {number} version
+   * @param {string} contractId
+   * @param {string} groupId
+   * @param {'STAGING'|'PRODUCTION'} network
+   * @param {string[]} notifyEmails
+   * @param {string} [note]
+   * @returns {Promise<string>} the activationLink returned by PAPI
+   */
+  async activate(
```
MysticatBot, line 425: issue (blocking): Every other parameter in `activate()` is validated (`propertyId`, `contractId`, `groupId`, `network`, `version`), but `notifyEmails` is passed directly to the request body without checking that it is an array. Passing `undefined`, `null`, or a string produces a malformed PAPI request that returns a confusing Akamai-side error rather than a clear client-side validation message.

The TypeScript declaration says `notifyEmails: string[]`, but at runtime nothing enforces it.

**Fix:**
```javascript
if (!Array.isArray(notifyEmails)) {
  throw new Error('notifyEmails must be an array');
}
```

packages/spacecat-shared-akamai-client/test/akamai-client.test.js
```diff
@@ -0,0 +1,656 @@
...
+  clientSecret: 'test-client-secret',
+  accessToken: 'test-access-token',
+};
+const CONTRACT_ID = 'ctr_1';
+const GROUP_ID = 'grp_1';
+const PROPERTY_ID = 'prp_123';
+
+function makeLog() {
+  return {
+    info: sinon.stub(),
```
MysticatBot: issue (blocking): The CLAUDE.md Test Conventions section states:

> Each test file uses a sandbox: `const sandbox = sinon.createSandbox()` with `afterEach(() => sandbox.restore())`

This test file creates stubs directly via `sinon.stub()` in a `makeLog()` helper without using a sandbox. While test isolation is achieved via `beforeEach` recreation and nock cleanup, the documented convention is not followed.

**Fix:** Add a sandbox and use it for stub creation:
```javascript
const sandbox = sinon.createSandbox();
afterEach(() => { sandbox.restore(); nock.cleanAll(); nock.enableNetConnect(); });

function makeLog() {
  return { info: sandbox.stub(), debug: sandbox.stub(), warn: sandbox.stub(), error: sandbox.stub() };
}
```

MysticatBot (changes requested):
Hey @nit23uec,

**Verdict: Request changes** - three convention/validation issues should be addressed before merge.
**Complexity: HIGH** - large diff (new package, 1890+ lines across 14 files) with dependency additions.
_Changes: Adds a new `@adobe/spacecat-shared-akamai-client` package providing a thin PAPI client with custom EdgeGrid signing (14 files)._

## Must fix before merge

1. [Important] Empty catch blocks silently swallow all errors in `findPropertiesByDomain` - `src/akamai-client.js:232` (details inline)
2. [Important] `notifyEmails` parameter not validated as array in `activate()` - `src/akamai-client.js:391` (details inline)
3. [Important] Test files do not use the documented sinon sandbox pattern - `test/akamai-client.test.js:38` (details inline)

<details>
<summary>Non-blocking (7): minor issues and suggestions</summary>

- nit: `ruleFormat` is interpolated unsanitized into Content-Type header - consider validating against `/^[a-z0-9-]+$/i` - `src/akamai-client.js:325`
- nit: `accountSwitchKey` is a public instance property while the other credential fields use private `#` fields - inconsistent visibility - `src/akamai-client.js:107`
- nit: No test covers the body truncation path at MAX_BODY_BYTES (131072 bytes) - a regression there would cause silent 401s with large POST payloads - `src/edgegrid-auth.js:66`
- nit: Factory function not exported from `index.js` - CLAUDE.md documents the `createFooClient` pattern, though newer clients (Cloudflare, Splunk) use `static createFrom` instead - `src/index.js`
- suggestion: Add retry/backoff for transient failures (429s, network errors) - every downstream service will need this otherwise - `src/akamai-client.js:119`
- suggestion: Add host allowlist validation (e.g. must end with `.akamaiapis.net`) as defense-in-depth against credential leakage to non-Akamai hosts - `src/akamai-client.js:103`
- suggestion: Consider a request timeout via AbortController to prevent hung PAPI calls from blocking callers indefinitely - `src/akamai-client.js:119`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 30s | <b>Cost:</b> $5.28 | <b>Commit:</b> <code>1976a5f9230c7c5aff83ddb5443efed8d24bb4dc</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-akamai-client/src/akamai-client.js
nit23uec: Fixed in 7180a331 — both catch blocks now log at `warn` level with the failing candidate and error message, so partial-failure result sets are observable instead of silent.

packages/spacecat-shared-akamai-client/src/akamai-client.js
nit23uec: Addressed, with a design change: rather than just adding the `Array.isArray` check on the `activate()` parameter, I moved `notifyEmails` to client-level config (constructor / `AKAMAI_NOTIFY_EMAILS` env var), validated as a required non-empty array at call time inside `activate()`. Every other `activate()` dependency (host, credentials) is already client-level, so this keeps the validation while removing an argument every caller had to thread through. See 7180a331.

packages/spacecat-shared-akamai-client/test/akamai-client.test.js
nit23uec: Fixed in 7180a331 — added `const sandbox = sinon.createSandbox()`, stubs in `makeLog()` now use it, and `afterEach` calls `sandbox.restore()` alongside the existing nock cleanup.

MysticatBot (dismissed):
Hey @nit23uec,

**Verdict: Approve** - all prior findings addressed; remaining items are non-blocking nits.
**Complexity: HIGH** - large diff (new package, 2016 lines across 14 files) with dependency additions.
_Changes: Moves notifyEmails to client-level config with validation, adds warn logging to catch blocks, and adopts the sinon sandbox test pattern (14 files)._

<details>

- nit: Test for `findPropertiesByDomain` error-swallowing verifies results but does not assert `log.warn` was called - the whole point of the fix commit - `test/akamai-client.test.js:521`
- nit: No test exercises a POST body exceeding `MAX_BODY_BYTES` (131072) to confirm truncation produces stable signatures - a future refactor removing the subarray would only fail in production on large rule-tree POSTs - `src/edgegrid-auth.js:66`
- nit: `activationIdFromLink(undefined)` returns `''` rather than `undefined` - callers can still check with `if (id)` but `undefined` would be more idiomatic for "no value" - `src/akamai-client.js:355`
- suggestion: Add a targeted `normalizeDomain` test for `'example.com.'` (DNS trailing-dot hostname without path) to directly exercise the trailing-dot regex - `test/akamai-client.test.js`

</details>

- Empty catch blocks in `findPropertiesByDomain` now log at warn level with error context
- `notifyEmails` moved to client-level config with `Array.isArray` + non-empty validation at `activate()` call time
- Test file uses `sinon.createSandbox()` with `afterEach` restore

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 12s | <b>Cost:</b> $6.41 | <b>Commit:</b> <code>7180a33125f1e8d97748f8a18bf0902f1e7d3cc7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-akamai-client/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "@adobe/spacecat-shared-akamai-client",
+  "version": "1.0.0",
+  "description": "Shared modules of the Spacecat Services - Akamai Client",
+  "type": "module",
+  "private": true,
+  "engines": {
+    "node": ">=22.0.0 <25.0.0",
```
MysticatBot, line 8: issue (blocking): `"private": true` will cause `npm publish` to refuse to publish this package, regardless of the `publishConfig.access: "public"` setting below. The semantic-release pipeline (`@semantic-release/npm` in `.releaserc.cjs`) will fail when it tries to cut the first release.

The most recent comparable client (`spacecat-shared-cloudflare-client`) does NOT have `"private": true`.

**Fix:** Remove the `"private": true` line.

packages/spacecat-shared-akamai-client/test/edgegrid-auth.test.js
```diff
@@ -0,0 +1,128 @@
...
+    });
+
+    it('produces a different signature for a POST body than for the same body on PUT', () => {
+      const base = {
+        url: 'https://akab-xxxxx.luna.akamaiapis.net/papi/v1/some/path',
+        body: '{"a":1}',
+        timestamp: FIXED_TS,
+        nonce: FIXED_NONCE,
+        ...CREDS,
+      };
```
MysticatBot, line 84: issue (blocking): No test exercises a POST body exceeding `MAX_BODY_BYTES` (131072). If the `subarray(0, MAX_BODY_BYTES)` truncation in `contentHash` is accidentally removed or the constant changed, the only signal is production 401 errors on large rule-tree POSTs - the existing reference-vector tests use small bodies and would still pass.

**Fix:** Add a test confirming appending bytes beyond the limit does not change the signature:
```javascript
it('truncates POST body at MAX_BODY_BYTES so trailing bytes do not affect the signature', () => {
  const base = { method: 'POST', url: 'https://host.net/path', timestamp: FIXED_TS, nonce: FIXED_NONCE, ...CREDS };
  const exactBody = 'x'.repeat(131072);
  const overBody = exactBody + 'EXTRA_BYTES_BEYOND_LIMIT';
  expect(signRequest({ ...base, body: exactBody })).to.equal(signRequest({ ...base, body: overBody }));
});
```

MysticatBot (changes requested):
Hey @nit23uec,

**Verdict: Request changes** - two issues should be addressed before merge.
**Complexity: HIGH** - large diff (new package, 2017 lines across 14 files) with dependency additions.
_Changes: Adds a new `@adobe/spacecat-shared-akamai-client` package providing a thin PAPI client with custom EdgeGrid signing (14 files)._
Note: CI checks are still pending - resolve before merge.

## Must fix before merge

1. [Important] `"private": true` in package.json contradicts `publishConfig.access: "public"` and will block npm publish via semantic-release - `packages/spacecat-shared-akamai-client/package.json:8` (details inline)
2. [Important] No test exercises a POST body exceeding `MAX_BODY_BYTES` (131072) to confirm truncation produces stable signatures - `test/edgegrid-auth.test.js` (details inline)

<details>

- nit: `ruleFormat` is interpolated unsanitized into Content-Type header - consider validating against `/^[a-z0-9-]+$/i` - `src/akamai-client.js:325`
- nit: No assertion that `log.warn` was called in the "swallows errors from individual lookups" test - the whole point of the prior fix was making partial failures observable - `test/akamai-client.test.js:521`
- nit: No direct test for `normalizeDomain('example.com.')` (DNS trailing-dot hostname without path) - the existing test exercises the regex coincidentally, not deliberately - `test/akamai-client.test.js`
- nit: `network` parameter in `activate()` accepts any string (TypeScript declares `'STAGING' | 'PRODUCTION'` but runtime just uppercases without validation) - `src/akamai-client.js:395`
- suggestion: CLAUDE.md documents a `createFooClient(config, log)` factory export convention - this PR uses the `static createFrom(context)` pattern instead, which matches the newer CloudflareClient but diverges from the documented convention. Consider adding `export function createAkamaiClient(config, log)` alongside or updating the CLAUDE.md - `src/index.js`

</details>

- Empty catch blocks in `findPropertiesByDomain` now log at warn level with error context
- `notifyEmails` moved to client-level config with `Array.isArray` + non-empty validation at `activate()` call time
- Test file uses `sinon.createSandbox()` with `afterEach` restore

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 10s | <b>Cost:</b> $6.38 | <b>Commit:</b> <code>3efdc89277f98cf0bdc2fef48fe052e2273282e6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-akamai-client/package.json
nit23uec: Keeping `"private": true` is intentional — it's the workaround that keeps the release pipeline **green**, not an oversight.

A brand-new package cannot do its first npm publish through this repo's CI. Publishing is OIDC Trusted-Publisher only now (the `NPM_TOKEN` was retired), and `npm trust github <pkg>` can only bind a package that **already exists** on npm. So on first merge `@semantic-release/npm` would attempt an OIDC first-publish with no trust binding and fail the release job.

`"private": true` makes `@semantic-release/npm` a no-op — both `verifyConditions` and `publish` short-circuit on `pkg.private === true` (see the plugin's `index.js` / `publish.js`). The package still gets tagged, changelogged, and GitHub-released; it just isn't published. `publishConfig.access: "public"` is intentionally retained: it's inert while private and is exactly what the eventual first publish needs (scoped `@adobe/*` packages default to `restricted`).

Removal happens during the manual bootstrap: adobe-bot does one interactive `npm publish` → the package is added to `scripts/setup-npm-trusted-publishers.sh` `PACKAGES` + a trust binding is registered → then `private` is dropped. This mirrors how `project-engine-client` was brought onto npm (#1706). Comparable clients like `cloudflare-client` don't carry the flag because they're already published and bound.

packages/spacecat-shared-akamai-client/test/edgegrid-auth.test.js
nit23uec: Addressed in 2fc67a4e — added an `edgegrid-auth` test that signs a body at exactly `MAX_BODY_BYTES` (131072) alongside the same body with extra bytes appended, and asserts the signatures are identical. This guards the `subarray(0, MAX_BODY_BYTES)` truncation from being silently removed or the constant changed (the reference-vector tests use small bodies and wouldn't catch it).

MysticatBot (approved):
Hey @nit23uec,

**Verdict: Approve** - all prior blocking findings addressed; remaining items are non-blocking nits.
**Complexity: HIGH** - large diff (new package, 2000+ lines across 14 files) with dependency additions.
_Changes: Adds DRY validation helpers (requirePropertyRef, assertNetwork), ruleFormat regex validation, and a MAX_BODY_BYTES truncation test (14 files)._

<details>

- nit: `searchBy()` return type in `index.d.ts` is `Promise<object[]>` - a narrower `PropertySearchResult` interface with the known PAPI fields would give callers better type hints - `src/index.d.ts:112`
- nit: `normalizeDomain()` is a public export but doesn't document that IPv6 bracket notation (e.g. `[::1]`) is unsupported - the split-on-colon logic would mangle it - `src/akamai-client.js:464`
- suggestion: Add a test exercising `encodePathSegment` with special characters in a propertyId (e.g. `/`, `?`, `#`) to prove the URL injection defense works end-to-end via nock URL matching - `test/akamai-client.test.js`
- suggestion: Add a test where the same `propertyId` appears in both a hostname and propertyName search response to explicitly verify the de-duplication and `matchedOn` merge logic - `test/akamai-client.test.js`

</details>

- MAX_BODY_BYTES truncation test added, guarding the `subarray(0, 131072)` path against silent regression
- `"private": true` concern acknowledged as intentional OIDC Trusted-Publisher bootstrap pattern (matches project-engine-client)

---

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 3s | <b>Cost:</b> $3.73 | <b>Commit:</b> <code>e3b66d1b994f9de661de2a34c35009f56b0db61c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1778, opened 2026-07-06
fix(project-engine-client): enforce https:// on the brand_urls mock

rainer-friederich:
## 1. Abstract
The vendored Project Engine mock now enforces the live Semrush validation on `POST .../brand_urls`: a brand URL must be a literal, lower-case `https://` URL, and the mock answers a non-conforming entry with the same 400 (and the same go-validator tag) the live gateway returns, instead of silently accepting it.

## 2. Reasoning
While reworking the serenity-docs#25 brand-URL feature, a prod write-probe showed `create_brand_urls` REQUIRES a literal `https://`: a scheme-less value 400s on the go-validator `url` tag, and a valid non-https URL (`http://`, `ftp://`) 400s on the `startswith=https://` tag. The mock accepted all of these, so an integration test could go green over a write the live gateway rejects — exactly the fidelity gap that let the earlier "resolve brand URLs to their scheme-less form before writing" mis-design ship (the scheme-less resolve value cannot be written as a brand URL at all).

Because the point of this gate is fidelity, the contract was probed rather than inferred: ~40 edge cases were POSTed to the live API and the mock reproduces the observed matrix, including the places where go's semantics differ from JavaScript's and a `URL`-based check silently gets the wrong answer.

Before: the mock's `POST .../brand_urls` handler created a row for any `{ url, type }` entry, regardless of scheme, and always returned 200.

After: each entry's `url` is validated against the live contract before any row is created. A non-conforming entry short-circuits the whole batch with the gateway's 400 `BasicResponse`, naming the tag live names. `BrandURLRequest.URL` carries `required,url,startswith=https://`, evaluated in that order:

- `required` — an empty (or null/missing) value; reported before `url` is ever evaluated.
- `url` — the value must parse, under go's rules: a non-empty scheme plus a host, a fragment, or opaque content. A scheme-less value (`lovesac.com`, `www.x.com`, `//x.com`), a scheme with only a rooted path (`https:/x.com`), a bare `https://`, and any value containing whitespace fail here.
- `startswith=https://` — a literal, case-SENSITIVE prefix check against the raw value, reached by anything that parsed. `http://`, `ftp://`, an upper- or mixed-case scheme (`HTTPS://X`, `Https://x`), every `file:` form, and the hostless-but-opaque forms (`mailto:`, `tel:`, `https:x.com`) fail here — not on `url`.

A conforming value is accepted and stored verbatim (no scheme or `www.` normalization), unchanged from today; an upper-case host, an IDN host, an IP, and a port/path/query are all accepted, as is `https://#frag` (a non-empty fragment satisfies `url` even with no host). The batch is atomic: one bad entry creates nothing.

The check lives in a pure `brandUrlHttpsTag` helper exposed on the request context (same convention as the existing `resolveUrl` / `tagId` lib helpers), so it is unit-tested independently of the coverage-excluded route handler. It reproduces go's parse by hand rather than delegating to JavaScript's `URL`, whose WHATWG parsing disagrees with go exactly where it matters: it trims surrounding whitespace, and it reports no host for the opaque forms go accepts.

## 4. Required information
- Jira / issue: serenity-docs#25 — https://github.com/adobe/serenity-docs/issues/25
- Other: write-probe of the live `create_brand_urls` contract (2026-07-13, throwaway benchmarks in a dev workspace, cleaned up in the same run)

- spacecat-api-service (consumer) — its serenity brand-URL write path (PR #2748) writes only `https://` brand URLs after the skip-primary-domain redesign; this mock keeps its integration tests honest to the live 400.
- mysticat-data-service (consumer) — the serenity_migration CLI (PR #737) likewise writes only `https://` brand URLs; same fidelity guarantee.

The contract was captured by a live write-probe against prod `adobe-hackathon.semrush.com` on 2026-07-13: throwaway competitor benchmarks were created in the LLMO-Dev-2 dev sub-workspace, ~40 candidate values were POSTed one per request, the accept/reject verdict and go-validator tag were recorded for each, and every benchmark and row created was deleted in the same run. No customer data was touched.

Two results are worth calling out because they contradict what the code would otherwise assume:
- go-validator's `startswith` is `strings.HasPrefix` against the raw value, so an upper-case scheme (`HTTPS://…`) is rejected, while an upper-case host (`https://EXAMPLE.NET`) is fine. Only the scheme must be literal.
- every `file:` form clears the `url` tag and fails on `startswith` — including `file://` and `file:///`, which go-validator's current source would reject on `url`. The deployed validator predates that path check, so the probed behaviour is what the mock models; the code says so at the point it matters.

## 7. Test plan
- Unit: the probed matrix is recorded case by case against the pure helper — accepts, the case-sensitivity of the scheme, scheme-less, non-https, opaque/hostless, `file:`, degenerate and whitespace forms, `required`, and the non-string unmarshal failure. The package's 100% branch-coverage gate is met.
- E2E (`test:e2e`, which boots the real mock server): the route answers the live 400 and tag over real HTTP for the scheme-less, `http://`, upper-case-scheme, `mailto:` and empty-url forms, a missing url is rejected by request validation, the batch is proven atomic (a good entry alongside a bad one is not created), and the existing create/list/delete e2e (which uses an `https://` URL) still passes through the new gate.
- No environment rollout applies — this is a test-only mock artifact in a shared package; it ships in the next `spacecat-shared-project-engine-client` release and is picked up by consumers when they bump.

- serenity-docs#25 redesign is a coordinated set: this mock fix hardens the contract that spacecat-api-service PR #2748 (https://github.com/adobe/spacecat-api-service/pull/2748) and mysticat-data-service PR #737 (https://github.com/adobe/mysticat-data-service/pull/737) rely on. It is independent to merge (the consumers do not import it at runtime), but merging it first and cutting a release lets those PRs' integration tests run against the tightened mock. Suggested order: this PR (+ release) → then #2748 / #737.

### DIFF
--- a/packages/spacecat-shared-project-engine-client/mock/brand-url-validation.js
+++ b/packages/spacecat-shared-project-engine-client/mock/brand-url-validation.js
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+// @ts-check
+
+/**
+ * The `url` validation the live Semrush Project Engine `POST .../brand_urls` enforces on every
+ * entry — the single source of truth behind the `brand_urls.js` route's 400, exposed on the
+ * per-request context as `context.brandUrlHttpsTag` (every route reads its lib helpers through
+ * `$.context`, never an import — see {@link Context}). Kept a pure function so it is unit-tested on
+ * its own (the route handler is coverage-excluded), the same convention as {@link resolveUrl} /
+ * {@link tagId}.
+ *
+ * Live contract (write-probed 2026-07-06 against prod `adobe-hackathon.semrush.com` — a throwaway
+ * benchmark in a test workspace, serenity-docs#25): a brand URL MUST be a literal `https://` URL.
+ * The go-validator on `BrandURLRequest.URL` runs two tags in order:
+ *   - `url`: the value must be a parseable absolute URL (scheme + host). A scheme-less value
+ *     (`instagram.com/lovesac`, `lovesac.com`, `www.x.com`) fails HERE → 400 `'url' tag`.
+ *   - `startswith=https://`: a valid but non-https URL (`http://x`, `ftp://x`) passes `url` but
+ *     fails HERE → 400 `'startswith' tag`.
+ * A conforming `https://…` value returns null (accepted, stored verbatim — the API does NOT
+ * normalize scheme/`www.`). Enforcing this in the mock keeps an IT from going green over a write
+ * the live gateway would 400 (the fidelity gap that let the url/resolve mis-design ship).
+ *
+ * @param {unknown} url the raw `url` field of a brand-URL entry
+ * @returns {'url' | 'startswith' | null} the failed go-validator tag, or null when accepted
+ */
+export const brandUrlHttpsTag = (url) => {
+  const value = typeof url === 'string' ? url : '';
+  let parsed;
+  try {
+    parsed = new URL(value);
+  } catch {
+    // Not a parseable absolute URL (scheme-less / garbage) — fails the `url` tag.
+    return 'url';
+  }
+  // A URL with no host (e.g. `mailto:hi@x.com`) is not a brand URL either → `url` tag.
+  if (!parsed.host) {
+    return 'url';
+  }
+  // Valid URL, but the value must literally begin with `https://` (case-insensitive scheme).
+  if (!value.toLowerCase().startsWith('https://')) {
+    return 'startswith';
+  }
+  return null;
+};
--- a/packages/spacecat-shared-project-engine-client/mock/brand-url-validation.js
+++ b/packages/spacecat-shared-project-engine-client/mock/brand-url-validation.js
@@ -13,42 +13,114 @@
 // @ts-check
 
 /**
- * The `url` validation the live Semrush Project Engine `POST .../brand_urls` enforces on every
- * entry — the single source of truth behind the `brand_urls.js` route's 400, exposed on the
- * per-request context as `context.brandUrlHttpsTag` (every route reads its lib helpers through
- * `$.context`, never an import — see {@link Context}). Kept a pure function so it is unit-tested on
- * its own (the route handler is coverage-excluded), the same convention as {@link resolveUrl} /
- * {@link tagId}.
+ * The `url` tag as the DEPLOYED gateway actually behaves — derived from the 2026-07-13 write-probe,
+ * NOT from reading go-validator's source (the two disagree; see the `file:` note below).
  *
- * Live contract (write-probed 2026-07-06 against prod `adobe-hackathon.semrush.com` — a throwaway
- * benchmark in a test workspace, serenity-docs#25): a brand URL MUST be a literal `https://` URL.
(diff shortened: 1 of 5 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean, well-tested patch that closes a real mock fidelity gap.
**Complexity: LOW** - small single-service change, test-only mock code.
_Changes: Adds https:// URL validation to the brand_urls mock POST handler to match the live Semrush Project Engine API's go-validator behavior (5 files)._

<details>

- nit: Unit test covers `undefined` but not `null` as input to `brandUrlHttpsTag` - adding `expect(brandUrlHttpsTag(null)).to.equal('url')` documents the coercion contract for readers - `packages/spacecat-shared-project-engine-client/test/mock/brand-url-validation.test.js`
- suggestion: Confirm existing e2e brand_urls creation tests (above line 1834) send `https://` URLs, which would implicitly validate the happy path through the new validation gate - `packages/spacecat-shared-project-engine-client/test/e2e/project-engine-mock.e2e.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 15s | <b>Cost:</b> $3.41 | <b>Commit:</b> <code>b8bd6e7be585233e2a36231002d77399399c2864</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean, focused mock fidelity fix with thorough test coverage; no blocking issues found on re-review.
**Complexity: HIGH** - medium diff; API surface risk flag (mock route handler).
_Changes: Adds https:// URL validation to the brand_urls mock POST handler, reproducing the live Semrush go-validator behavior so integration tests cannot pass over writes the live gateway would reject (5 files)._
Note: CI "Test" check is still pending - confirm it passes before merge.

<details>

- nit: No unit test exercises a non-whitespace control character (e.g., U+0001 or U+007F) against `isGoParseableUrl` - the regex handles the U+0000-U+001F and U+007F ranges explicitly, but coverage is only indirect via whitespace overlap. Adding a test case like `expect(brandUrlHttpsTag('https://x' + String.fromCharCode(1) + '.com')).to.equal('url')` would pin the control-byte path directly. - `packages/spacecat-shared-project-engine-client/test/mock/brand-url-validation.test.js`
- suggestion: The batch-atomicity e2e test asserts `response.status === 400` and verifies the good entry was not created, but does not assert the error message content. A `expect(error.message).to.match(/failed on the 'url' tag/)` would confirm the rejection came from the validation gate rather than from some other 400 source. - `packages/spacecat-shared-project-engine-client/test/e2e/project-engine-mock.e2e.js:257`
</details>

- Unit test for `null` input to `brandUrlHttpsTag` was already present at line 377 (prior nit was based on an incorrect reading of the test file).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 15s | <b>Cost:</b> $4.02 | <b>Commit:</b> <code>673ae62f5f2b433f93a79d253d2a020683838fdb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - both prior nits addressed; no new issues on the incremental delta.
**Complexity: HIGH** - medium diff; API surface risk flag (mock route handler).
_Changes: Adds https:// URL validation to the brand_urls mock POST handler to match the live Semrush go-validator behavior (5 files)._
Note: CI "Test" and "E2E" checks are still pending - confirm they pass before merge.

- Control-byte unit test now pins U+0001, U+007F, and U+0000 directly against `brandUrlHttpsTag`.
- Batch-atomicity e2e test now asserts the error message content (`failed on the 'url' tag`), confirming the 400 came from the validation gate.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 20s | <b>Cost:</b> $1.75 | <b>Commit:</b> <code>1bf8bc9e59158a2b2c5b711b914d25449a83eb87</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1784, opened 2026-07-07
fix(akamai-client): re-sign EdgeGrid requests across redirects

nit23uec:
## Summary

Fixes a bug in `@adobe/spacecat-shared-akamai-client` where EdgeGrid-signed requests failed with `401 "The signature does not match"` whenever PAPI returned a redirect.

PAPI 301-redirects some GETs to a concrete URL — most notably `GET /papi/v1/properties/{id}/versions/latest` → `/versions/{N}`. The `EG1-HMAC-SHA256` signature is bound to the **exact request URL**, so the transparently-followed redirect carried an `Authorization` header signed for the *previous* URL and was rejected:

```
401 pep-authn/deny  "The signature does not match"
```

This broke `getLatestVersion()` against live Akamai — and therefore any flow that starts from the latest property version.

## Fix

`#request` now uses `redirect: 'manual'`, follows 3xx responses itself, and **re-signs each hop** for the new (Location-resolved) URL, capped at 5 hops. Method/body are preserved (PAPI only redirects idempotent GETs here).

## Verification

Reproduced and fixed against a **live Akamai property** using real EdgeGrid credentials:

- Before: `getLatestVersion` → `401 signature does not match` (error `instance` pointed at the redirected `/versions/{N}` URL).
- After: `getLatestVersion` → returns the concrete version (e.g. `32`); `getRuleTree(v32)` and `listActivations` (which don't redirect) were unaffected throughout.

This unblocks the Akamai CDN-onboarding `plan`/`deploy` flow in `spacecat-api-service` (adobe/spacecat-api-service#2769), which pins this package via gitpkg and needs this commit.

## Test plan
- [x] New nock tests: redirect is followed + re-signed; too-many-redirects is capped
- [x] Full package suite green (69 passing), 100% statements/lines/functions retained
- [x] `eslint` clean
- [ ] Release a new version so downstreams (api-service) can pin it

### DIFF
--- a/packages/spacecat-shared-akamai-client/src/akamai-client.js
+++ b/packages/spacecat-shared-akamai-client/src/akamai-client.js
@@ -15,6 +15,9 @@ import { hasText, tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
 import { signRequest } from './edgegrid-auth.js';
 
 const REQUIRED_CONFIG_KEYS = ['host', 'clientToken', 'clientSecret', 'accessToken'];
+// The HTTP redirect statuses we follow manually (re-signing each hop). Excludes 300 (multiple
+// choices) and 304 (not modified), which are not location-driven redirects to follow.
+const REDIRECT_STATUSES = new Set([301, 302, 303, 307, 308]);
 const SEARCH_KEYS = ['hostname', 'edgeHostname', 'propertyName'];
 const ACTIVATION_NETWORKS = ['STAGING', 'PRODUCTION'];
 // Akamai rule formats are "latest" or a dated token like "v2024-01-01". A strict
@@ -184,17 +187,31 @@ export default class AkamaiClient {
         throw new Error(`PAPI ${method} ${path} request failed: ${e.message}`);
       }
 
-      const isRedirect = res.status >= 300 && res.status < 400;
-      const location = isRedirect ? res.headers.get('location') : null;
+      const location = REDIRECT_STATUSES.has(res.status) ? res.headers.get('location') : null;
       if (!location) {
         break;
+      }
+      // Only safe, body-less methods are followed. PAPI only redirects idempotent GETs; replaying
+      // a POST/PUT body to a redirect target would be semantically wrong, so refuse it explicitly
+      // rather than silently re-issue the mutation elsewhere.
+      if (method !== 'GET' && method !== 'HEAD') {
+        throw new Error(`PAPI ${method} ${path} -> unexpected redirect (${res.status})`);
       }
       if (hop >= 5) {
         throw new Error(`PAPI ${method} ${path} -> too many redirects`);
       }
-      // Resolve relative Locations against the current URL, then re-sign for the new URL on the
-      // next iteration. Method and body are preserved (PAPI only redirects idempotent GETs here).
-      url = new URL(location, url).toString();
+      // Re-signing attaches the Authorization header to the redirect target, so refuse to cross to
+      // a different host (mirrors browsers stripping Authorization on cross-origin redirects) — the
+      // client only ever legitimately talks to its configured EdgeGrid host. Relative Locations
+      // resolve against the current URL and stay same-host.
+      const redirectUrl = new URL(location, url);
+      if (redirectUrl.host !== new URL(url).host) {
+        throw new Error(
+          `PAPI ${method} ${path} -> redirect to different host rejected: ${redirectUrl.host}`,
+        );
+      }
+      // Re-sign for the new URL on the next iteration.
+      url = redirectUrl.toString();
     }
 
     if (!res.ok) {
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-akamai-client/src/akamai-client.js
```diff
@@ -149,31 +149,52 @@ export default class AkamaiClient {
...
+      const isRedirect = res.status >= 300 && res.status < 400;
+      const location = isRedirect ? res.headers.get('location') : null;
+      if (!location) {
+        break;
+      }
+      if (hop >= 5) {
+        throw new Error(`PAPI ${method} ${path} -> too many redirects`);
+      }
+      // Resolve relative Locations against the current URL, then re-sign for the new URL on the
+      // next iteration. Method and body are preserved (PAPI only redirects idempotent GETs here).
```
MysticatBot: issue (blocking): **No origin validation on redirect target - EdgeGrid credentials sent to arbitrary hosts.**

The redirect loop resolves `Location`, re-signs the request with EdgeGrid credentials, and sends the `Authorization` header to whatever host the Location specifies. If the PAPI endpoint (or any intermediary) returned a Location pointing to a different origin, the client would generate a fresh EG1-HMAC-SHA256 signature for that URL and deliver it there.

Browsers strip `Authorization` on cross-origin redirects for exactly this reason. The fix is a 3-line same-origin check:

```js
const redirectUrl = new URL(location, url);
if (redirectUrl.host !== new URL(url).host) {
  throw new Error(
    `PAPI ${method} ${path} -> redirect to different host rejected: ${redirectUrl.host}`
  );
}
url = redirectUrl.toString();
```

Exploitability is low (requires compromised Akamai infra or TLS MITM), but the fix is trivial and this is standard practice for any client that re-attaches auth headers across redirects.

MysticatBot (changes requested):
Hey @nit23uec,

**Verdict: Request changes** - one defense-in-depth gap in redirect credential handling.
**Complexity: LOW** - small bug fix, single service.
_Changes: Adds manual redirect-following with EdgeGrid re-signing to fix 401s on PAPI redirects (2 files)._

## Must fix before merge
1. [Important] No origin validation on redirect target - credentials re-signed and sent to any host in the `Location` header - `packages/spacecat-shared-akamai-client/src/akamai-client.js:196` (details inline)

<details>

- nit: The 3xx range check (`>= 300 && < 400`) includes non-redirect codes like 304. Consider narrowing to `[301, 302, 303, 307, 308]` for clarity, though the `!location` guard prevents misbehavior today - `packages/spacecat-shared-akamai-client/src/akamai-client.js:186`
- suggestion: For 301/302/303, standard HTTP semantics downgrade the method to GET and drop the body. The comment correctly notes PAPI only redirects GETs (so `bodyStr` is undefined), but adding a method guard (`if (method !== 'GET' && method !== 'HEAD') throw ...`) would make the safe-method-only contract explicit and prevent future surprises - `packages/spacecat-shared-akamai-client/src/akamai-client.js:168`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 14s | <b>Cost:</b> $3.77 | <b>Commit:</b> <code>6ea8f2571de0a56cc9e21b972a7616d0feb2456d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-akamai-client/src/akamai-client.js
nit23uec: Fixed in d0af229d. Added a same-host guard before following any redirect — a `Location` pointing at a different host now throws (`redirect to different host rejected: <host>`) instead of re-signing and sending the `Authorization` header there. Confirmed against live Akamai that the real `/versions/latest` 301 is same-host, so the legitimate flow is unaffected. New test: `refuses to follow a redirect to a different host`.

MysticatBot (approved):
Hey @nit23uec,

**Verdict: Approve** - all prior findings addressed correctly, no new issues.
**Complexity: MEDIUM** - 138 changed lines across a single service.
_Changes: Adds manual redirect-following with EdgeGrid re-signing to fix 401s on PAPI redirects (2 files)._

<details>

- nit: `new URL(url)` is constructed on every loop iteration for the host comparison when the host cannot change across same-host hops; could extract a `baseHost` once before the loop. Cosmetic at max 5 iterations - `packages/spacecat-shared-akamai-client/src/akamai-client.js:191`
</details>

- Cross-host credential leakage: same-host guard added, throws on different-host Location
- Broad 3xx range: narrowed to explicit REDIRECT_STATUSES set (301/302/303/307/308)
- Unsafe method following: redirects now restricted to GET/HEAD only

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 18s | <b>Cost:</b> $3.58 | <b>Commit:</b> <code>d0af229d8b831624ea3b52e37b82d97354366217</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1788, opened 2026-07-08
fix(html-analyzer): strip FontFaceObserver/Next.js font-detection elements from HTML diff (LLMO-6061)

sneharora1:
## Summary
- Puppeteer captures FontFaceObserver/Next.js `@next/font` test elements before the library removes them; the Chrome extension sees the cleaned DOM so it is unaffected
- Added `isFontDetectionLeaf` helper plus `removeFontDetectionElements` (browser) and `removeFontDetectionElementsCheerio` (Node.js) to filter these out of both code paths
- Rebased on current main (includes video player container filtering from #1785)

## Test plan
- [ ] `npm test -w packages/spacecat-shared-html-analyzer` — all 79 tests pass
- [ ] Font-metrics test string `mmMwWLliI0fiflO` is stripped from output
- [ ] Repeated `word` tokens (10+) are stripped from output
- [ ] Legitimate content containing the word "word" is preserved

Closes LLMO-6061

### DIFF
--- a/packages/spacecat-shared-html-analyzer/src/html-filter.js
+++ b/packages/spacecat-shared-html-analyzer/src/html-filter.js
@@ -209,7 +209,7 @@ function removeAccessibilityElementsCheerio($) {
 
 /**
  * Returns true if the text is entirely composed of font-detection noise:
- * either the font-metrics test string (substring match to cover &N variants)
+ * either the font-metrics test string (startsWith to cover &N suffix variants)
  * or exclusively repeated "word" tokens (≥ MIN_REPS).
  * @param {string} text - trimmed textContent of the element
  * @returns {boolean}
@@ -218,7 +218,7 @@ export function isFontDetectionLeaf(text) {
   if (!text) {
     return false;
   }
-  if (text.includes(FONT_DETECTION_TEST_STRING)) {
+  if (text.startsWith(FONT_DETECTION_TEST_STRING)) {
     return true;
   }
   const tokens = text.trim().split(/\s+/);
--- a/packages/spacecat-shared-html-analyzer/src/html-filter.js
+++ b/packages/spacecat-shared-html-analyzer/src/html-filter.js
@@ -209,9 +209,8 @@ function removeAccessibilityElementsCheerio($) {
 
 /**
  * Returns true if the text is entirely composed of font-detection noise:
- * either the font-metrics test string or exclusively repeated "word" tokens.
- * Used on both leaf elements and (in a second pass) on container text to catch
- * split-span cases where each child holds fewer than MIN_REPS occurrences.
+ * either the font-metrics test string (substring match to cover &N variants)
+ * or exclusively repeated "word" tokens (≥ MIN_REPS).
  * @param {string} text - trimmed textContent of the element
  * @returns {boolean}
  */
@@ -228,11 +227,6 @@ export function isFontDetectionLeaf(text) {
 
 /**
  * Remove FontFaceObserver / Next.js font-detection test elements (browser environment).
- * Two-pass strategy:
- *   1. Leaf pass — removes single-element noise nodes.
- *   2. Container pass — removes wrappers whose combined text is entirely noise,
- *      catching split-span cases (<div><span>word</span><span>word</span>…</div>)
- *      where each child individually falls below the repetition threshold.
  * Exported for direct unit testing without a full browser environment.
  * @param {Element} element - DOM element to filter
  */
@@ -245,19 +239,10 @@ export function removeFontDetectionElements(element) {
       el.remove();
     }
   });
-  element.querySelectorAll('*').forEach((el) => {
-    if (el.children.length === 0) {
-      return; // already handled in pass 1
-    }
-    if (isFontDetectionLeaf((el.textContent || '').trim())) {
-      el.remove();
-    }
-  });
 }
 
 /**
  * Remove FontFaceObserver / Next.js font-detection test elements (Node.js / cheerio).
- * Two-pass strategy mirrors the browser version — see removeFontDetectionElements.
  * @param {CheerioAPI} $ - Cheerio instance
  */
 function removeFontDetectionElementsCheerio($) {
@@ -265,15 +250,6 @@ function removeFontDetectionElementsCheerio($) {
     const $el = $(el);
     if ($el.children().length > 0) {
       return;
-    }
-    if (isFontDetectionLeaf($el.text().trim())) {
-      $el.remove();
-    }
-  });
-  $('*').each((i, el) => {
-    const $el = $(el);
-    if ($el.children().length === 0) {
-      return; // already handled in pass 1
     }
     if (isFontDetectionLeaf($el.text().trim())) {
       $el.remove();
--- a/packages/spacecat-shared-html-analyzer/src/html-filter.js
+++ b/packages/spacecat-shared-html-analyzer/src/html-filter.js
@@ -100,6 +100,15 @@ const VIDEO_PLAYER_SELECTORS = [
   '.flowplayer', // Flowplayer
 ].join(', ');
 
+// FontFaceObserver / Next.js @next/font inject transient test elements into the DOM to
+// detect when custom fonts have loaded. Puppeteer captures these before the library cleans
+// them up; the Chrome extension sees the page after cleanup so it's unaffected.
+// The two patterns are:
+//   1. A <div> containing only 10+ repetitions of "word" (character-width measurement)
+//   2. Spans containing "mmMwWLliI0fiflO" (the classic font-metrics test string)
+const FONT_DETECTION_TEST_STRING = 'mmMwWLliI0fiflO';
+const FONT_DETECTION_WORD_MIN_REPS = 10;
+
 /**
  * Validates if an element is likely a cookie banner based on text content
  * Optimized: Set lookup + early exit for common keywords (3x faster)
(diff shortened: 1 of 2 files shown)


packages/spacecat-shared-html-analyzer/src/html-filter.js
```diff
@@ -198,6 +207,60 @@ function removeAccessibilityElementsCheerio($) {
...
+ */
+function isFontDetectionLeaf(text) {
+  if (!text) {
+    return false;
+  }
+  if (text.includes(FONT_DETECTION_TEST_STRING)) {
+    return true;
+  }
+  // Must consist solely of "word" tokens and whitespace
+  if (/[^word\s]/.test(text)) {
```
anuj-adobe: **Correctness bug:** this character-class check doesn't actually verify the text is composed solely of the token `"word"` — it only rejects characters outside `{w,o,r,d,whitespace}`. Verified locally:

```js
// False positive: legitimate-ish text with 10 literal "word" occurrences
// mixed with other real words built from the same 4 letters passes and
// gets the whole leaf (including non-noise content) removed:
isFontDetectionLeaf('word door word wood word row word odd word rod word door word wood word row word odd word rod')
// => true

// False negative: a concatenated run with no separators (plausible shape
// for an actual measurement string) is NOT caught, since \bword\b needs
// boundaries that adjacent repetitions don't have:
isFontDetectionLeaf('word'.repeat(15))
// => false
```

The doc comment above says "Must consist solely of 'word' tokens" — to actually enforce that, tokenize on whitespace and check every token equals `"word"`, e.g.:
```js
text.trim().split(/\s+/).every((t) => t === 'word')
```

packages/spacecat-shared-html-analyzer/src/html-filter.js
```diff
@@ -198,6 +207,60 @@ function removeAccessibilityElementsCheerio($) {
...
+  return (text.match(/\bword\b/g) || []).length >= FONT_DETECTION_WORD_MIN_REPS;
+}
+
+/**
+ * Remove FontFaceObserver / Next.js font-detection test elements (browser environment).
+ * Only targets leaf elements (no element children) to avoid removing ancestor containers
+ * that also hold legitimate content.
+ * @param {Element} element - DOM element to filter
+ */
+function removeFontDetectionElements(element) {
```
anuj-adobe: **Test coverage gap:** `removeFontDetectionElements` (the DOM/browser variant) isn't exercised by any test in this PR. `isBrowser()` (utils.js) requires `window`/`document`, and `test/setup-env.js` sets up no jsdom — so under Mocha every new test takes the `filterHtmlNode`/cheerio branch, meaning only `removeFontDetectionElementsCheerio` actually runs. A future edit that diverges the two duplicated implementations (or a DOM-vs-cheerio API mismatch) would ship untested. Worth adding a jsdom-backed test (or directly unit-testing this function) to cover the browser path.

anuj-adobe: **Robustness concern:** the leaf-only restriction (skip any element with element children) can miss the same payload if it's split across sibling leaf elements under a shared wrapper instead of one flat text leaf — e.g. `<div><span>word</span> <span>word</span> ...</div>` where each span holds fewer than 10 repetitions. Neither the (non-leaf) wrapper nor any individual (under-threshold) span would be removed in that case, so the diff noise this PR targets could still leak through for markup shaped slightly differently than the single-leaf test fixture.

anuj-adobe (commented):
Automated review: found one correctness bug and two coverage/robustness gaps in the new font-detection filter. Details inline.

packages/spacecat-shared-html-analyzer/src/html-filter.js
sneharora1: Fixed in f0f057a. Replaced the character-class regex with `text.trim().split(/\s+/).every((t) => t === 'word')` combined with the existing `>= FONT_DETECTION_WORD_MIN_REPS` length check. This correctly rejects mixed tokens like `door`, `wood`, `row` and also catches concatenated runs like `'word'.repeat(15)` when they are whitespace-separated. Added a dedicated `isFontDetectionLeaf (unit)` test suite covering: empty string, the mmMwWLliI0fiflO string, 10+ spaced tokens, fewer than 10 tokens, mixed `{w,o,r,d}`-letter words (false-positive case from the review), concatenated repetitions without spaces (false-negative case from the review), and legitimate prose.

packages/spacecat-shared-html-analyzer/src/html-filter.js
sneharora1: Fixed in f0f057a. Since jsdom is not available in this package's test setup, I took the 'directly unit-testing this function' path from the review suggestion: exported both `isFontDetectionLeaf` and `removeFontDetectionElements` from `html-filter.js` and added a `removeFontDetectionElements (browser path unit)` suite in `index.test.js`. The suite uses a hand-rolled DOM mock (stubbing `children`, `textContent`, `remove`, and `querySelectorAll`) to exercise the browser code path directly without needing a real DOM environment. Covers: leaf removal for mmMwWLliI0fiflO, leaf removal for word-repetition, no-op for legitimate content, and container removal for the split-span case.

sneharora1: Fixed in f0f057a. Added a second pass in both `removeFontDetectionElements` and `removeFontDetectionElementsCheerio` that runs after the leaf pass and checks non-leaf containers whose combined `textContent` matches `isFontDetectionLeaf`. This catches the split-span case (`<div><span>word</span><span>word</span>...<\/div>`) where each child individually falls below the 10-repetition threshold but the parent's combined text does not. The strict `every(t => t === 'word')` check from the regex fix prevents false positives on containers that mix noise with real content. Added a corresponding test: 'should remove split-span font-detection noise where each span holds fewer than 10 repetitions' in the Cheerio path, and a matching case in the browser-path unit suite.

MysticatBot (approved):
Hey @sneharora1,

**Verdict: Approve** - well-scoped bug fix with solid test coverage addressing all prior review feedback.
_Changes: Adds font-detection element stripping (FontFaceObserver/Next.js @next/font) to the HTML filter pipeline in both browser and cheerio paths (2 files)._

<details>

- nit: `isFontDetectionLeaf` JSDoc says "entirely composed of font-detection noise" but the `includes()` check matches elements that merely contain `mmMwWLliI0fiflO` as a substring - the behavior appears intentional (test fixtures use `mmMwWLliI0fiflO&1`) but the doc is slightly inaccurate - `packages/spacecat-shared-html-analyzer/src/html-filter.js:209`
- suggestion: Add a unit test pinning the substring-match behavior (`isFontDetectionLeaf('mmMwWLliI0fiflO&1')` returns true) so a future tightening to strict equality doesn't accidentally break the real-world case - `packages/spacecat-shared-html-analyzer/test/index.test.js`
- suggestion: Add boundary tests for exactly 9 and 10 "word" tokens to document the threshold edge - `packages/spacecat-shared-html-analyzer/test/index.test.js`
- suggestion: Consider scoping selectors from `*` to `div, span` (the only elements FontFaceObserver/Next.js inject) or adding a `text.length > 200` early-return guard in `isFontDetectionLeaf` to reduce traversal cost on large DOMs - `packages/spacecat-shared-html-analyzer/src/html-filter.js`

</details>

- Regex correctness bug (character-class check) replaced with strict token comparison
- Browser-path test coverage gap addressed with exported functions and DOM mock suite
- Split-span edge case handled via two-pass container removal strategy

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 24s | <b>Cost:</b> $5.86 | <b>Commit:</b> <code>f2ed9273b76ac9a77582151edee05e89f7f857d4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-html-analyzer/src/html-filter.js
```diff
@@ -198,6 +207,80 @@ function removeAccessibilityElementsCheerio($) {
...
+ * Used on both leaf elements and (in a second pass) on container text to catch
+ * split-span cases where each child holds fewer than MIN_REPS occurrences.
+ * @param {string} text - trimmed textContent of the element
+ * @returns {boolean}
+ */
+export function isFontDetectionLeaf(text) {
+  if (!text) {
+    return false;
+  }
+  if (text.includes(FONT_DETECTION_TEST_STRING)) {
```
nit23uec: it should be an exact match rather than includes.

packages/spacecat-shared-html-analyzer/src/html-filter.js
```diff
@@ -198,6 +207,80 @@ function removeAccessibilityElementsCheerio($) {
...
+  }
+  const tokens = text.trim().split(/\s+/);
+  return tokens.length >= FONT_DETECTION_WORD_MIN_REPS && tokens.every((t) => t === 'word');
+}
+
+/**
+ * Remove FontFaceObserver / Next.js font-detection test elements (browser environment).
+ * Two-pass strategy:
+ *   1. Leaf pass — removes single-element noise nodes.
+ *   2. Container pass — removes wrappers whose combined text is entirely noise,
```
nit23uec: is there a use case that needs a container pass?

packages/spacecat-shared-html-analyzer/src/html-filter.js
```diff
@@ -198,6 +207,80 @@ function removeAccessibilityElementsCheerio($) {
...
+    }
+    if (isFontDetectionLeaf((el.textContent || '').trim())) {
+      el.remove();
+    }
+  });
+}
+
+/**
+ * Remove FontFaceObserver / Next.js font-detection test elements (Node.js / cheerio).
+ * Two-pass strategy mirrors the browser version — see removeFontDetectionElements.
```
nit23uec: is there a hard orderign dependency also between these 2 passes?

packages/spacecat-shared-html-analyzer/src/html-filter.js
sneharora1: @nit23uec, FONT_DETECTION_TEST_STRING is mmMwWLliI0fiflO — real captured elements sometimes have it with extra characters like mmMwWLliI0fiflO&1, mmMwWLliI0fiflO&2 etc depending upon number of fonts being loaded.

packages/spacecat-shared-html-analyzer/src/html-filter.js
sneharora1: The container pass was added to address anuj-adobe's theoretical robustness concern: <div><span>word</span><span>word</span>…</div> where each span individually has fewer than 10 "word" tokens but the container's combined text does. However, FontFaceObserver and Next.js @next/font both inject flat single-element structures — a single <span> with the full test string. The split-span case doesn't appear to actually occur with either library. So the container pass is likely dead code addressing a pattern that doesn't exist. Will revert this change and simiplify the code.

packages/spacecat-shared-html-analyzer/src/html-filter.js
sneharora1: no longer applicable

packages/spacecat-shared-html-analyzer/src/html-filter.js
Danialkafeel: @sneharora1 maybe we can use `.startswith` which will be much safer than includes (to avoid unwanted removal).

sneharora1: done

Danialkafeel (approved):

nit23uec (changes requested):
wrong layer to fix this issue.

PR #666, opened 2025-03-19
feat: support external domain onedrive

alinarublea:
Please ensure your pull request adheres to the following guidelines:
- [ ] make sure to link the related issues in this description
- [ ] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes

## Related Issues

Thanks for contributing!

PR #1793, opened 2026-07-08
fix(cloud-manager-client): build BYOG PR URLs (Azure/Bitbucket) + hostname detection + retry-once

rpapani:
## Problem
`CloudManagerClient#createPullRequest` only constructed a `pullRequestUrl` for **GitHub** and **GitLab**. For **Azure DevOps** and **Bitbucket** BYOG repos it returned `null`, so `pullRequestUrl` was never set — even though the CM Repo API had created the PR (and returned its `externalNumber`). Downstream, the ASO UI showed no "View PR" link for those repos. Separately, PR creation had no resilience to transient CM-API blips.

## Changes
1. **Azure DevOps + Bitbucket PR URLs.** Added path templates: Azure DevOps `/pullrequest/{n}` (`dev.azure.com`, `*.visualstudio.com`), Bitbucket Cloud `/pull-requests/{n}` (`bitbucket.org`). Unknown hosts still return `null`.
2. **Hostname-based provider detection.** Provider is now detected from the parsed **hostname** (`extractRepoHost` via `URL()`), not a substring match on the whole URL — so a provider domain appearing in the repo *path* (e.g. `bitbucket.org/team/github.com-mirror`) can't be misclassified. (Addresses review feedback.)
3. **Retry PR creation once on transient failure.** `createPullRequest` retries a single time after a short delay on a transient failure (5xx / 429 / network error); permanent 4xx failures throw immediately.

## Tests
Unit tests for all four providers (incl. legacy `visualstudio.com`), the path-false-positive case, an unparseable `repoUrl`, and the retry paths (transient-then-success, network error, two-consecutive-failures). Package stays at **100%** coverage.

## Related PRs (cross-repo)
- **adobe/spacecat-autofix-worker#666** — consumes this client; CWV code-patch fix lifecycle.
- **OneAdobe/experience-success-studio-ui#2093** — ASO UI Deployed-tab.


packages/spacecat-shared-cloud-manager-client/src/index.js
```diff
@@ -1008,6 +1026,10 @@ export default class CloudManagerClient {
       provider = CM_REPO_TYPE.GITHUB;
     } else if (repoUrl.includes('gitlab.com') || repoUrl.includes('gitlab.')) {
       provider = CM_REPO_TYPE.GITLAB;
+    } else if (repoUrl.includes('dev.azure.com') || repoUrl.includes('visualstudio.com')) {
```
ramboz: N2 (nice-to-have): provider detection uses substring `includes()`, so a `repoUrl` whose *path* contains a provider domain (e.g. `https://bitbucket.org/team/github.com-mirror`) can be misclassified — the checks are ordered, so `github.com` would win here. This is the pre-existing pattern (github/gitlab already do it) just extended, and the risk is low: `repoUrl` comes from the customer's CM config and the worst case is a wrong PR-URL label. Worth hardening later by parsing the hostname. Also note `bitbucket.org` intentionally won't match self-hosted Bitbucket Server/Data Center (different path shape) — fine given the PR targets Bitbucket Cloud, just flagging the boundary.

ramboz (approved):
# PR Review: fix(cloud-manager-client): build PR URL for Azure DevOps and Bitbucket

## Summary
Adds Azure DevOps + Bitbucket Cloud provider detection and PR/MR path templates to `#buildPullRequestUrl`, and — **not mentioned in the description** — wraps the CM PR-creation `fetch` in a single-retry loop for transient (5xx / 429 / network) failures. The logic is correct, backward-compatible, and well-tested. Ready to merge; the one Should-Fix is doc-only.

This is one half of a cross-repo effort with `adobe/spacecat-autofix-worker#666`: this client must be released to npm first, then #666 bumps the dep. #666 relies on the new Azure/Bitbucket URL building here so the CWV "View PR" link renders for those BYOG repos, and the retry added here reduces how often #666's "branch pushed, PR failed" path is hit.

## Strengths
- The retry loop is carefully written. Permanent 4xx (except 429) throw immediately; transient 5xx/429 and thrown fetches (network/timeout) retry once; exhaustion throws a distinct `...failed after N attempts` message. The `if (response && !transient)` guard correctly distinguishes a re-thrown permanent HTTP error from a network throw, and every attempt-2 path either `break`s or throws — so the loop can never fall through to `response.json()` with an undefined/non-ok response.
- `#PR_PATH_BY_PROVIDER` is a clean frozen lookup table; unknown hosts still return `null`, so there is no regression for providers CM doesn't map.
- Good, targeted test coverage: all four providers (incl. legacy `visualstudio.com`), transient-5xx-then-success, network-error-then-success, and two-transient-then-throw.

## Issues

### Blockers (Must Fix Before Merge)
None.

### Should Fix
**S1. PR description + commit subject omit the retry behavior** — `packages/spacecat-shared-cloud-manager-client/src/index.js`
The title/body only describe the provider URL additions, but the diff also adds a single-retry loop to `createPullRequest`. That is a real behavior change for **all** callers of this shared method (eds-csp, accessibility, vulnerabilities, cwv): up to ~1.5s extra latency on a transient failure, plus a changed thrown-error message format (`HTTP <status> - ...` and `...failed after 2 attempts: ...`). Since merging triggers a patch release, the semantic-release changelog — generated from the commit subject — will not mention the retry. Please update the PR body/commit subject so the shipped behavior is captured.
*Perspective: Engineering Practices / Product*

### Nice to Have
**N1. The 422 test doesn't guard the no-retry-on-4xx contract** — `packages/spacecat-shared-cloud-manager-client/test/cloud-manager-client.test.js:1945`
`it('throws on failed PR creation')` uses a single nock interceptor and asserts `rejectedWith('Pull request creation failed')`. That matcher is a prefix that **also** matches the retry-exhausted message `Pull request creation failed after 2 attempts`, and there's no `scope.isDone()` assertion — so a regression that retried a 4xx would still pass this test. Consider asserting `expect(scope.isDone()).to.be.true` (proving exactly one request) or matching the exact `Pull request creation failed: HTTP 422` message.
*Perspective: QA / Testing*

**N2. Host detection via substring `includes()`** — see inline comment. Pre-existing pattern extended to the new providers; low risk. Noted inline.
*Perspective: Architecture*

## Verdict
**Ready to merge:** Yes (approving) — address S1 (description) before/at merge so the release notes are accurate.
**Reasoning:** Correct, backward-compatible, well-tested; the only Should-Fix is documentation of a shipped behavior change.

---
*Note: reviewed against the PR branch with full repo context. I did not execute the test suite locally (monorepo `node_modules` not installed); relying on CI + static analysis. No downstream code in the sibling repos string-matches the old error message, so the message-format change is safe.*

packages/spacecat-shared-cloud-manager-client/src/index.js
rpapani: Good catch — hardened as suggested. Provider detection now parses the hostname via `URL()` (`extractRepoHost`) and matches on that, so a provider domain appearing in the repo *path* (e.g. `https://bitbucket.org/team/github.com-mirror`) is classified by its host (bitbucket) rather than the path. Behaviour is unchanged for normal https repo URLs. Added tests for the path false-positive and an unparseable `repoUrl`. Done in e03cc82e.

packages/spacecat-shared-cloud-manager-client/test/cloud-manager-client.test.js
```diff
@@ -1895,5 +2007,86 @@ describe('CloudManagerClient', () => {
         },
       )).to.be.rejectedWith('Pull request creation failed');
```
ramboz: N1 (nice-to-have, non-blocking): `rejectedWith('Pull request creation failed')` is a prefix that also matches the retry-exhausted message `Pull request creation failed after 2 attempts`. With a single interceptor and no `scope.isDone()`, a regression that started retrying a 422 would still pass this test. To actually pin the no-retry-on-4xx contract, either assert `expect(scope.isDone()).to.be.true` (proving exactly one request was made) or match the exact `Pull request creation failed: HTTP 422` message. Fine to defer.

ramboz (approved):
Re-approving at e03cc82.

The hostname-detection refactor cleanly closes the earlier N2 (path false-positive): provider is now resolved from the parsed hostname via `extractRepoHost`/`URL()`, with tests for both the `bitbucket.org/team/github.com-mirror` path false-positive and an unparseable `repoUrl`. Behavior is unchanged for normal https repo URLs. I re-traced the retry control-flow as well — every loop exit lands on either a `break` (ok response) or a throw, so `response.json()` never runs on a non-ok/undefined response.

S1 (retry undocumented) is resolved: the PR body + title now describe the retry, and it has its own `fix:` commit so semantic-release will capture it in the changelog regardless of merge strategy.

No blockers. One optional, non-blocking test-robustness nit left inline (N1).

packages/spacecat-shared-cloud-manager-client/test/cloud-manager-client.test.js
rpapani: Good catch — pinned it in 47e688d2. The 4xx test now asserts the **exact** permanent-failure message (`Pull request creation failed: HTTP 422`), which is distinct from the `...failed after N attempts` retry-exhausted message, plus `expect(scope.isDone()).to.be.true` — the single interceptor proves no second request (retry) was attempted on a permanent 4xx. A regression that started retrying a 4xx now fails the test. 95 passing, 100% coverage.

PR #1799, opened 2026-07-09
feat(project-engine-client): model dimension-root tags in the mock

rainer-friederich:
## 1. Abstract
The Project Engine mock now models the dimension-root tag taxonomy: `category`, `intent`, `source` and `type` are bare-named root tags, every tag value is a bare-named descendant carrying `parent_id`, and a tag's dimension is `path[0]`. Mock tag ids are keyed on `(parent, name)`, every endpoint that returns a tag serializes it through one shared view, `GET /aio/tags` returns a full ancestry breadcrumb, and `DELETE /aio/tags` detaches the removed tag from every prompt carrying it.

## 2. Reasoning
The Brand Presence dashboard is a Semrush-owned micro-frontend whose tag filter is a single tree-aware dropdown ordered by the upstream, not alphabetically. A `<dimension>:<value>` name prefix therefore produces no visual grouping — a customer sees `category:Premium Modular Furniture`, `intent:Commercial`, `source:human` and `type:non-branded` interleaved in one flat list. The dropdown does render parent/child hierarchy, so making each dimension a root tag turns that list into four collapsible groups with no change to the micro-frontend. The tag tree is the only lever we own.

Dropping the prefix removes a namespace that several read paths silently depend on, and this package is where that surfaces first. Two defects made the mock unable to model the target shape, and two more made it disagree with production:

- Mock tag ids were a pure function of the tag name. Live tag uniqueness is scoped per `(project, parent)`, so once names go bare, two sub-categories called `Pricing` under different categories collapse onto one id. With prefixed names that collision was theoretical; with bare names it is routine, and it would silently corrupt the api-service integration suite.
- `GET /aio/tags` returned `path[]` as the direct parent only. That is indistinguishable from a full breadcrumb at depth 2, but a sub-category sits at depth 3 under this model, so `path[0]` resolved to its category rather than to the `category` dimension root. Both downstream consumers key a tag's dimension off `path[0]`, so every integration test would have passed against a mock that disagreed with live.
- `POST /aio/prompts/by_tags` echoed each prompt's stored tag rows verbatim, so an embedded tag carried neither `children_count` nor a `path`, and any rename or re-parent left a stale snapshot attached to the prompt. Live has exactly one tag serializer: fetch the same tag id from the prompt listing and from the tag tree and the two objects are equal. A consumer that reads a prompt's dimension off the embedded `path[0]` — which is precisely what WP2 and WP3 do — would therefore have read `undefined` against a mock that never told it so.
- `DELETE /aio/tags` removed the standalone tag but left it embedded on every prompt carrying it, so `by_tags` kept matching an id that no longer existed. Live detaches the tag from every carrying prompt, and a prompt whose only tag was deleted becomes fully unassigned. This divergence predates the program; it is corrected here because the migration CLI's reshape deletes tags, and because api-service has a tag-delete route whose integration tests would otherwise be validated against a mock that lies.

The taxonomy the mock serves changes shape. Before, a root was a colon-prefixed `category:<name>` and its children were bare, one level deep; the closed dimensions `intent`, `source` and `type` were flat prefixed roots, one node per value. After, the root level holds exactly four bare-named dimension roots, the closed dimensions carry their fixed vocabularies as children, a customer category sits at depth 2 and a sub-category at depth 3. No tag name contains a `:`.

Behaviour deltas:

- **Tag ids are derived from `(parent, name)`, not from the name alone.** The same bare name under two different parents now yields two distinct tags — which is what live does, and what the model needs so that a sub-category named `human` and the `source` value `human` can coexist on one prompt.
- **Every endpoint that returns a tag serializes it through one shared view.** A tag object is identical whether it is embedded on a prompt by `by_tags` or listed by the tag tree; what varies is the tag's depth, not the endpoint it came from. A dimension root is `{ children_count, id, name, prompts_count }` — the `parent_id` and `path` keys are absent, not `null` — and a descendant adds both. `children_count` and `path` are derived from the stored collection at read time rather than persisted, so a rename or a re-parent shows up on the next read of either endpoint instead of leaving a stale breadcrumb embedded on a prompt.
- **`GET /aio/tags` returns the full root-first ancestry in `path[]`,** excluding the tag itself, with each leaf beyond the root carrying its own `parent_id`. A depth-3 sub-category reads back as `[<dimension root>, <category>]`, so `path[0]` identifies the dimension at any depth.
- **`DELETE /aio/tags` detaches each removed id from every prompt that carries it.** The prompt survives; if that was its last tag it becomes fully unassigned, and it stops matching `by_tags` on the removed id. Deleting a parent still does not cascade to its children, which remain orphaned — that limitation is unchanged and still documented.
- **`POST /aio/prompts/tagged` registers the root tags it mints.** It embedded a tag on the created prompt without adding it to the tag collection, so a prompt created that way referenced an id the tree did not hold, and the shared serializer could only fall back to echoing a bare `{ id, name }`. The minted roots are now written (resolve-before-create, after the quota gate, so an over-allocation still writes nothing), which is what makes the one-serializer invariant total rather than true only of seeded tags.
- **`upsertMany` reports a collision for any already-stored id.** Because the id now encodes the parent, an id clash *is* a same-name/same-parent clash, which live answers with a hard 500. One divergence is deliberate and loud: a `PATCH` keeps a tag's id stable across a rename or a re-parent, so re-creating the tag's original `(parent, name)` collides here where live would mint a fresh opaque id. Failing loudly beats silently handing the caller back a tag that has moved elsewhere in the tree.
- **The seeds ship the new tree.** Four dimension roots, the closed vocabularies in full (all five intents, including `Navigational`, which every live customer project carries), and a depth-2 category with two depth-3 sub-categories. One sub-category is named `human` and shares that bare name with the `source` value `human`, differing only by parent, so the cross-dimension collision the model must survive is exercised by the suite rather than assumed. `SEED_IDS` gains the four dimension-root ids and the ids for those colliding and closed-dimension tags.

Two contract facts are now recorded where the next reader will hit them rather than being folk knowledge:

- `POST /aio/prompts/tagged` is name-keyed and can only ever address **root** tags. Its request shape carries names and has no field for a parent, so a nested tag is unreachable and a name absent from the root level mints a new root — which the mock now does, registering the minted roots in the project's tag collection so a prompt never references a tag the tree does not hold. Under this model every tag value is a descendant, so this endpoint cannot express a prompt's tags at all; the id-based `POST /aio/prompts { items, tag_ids }` replaces it. The route header says so, rather than inviting someone to invent a name-to-nested-tag lookup the real API does not have.
- A `PATCH` to a child that omits `parent_id` promotes it to a root, stranding it outside its dimension while every carrying prompt stays attached. That is the single destructive operation in the live reshape, and the tag-PATCH route header now names it.

Nothing in the published client (`src/`) changes, and the vendored swagger and its correction overlay are untouched — `model.AIOTagLeaf.parent_id` was already optional, so the wire schema needed no edit.

Documentation that described the old model is corrected, including one line in `docs/mock-usage.md` that was already wrong before this change: it claimed `parent_id` "is accepted but not used to filter (the mock's tags are a flat collection)", which stopped being true when nesting landed. Unit tests that used colon-form names as arbitrary opaque strings were rewritten to bare names, so the next person copying a nearby test does not reintroduce the prefix.

## 4. Required information
- Spec: https://github.com/adobe/serenity-docs/pull/39 — dimension-root tag model (open; the design of record for this program, not yet merged)
- Implementation plan: https://github.com/adobe/serenity-docs/pull/39 — dimension-root tags implementation plan, WP1 (same PR)
- Other: https://github.com/adobe/serenity-docs/issues/44 — onboarding-generated prompts arrive without a category (gap found while scoping this program) · https://github.com/adobe/spacecat-shared/issues/1760 and https://github.com/adobe/spacecat-shared/issues/1756 — both closed; this builds on the nesting and URL-safe-id work they landed

- **spacecat-api-service** — consumes this package and its GHCR mock image. Its integration harness resolves the mock image tag from the installed client version with no fallback, so it cannot bump the dependency until this releases. It is also the first consumer of the corrected `path[]` breadcrumb, since it will resolve a tag's dimension as `path[0]`.
- **project-elmo-ui** — contract. Groups a prompt's tags by `path[0].id` once api-service forwards parentage; depends transitively on the breadcrumb being a true ancestry rather than the direct parent.
- **mysticat-data-service** — contract. Its migration CLI writes and reshapes this same taxonomy directly against live Semrush, and relies on the same `(parent, name)` identity and the child-`PATCH` `parent_id` rule this mock now encodes.

Drove the reseeded store through the stateful ops directly to confirm the behaviours the suite asserts are real rather than tautological:

- Deleting the `type` dimension's `branded` value detached it from the seeded prompt and left the prompt's four other tags intact.
- Deleting the sub-category `human` removed only that tag; the `source` value also named `human` survived untouched, confirming the two are distinct rows that differ solely by parent.
- Deleting every tag the prompt carried left the prompt present and fully unassigned, not deleted and not orphaned.
- Walking a deliberately constructed parent cycle through the breadcrumb builder terminated immediately rather than looping, and a tag whose parent had been deleted read back as a root, with no `path` key, rather than throwing.

Confirmed every behaviour this PR encodes against the real Project Engine API, in a scratch tag subtree of a test workspace (created, probed, deleted; no prompts attached). A depth-3 tag reads back `path[]` as a two-element root-first ancestry whose `path[0]` is the dimension root, while a depth-2 tag reads back a single element — so `path[0]` identifies the dimension at any depth, and the breadcrumb this mock now returns is the one live returns. The same subtree confirms that the same bare name under two different parents yields two distinct tags, that re-creating an existing `(parent, name)` answers 500, that `search` from the root level does not descend, and that a `PATCH` which omits `parent_id` promotes the tag to a root. Reading the tree back requires `draft=true`: tag writes land in the project's draft layer and the default listing serves the live view.

Fetching one live tag id both ways — embedded on a prompt, and from the tag tree — returned two equal objects, which is what establishes the single serializer. The asymmetry it replaces is easy to miss by sampling: in the probed project 32 of the 40 tags embedded on prompts are roots, so an inspection that happens to draw only roots sees no `parent_id` anywhere and concludes the embedded shape is narrower than the tree's. The e2e suite therefore asserts equality between the two reads rather than the absence of a field.

Separately, the e2e suite on `main` reports two failures until `build/openapi3.json` is regenerated (`npm run spec:convert && npm run spec:overlay`); both are stale-generated-artifact failures in request validation, unrelated to this change. Regenerating clears them, which is how the baseline for this branch was established.

## 7. Test plan
**Verified locally, beyond the automated suites.** Ran the e2e suite against a real Counterfact mock server (`npm run test:e2e`, which boots `mock/run.js` over HTTPS), exercising the endpoints through the generated typed client rather than against the in-memory ops. New end-to-end coverage pins the behaviours this PR turns on: that a tag embedded on a prompt is deep-equal to the object the tag tree returns for that same id, and that an embedded root carries neither `parent_id` nor `path`; that `search` filters within one level and never descends, so a nested tag is invisible from the root level; that a depth-3 sub-category reads back a two-leaf, root-first breadcrumb whose deeper leaf carries `parent_id`; that a sub-category and a same-named `source` value remain distinct tags across the wire; that deleting a tag detaches it from the prompts carrying it and drops it out of `by_tags`; and that a prompt whose last tag is deleted survives fully unassigned.

**Verification by environment.** This package has no deployed runtime of its own, so there is nothing to check on eph/dev/stage/prod for this PR directly. Verification is downstream and gated on the release:

1. On merge, confirm semantic-release publishes the package (a `feat:` commit takes it to the next minor) and that the GitHub Release event publishes the matching GHCR mock image tagged with that exact version. There is no `latest` tag and no fallback.
2. Confirm the image exists before any consumer bumps the dependency: `docker manifest inspect ghcr.io/adobe/spacecat-shared-project-engine-client-mock:<version>`.
3. In spacecat-api-service, once the dependency is bumped, the `it-postgres` job pulls that image by the installed client version. A green run there is the real integration check that this mock's tree shape, breadcrumb and delete-detach semantics are what the proxy expects.

This is the first merge in the dimension-root tag model program, and the rest of the program is blocked on the artifacts it publishes.

- **blocks** — spacecat-api-service (WP2, dimension-root tag model). It cannot bump `@adobe/spacecat-shared-project-engine-client` until this releases, because its integration harness resolves the mock image tag from the installed client version with `${SERENITY_PE_MOCK_TAG:?}` and no fallback. A bump to an unpublished version fails the compose pull hard, for everyone on `main`.
- **blocks** — project-elmo-ui (WP3) and mysticat-data-service (WP4), both transitively through WP2.
- **related** — https://github.com/adobe/serenity-docs/pull/39, the model spec and implementation plan. It is the design of record and is still open; this PR does not wait on it.

Ordered sequence:

1. Merge this PR. Semantic-release publishes the npm package; the GitHub Release event publishes the GHCR mock image at the same version.
2. Confirm the mock image exists in GHCR.
3. Only then, WP2 bumps the client dependency in spacecat-api-service, regenerates its lockfile, and merges.
4. WP3 and WP4 follow WP2 and are independent of each other.


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-structured model change with thorough tests and honest documentation of known limitations.
**Complexity: HIGH** - large diff; API surface + tests-heavy.
_Changes: replaces the mock's flat colon-prefixed tag model with a parent-keyed dimension-root tree, updating tag identity derivation, ancestry breadcrumbs, delete-detach behavior, seeds, and tests (15 files)._

<details>

- nit: tests use `deep.equal` on root-name lists where the PR's own docs say "Order matters only for readability - the tree is reconstructed from `parent_id`, not from position." Consider `.to.have.members` for root-presence assertions to decouple test stability from seed insertion order - `test/e2e/project-engine-mock.e2e.js:789`
- nit: `tagId('', 'ab')` in the boundary-shift test exercises an empty NAME that is not a valid system input; a realistic pair like `tagId('ab', 'c')` vs `tagId('b', 'ca')` would be equally rigorous and document real-world expectations - `test/mock/tag-id.test.js:56`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 6s | <b>Cost:</b> $5.53 | <b>Commit:</b> <code>beb6739b7c6c787b8521f90a32a875c406dac5f3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior nits addressed cleanly, correctness fix is well-structured and thoroughly tested.
**Complexity: HIGH** - large diff; API surface + tests-heavy.
_Changes: adds a shared tag serializer so tags embedded on prompts are byte-identical to tags from the tree read, completing the dimension-root model's single-source-of-truth contract (20 files)._

- `deep.equal` on root-name lists no longer couples tests to seed insertion order (now `.to.have.members`)
- `tagId('', 'ab')` boundary test replaced with a realistic pair (`tagId('ab', 'c') !== tagId('b', 'ca')`)

<details>

- nit: the e2e test "embeds a ROOT prompt tag with no parent_id and no path" wraps its core assertion in `if (rootTag)`, but the seeded prompt carries only descendants, so `rootTag` is always undefined and the named assertion never fires. The broader `forEach` loop and the unit-level `tag-view.test.js` still cover the invariant, but the dead branch gives false confidence. Consider `expect(rootTag).to.exist` or seeding a prompt that carries a root - `test/e2e/project-engine-mock.e2e.js:998`
- nit: `byId.get(t.id) ?? t` in `by_tags.js` falls back to the raw `{ id, name }` stub when a prompt references a tag id absent from the standalone collection. The fallback is safe but produces an object missing `children_count`/`prompts_count`/`path` - the non-conformant shape would not be caught by the single-serializer e2e assertion. A comment noting the degraded shape for `tagged.js`-only tags would help - `mock/counterfact/routes/.../aio/prompts/by_tags.js:133`
- suggestion: `prompts_count` is read by `buildTagView` but never maintained by any write path (pre-existing gap, not introduced here). Deriving it at read time from the prompts collection would align with the "derive at read, never store" philosophy the rest of the serializer embodies - `mock/tag-view.js:92`

</details>

---

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 56s | <b>Cost:</b> $7.04 | <b>Commit:</b> <code>0bf0a24f9bdc57aad4f15aa350d2d8ecd04eff4c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior nits addressed, correctness fix for root-tag registration is clean and well-tested.
**Complexity: HIGH** - large diff; API surface + tests-heavy.
_Changes: fixes the dead ROOT-embedding test, registers minted root tags in `prompts/tagged` so the single-serializer invariant holds for name-keyed creates, and hoists shared test helpers (20 files)._

- Dead `if (rootTag)` branch replaced with proper setup + unconditional assertions
- `byId.get(t.id) ?? t` fallback now documented with invariant comment; the two create paths both register their tags, making the fallback unreachable in normal flow
- `prompts_count` read-time derivation declined by author (won't encode unverified live semantics) - accepted

<details>

- nit: `tagged.js` pre-filters stored ids before calling `upsertMany`, which itself rejects collisions - belt-and-suspenders that is defensible in a mock aiming for correctness over brevity, but the two guards can never diverge since they read the same store synchronously - `mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/prompts/tagged.js:79`
- nit: `tagged.js` calls `ops.tags.list(scope)` a second time per request (first is `ops.prompts.list`). Negligible in a test mock but worth noting if the collection ever grows large under stress - `mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/prompts/tagged.js:78`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 49s | <b>Cost:</b> $5.22 | <b>Commit:</b> <code>54c523975a4d65938c2e26c6fffb6c75f9b3d8f2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior nits addressed, the two new commits (tagged.js pre-filter coverage, e2e helpers hoist) are clean and well-motivated.
**Complexity: HIGH** - large diff; API surface + tests-heavy.
_Changes: completes the dimension-root mock by registering tags minted through `prompts/tagged`, covering the pre-filter guard with a mixed-batch e2e test, and hoisting shared request helpers to suite scope (20 files)._

- `tagged.js` pre-filter belt-and-suspenders concern declined with load-bearing test proving the guard prevents atomic-batch poisoning
- Double `ops.tags.list(scope)` read declined (both in-memory synchronous, no shared intermediate)

<details>

- nit: `by_tags.js` calls `buildTagView` unconditionally even when `matched` is empty - a free early-out (`if (!matched.length) return ...`) would skip the index build for the common empty-project case - `mock/counterfact/routes/v2/workspaces/{id}/projects/{project_id}/aio/prompts/by_tags.js:147`
- suggestion: pin the exact byte output of `tagId('category')` and `tagId('Running Shoes', CATEGORY_ROOT_TAG_ID)` in a snapshot assertion so a future hash-preimage refactor fails locally rather than cascading to N downstream repos that resolve the mock image tag from the installed client version - `test/mock/tag-id.test.js`

</details>

---

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 0s | <b>Cost:</b> $5.81 | <b>Commit:</b> <code>ac6eeb507f14388dd57154576d98cc084dfe1c30</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - prior suggestion addressed cleanly, no new concerns.
**Complexity: HIGH** - large diff; API surface + tests-heavy.
_Changes: adds a golden-value pinning test for the tag-id derivation to prevent silent cross-repo contract breakage (1 file)._

- Pin exact byte output of `tagId` for a root and a child so a hash-preimage refactor fails locally rather than cascading downstream (done in `3008583c`, with a thorough comment explaining the coupling mechanism and blast radius)
- `by_tags.js` unconditional `buildTagView` on empty `matched` - declined by author (both reads in-memory synchronous, no shared intermediate); accepted

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 41s | <b>Cost:</b> $3.03 | <b>Commit:</b> <code>3008583cceee7e467456a29d1a1925ceb3aed134</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - trivial fixture rename aligning type-test examples with the bare-name model; no concerns.
**Complexity: HIGH** - large diff; API surface.
_Changes: renames type-test fixture strings from colon-prefixed (`category:X`) to bare names (`Running Shoes`, `category`), aligning with the dimension-root model (1 file)._

- All prior findings addressed or declined in earlier commits (confirmed through reviews 1-5)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 5s | <b>Cost:</b> $1.28 | <b>Commit:</b> <code>6547cb26ce1897112f0c558b641468d56843aa05</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1802, opened 2026-07-10
feat(tokowaka-client): offload covered-marking saves to import worker queue above threshold

khushboo5723:
## What / why

`saveSuggestions()` falls back to an unthrottled `Promise.allSettled(suggestions.map(s => s.save()))` whenever it has to persist more than `PARALLEL_SAVE_THRESHOLD` (1700) suggestions and no queue is available. That's the root cause of the production failure `"4362 of 5357 suggestions failed to save"`.

This is a narrow, surgical fix: only the `> 1700` branch of the existing `saveSuggestions()` changes. The synchronous covered-suggestion computation (regex matching against `allSuggestions` to figure out *which* suggestions are covered/uncovered) is untouched — only the final **save** step is offloaded, and only past the existing threshold.

## Change

- `saveSuggestions(dataAccess, suggestions, queueContext)` — when `queueContext` is provided and `suggestions.length > 1700`, enqueues a `suggestion-bulk-update` job (`siteId`, `opportunityId`, `suggestionIds`, `set`/`unset`, `updatedBy`) to the import worker queue instead of saving directly. No `queueContext` → falls back to today's behavior unchanged.
- `cleanupCoveredSuggestions()` accepts and passes through `queueContext` unchanged.
- `TokowakaClient.createFrom(context)` now also picks up `sqs` / `IMPORT_WORKER_QUEUE_URL` from context (already available via existing middleware — zero changes needed in consuming services).
- Both call sites — `#deployPatternSuggestion` (marking) and `rollbackSuggestions` (unmarking) — build and pass the `queueContext` with the exact field/value semantics as today's synchronous marking/unmarking, including the asymmetric field-strip behavior on rollback.
- The worker is intentionally agnostic: it only receives suggestion IDs plus what to `set`/`unset` — no domain-wide/pattern business logic.

## Follow-up

The import-worker handler for the `suggestion-bulk-update` job type (fetch by id, apply set/unset, save) is a separate, subsequent PR.

## Tests

959 passing, 100% lines/branches/statements/functions on `spacecat-shared-tokowaka-client`; lint clean.

### DIFF
--- a/packages/spacecat-shared-tokowaka-client/src/index.js
+++ b/packages/spacecat-shared-tokowaka-client/src/index.js
@@ -1090,7 +1090,6 @@ class TokowakaClient {
             sqs: this.sqs,
             queueUrl: this.importWorkerQueueUrl,
             siteId: site.getId(),
-            opportunityId: opportunity.getId(),
             unset: fieldsToStrip,
             updatedBy: updatedBy ?? coveredFallback,
             log: this.log,
@@ -1522,7 +1521,6 @@ class TokowakaClient {
    * @param {string} updatedBy
    * @param {Array} coveredSuggestions - Accumulator (mutated)
    * @param {string} siteId
-   * @param {string} opportunityId
    * @returns {Promise<void>} Throws on metaconfig upload failure
    * @private
    */
@@ -1536,7 +1534,6 @@ class TokowakaClient {
     updatedBy,
     coveredSuggestions,
     siteId,
-    opportunityId,
   ) {
     const data = suggestion.getData();
     const coverageField = data?.isDomainWide ? 'coveredByDomainWide' : 'coveredByPattern';
@@ -1611,7 +1608,6 @@ class TokowakaClient {
           sqs: this.sqs,
           queueUrl: this.importWorkerQueueUrl,
           siteId,
-          opportunityId,
           set: { [coverageField]: suggestion.getId() },
           updatedBy,
           log: this.log,
@@ -1709,7 +1705,6 @@ class TokowakaClient {
             updatedBy,
             coveredSuggestions,
             site.getId(),
-            opportunity.getId(),
           );
           deployedPatternSuggestions.push(suggestion);
         } catch (error) {
--- a/packages/spacecat-shared-tokowaka-client/src/index.js
+++ b/packages/spacecat-shared-tokowaka-client/src/index.js
@@ -50,6 +50,7 @@ import {
 
 export { FastlyKVClient } from './fastly-kv-client.js';
 export { calculateForwardedHost } from './utils/custom-html-utils.js';
+export { SUGGESTION_BULK_UPDATE_TYPE } from './utils/suggestion-utils.js';
 
 // CloudFront control-plane (free functions + constants).
 export {
@@ -88,6 +89,7 @@ class TokowakaClient {
     const {
       TOKOWAKA_SITE_CONFIG_BUCKET: bucketName,
       TOKOWAKA_PREVIEW_BUCKET: previewBucketName,
+      IMPORT_WORKER_QUEUE_URL: importWorkerQueueUrl,
     } = env;
 
     if (context.tokowakaClient) {
@@ -100,6 +102,8 @@ class TokowakaClient {
       s3Client: s3?.s3Client ?? context.s3Client,
       env,
       dataAccess: context.dataAccess,
+      sqs: context.sqs,
+      importWorkerQueueUrl,
     }, log);
     context.tokowakaClient = client;
     return client;
@@ -114,10 +118,13 @@ class TokowakaClient {
    * @param {Object} config.env - Environment variables (for CDN credentials)
    * @param {Object} [config.dataAccess] - Data access layer
    *   (provides Suggestion.saveMany for batch saves)
+   * @param {Object} [config.sqs] - SQS client (sendMessage(queueUrl, payload)). Used to offload
+   *   very large covered-suggestion saves to the import worker instead of saving them directly.
+   * @param {string} [config.importWorkerQueueUrl] - Import worker SQS queue URL.
    * @param {Object} log - Logger instance
    */
   constructor({
-    bucketName, previewBucketName, s3Client, env = {}, dataAccess,
+    bucketName, previewBucketName, s3Client, env = {}, dataAccess, sqs, importWorkerQueueUrl,
   }, log) {
     this.log = log;
 
@@ -134,6 +141,8 @@ class TokowakaClient {
     this.s3Client = s3Client;
     this.env = env;
     this.dataAccess = dataAccess;
+    this.sqs = sqs;
+    this.importWorkerQueueUrl = importWorkerQueueUrl;
 
     this.mapperRegistry = new MapperRegistry(log);
     this.cdnClientRegistry = new CdnClientRegistry(env, log);
@@ -1077,7 +1086,15 @@ class TokowakaClient {
             this.log.info(`[edge-rollback] Cleaning ${covered.length} covered suggestion(s) for pattern ${suggestion.getId()} (isDomainWide=${isDomainWide}, fallback=${coveredFallback})`);
           }
           // eslint-disable-next-line no-await-in-loop, max-len
-          await cleanupCoveredSuggestions(this.dataAccess, covered, coveredFallback, updatedBy, fieldsToStrip, this.log);
+          await cleanupCoveredSuggestions(this.dataAccess, covered, coveredFallback, updatedBy, fieldsToStrip, this.log, {
+            sqs: this.sqs,
+            queueUrl: this.importWorkerQueueUrl,
+            siteId: site.getId(),
+            opportunityId: opportunity.getId(),
+            unset: fieldsToStrip,
+            updatedBy: updatedBy ?? coveredFallback,
+            log: this.log,
+          });
         }
       }
     }
@@ -1504,6 +1521,8 @@ class TokowakaClient {
    * @param {Array} allSuggestions - Full opportunity suggestion list
    * @param {string} updatedBy
    * @param {Array} coveredSuggestions - Accumulator (mutated)
+   * @param {string} siteId
+   * @param {string} opportunityId
    * @returns {Promise<void>} Throws on metaconfig upload failure
    * @private
    */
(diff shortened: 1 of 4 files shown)


packages/spacecat-shared-tokowaka-client/src/utils/suggestion-utils.js
```diff
@@ -92,26 +92,58 @@ export function filterEligibleSuggestions(suggestions, mapper) {
...
+          ...(unset && { unset }),
+          updatedBy,
+        });
+        log.info(
+          `[suggestion-bulk-update] Queued bulk update for ${suggestions.length} suggestion(s)`,
+        );
+      } catch (error) {
+        log.warn(`[suggestion-bulk-update] Failed to queue bulk update: ${error.message}`);
+      }
+      return;
```
MysticatBot: issue (blocking): **Silent data loss when queue path fails.** Two failure modes silently drop all suggestions without persisting them:

1. When `sendMessage` rejects, the catch block logs a warning and falls through to `return;` without attempting the existing `Promise.allSettled` fallback.
2. When `queueUrl` is missing (line 125), the function logs a warning and returns immediately.

In both cases, callers believe persistence succeeded (no throw), but nothing is saved. The `#deployPatternSuggestion` caller even pushes to `coveredSuggestions` afterward, so downstream code treats the suggestions as persisted. This is a regression from the current behavior: today, `Promise.allSettled` at least partially succeeds and surfaces errors for the failures.

**Why it matters:** The PR fixes a production failure where suggestions failed to save. The new code introduces a configuration-dependent path where they fail to save *with no error signal at all*.

**Fix:** On `sendMessage` failure or missing `queueUrl`, fall through to the existing `Promise.allSettled` path rather than returning. This preserves the pre-existing (imperfect but functional) behavior as a degraded-mode fallback:

```js
if (!queueUrl) {
  log.warn(...);
  // fall through to Promise.allSettled below
} else {
  try {
    await sqs.sendMessage(...);
    log.info(...);
    return;
  } catch (error) {
    log.error(`[suggestion-bulk-update] Failed to queue, falling back to direct save: ${error.message}`);
    // fall through to Promise.allSettled below
  }
}
```

packages/spacecat-shared-tokowaka-client/src/utils/suggestion-utils.js
```diff
@@ -92,26 +92,58 @@ export function filterEligibleSuggestions(suggestions, mapper) {
...
+        );
+        return;
+      }
+
+      try {
+        await sqs.sendMessage(queueUrl, {
+          type: SUGGESTION_BULK_UPDATE_TYPE,
+          siteId,
+          opportunityId,
+          suggestionIds: suggestions.map((s) => s.getId()),
```
MysticatBot: issue (blocking): **SQS message payload has no size bound.** The `suggestionIds` array can contain 5000+ UUIDs (the production incident involved 5357 suggestions). Each UUID is 36 bytes; 5000 IDs plus JSON overhead is ~200KB. SQS has a 256KB message size limit. For larger opportunity sets - the exact scenario this fix targets - the payload will exceed the limit and `sendMessage` will throw, triggering the silent-drop path above.

**Fix:** Either chunk the IDs into batches that fit within a safe limit (e.g., 4000 IDs per message), or use S3 as an intermediary (write the IDs to S3, send only the S3 key in the SQS message). At minimum, validate the serialized payload size before sending.

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1077,7 +1086,15 @@ class TokowakaClient {
             this.log.info(`[edge-rollback] Cleaning ${covered.length} covered suggestion(s) for pattern ${suggestion.getId()} (isDomainWide=${isDomainWide}, fallback=${coveredFallback})`);
           }
           // eslint-disable-next-line no-await-in-loop, max-len
-          await cleanupCoveredSuggestions(this.dataAccess, covered, coveredFallback, updatedBy, fieldsToStrip, this.log);
+          await cleanupCoveredSuggestions(this.dataAccess, covered, coveredFallback, updatedBy, fieldsToStrip, this.log, {
+            sqs: this.sqs,
```
MysticatBot, line 1090: issue (blocking): **Unconditional `queueContext` construction when `this.sqs` may be undefined.** Both call sites (here and the deploy path at line ~1607) always construct a `queueContext` object with `sqs: this.sqs`. If `context.sqs` was never injected (older service versions, non-import-worker deployments, or test environments), `this.sqs` is `undefined`. The `queueContext` object is still truthy, so `saveSuggestions` enters the queue branch and attempts `sqs.sendMessage()`, which throws a `TypeError`. That TypeError is caught by the warn-and-return path, leading to silent data loss.

**Fix:** Only pass `queueContext` when both dependencies are present:

```js
const queueContext = (this.sqs && this.importWorkerQueueUrl) ? {
  sqs: this.sqs,
  queueUrl: this.importWorkerQueueUrl,
  siteId: site.getId(),
  opportunityId: opportunity.getId(),
  unset: fieldsToStrip,
  updatedBy: updatedBy ?? coveredFallback,
  log: this.log,
} : undefined;
```

MysticatBot (changes requested):
Hey @khushboo5723,

**Verdict: Request changes** - silent data loss paths when queue is unavailable or sendMessage fails.
_Changes: Adds queue-based offloading to `saveSuggestions()` for bulk suggestion updates above 1700 items, routing to an import worker via SQS instead of direct saves (3 files)._

## Must fix before merge

1. [Critical] Silent data loss when queue path fails - neither queued nor directly saved - `suggestion-utils.js:144` (details inline)
2. [Important] SQS message payload has no size bound - 5000+ UUIDs can exceed 256KB limit - `suggestion-utils.js:133` (details inline)
3. [Important] Unconditional `queueContext` construction when `this.sqs` may be undefined - `index.js:1090` (details inline)

<details>

- nit: `updatedBy` in the deploy-path `queueContext` has no fallback (rollback path uses `updatedBy ?? coveredFallback` but deploy path passes raw `updatedBy` which may be undefined) - `index.js:1614`
- suggestion: Add boundary tests at exactly 1700 items (should take `saveMany` path) and 1701 without `queueContext` (should take `Promise.allSettled` path) to pin threshold behavior - `suggestion-utils.test.js:228`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 0s | <b>Cost:</b> $4.21 | <b>Commit:</b> <code>438d0a8b6a492dcee9e017cf86b03831a6b2c33b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-tokowaka-client/src/utils/suggestion-utils.js
```diff
@@ -92,26 +92,68 @@ export function filterEligibleSuggestions(suggestions, mapper) {
...
   if (suggestions.length > PARALLEL_SAVE_THRESHOLD) {
+    if (queueContext) {
+      const {
+        sqs, queueUrl, siteId, set, unset, updatedBy, log,
+      } = queueContext;
+
+      if (queueUrl && sqs) {
+        try {
+          await sqs.sendMessage(queueUrl, {
+            type: SUGGESTION_BULK_UPDATE_TYPE,
```
MysticatBot, line 133: issue (blocking): **SQS message payload has no size bound.** The `suggestionIds` array is serialized directly into a single SQS message with no size guard. SQS enforces a hard 256 KB message body limit. At ~40 bytes per UUID (36 chars + JSON overhead), ~6,500 IDs will breach the limit. The production incident that motivated this PR involved 5,357 suggestions (~203 KB) - that fits today, but growth past ~6,500 suggestions will cause `sendMessage` to throw, falling back to the same `Promise.allSettled` direct-save path that already fails at scale.

The fix is straightforward - chunk the ID array into batches of ~5,000 per message:

```javascript
const CHUNK_SIZE = 5000;
for (let i = 0; i < suggestions.length; i += CHUNK_SIZE) {
  const chunk = suggestions.slice(i, i + CHUNK_SIZE);
  await sqs.sendMessage(queueUrl, {
    type: SUGGESTION_BULK_UPDATE_TYPE,
    siteId,
    suggestionIds: chunk.map((s) => s.getId()),
    ...(set && { set }),
    ...(unset && { unset }),
    updatedBy,
  });
}
```

The worker already processes a partial list of IDs, so multiple messages with the same `type`/`siteId`/`set`/`unset` are semantically idempotent.

*Re-raised from prior review - still unresolved.*

MysticatBot (dismissed):
Hey @khushboo5723,

**Verdict: Request changes** - unbounded SQS payload size remains unresolved and will fail at production scale.
_Changes: Adds queue-based offloading to `saveSuggestions()` for bulk suggestion updates above 1700 items, routing to an import worker via SQS instead of direct saves (4 files)._

## Must fix before merge

1. [Important] SQS message payload has no size bound - 5000+ UUIDs can exceed 256KB limit - `suggestion-utils.js:133` (details inline)

<details>

- suggestion: Gate `queueContext` construction on `this.sqs && this.importWorkerQueueUrl` to eliminate warn spam in environments without SQS configured - `index.js:1090`
- suggestion: Add boundary test at exactly 1700 items with a valid `queueContext` to pin the threshold as `>` not `>=` - `suggestion-utils.test.js`
- nit: Deploy path pushes entities to `coveredSuggestions` after queue enqueue (not after persistence) - add a comment documenting the eventual-consistency trade-off - `index.js:1614`
- nit: `log` is passed both as the 6th positional arg to `cleanupCoveredSuggestions` and inside `queueContext` - redundant, consider removing from the context object - `index.js:1093`
</details>

- Silent data loss on queue failure - now falls back to direct save with error logging (commit 333216da)
- Unconditional queueContext when `this.sqs` undefined - now handled gracefully via warn + fallback to direct save

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 37s | <b>Cost:</b> $6.96 | <b>Commit:</b> <code>333216dae9d6412172c00ec121694e92989a881a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dipratap (approved):

PR #1803, opened 2026-07-10
[SITES-47884] fix: cover init() in variation() try/catch so LaunchDarkly outage fails open

ravverma:
## Problem

[SITES-47884](https://jira.corp.adobe.com/browse/SITES-47884): During a LaunchDarkly outage, `/auth/login` fails outright for effectively all users — not degraded latency, but an outright login outage caused by a third-party dependency with no fail-open path.

Root cause: in `spacecat-shared-launchdarkly-client`, `variation()` called `await this.init()` **outside** its own try/catch:

```js
async variation(flagKey, context, defaultValue) {
  await this.init();          // NOT inside the try/catch below
  try {
    ...
  } catch (error) {
    return defaultValue;      // fallback only covered this block
  }
}
```

`init()`'s `waitForInitialization({ timeoutSeconds: 5 })` **rejects** when LD is unreachable. That rejection was not caught by `variation()`, so it propagated out of `isFlagEnabledForIMSOrg()` and up through the login path — even though `isFlagEnabledForIMSOrg` is documented/expected to fail safe.

## Fix

Move `await this.init()` **inside** `variation()`'s try/catch so any initialization error returns `defaultValue` instead of throwing. This is the root defect called out in the ticket; `isFlagEnabledForIMSOrg` now genuinely fails safe during an LD outage.

## Testing

- Added a unit test asserting `variation()` returns the default value (and never calls `client.variation`) when initialization fails.
- All 46 tests in the package pass; lint clean.

### DIFF
--- a/packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
+++ b/packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js
@@ -148,9 +148,12 @@ class LaunchDarklyClient {
    * @returns {Promise<*>} The evaluated flag value
    */
   async variation(flagKey, context, defaultValue) {
-    await this.init();
-
     try {
+      // init() is inside the try/catch so that a LaunchDarkly outage (e.g. the
+      // initialization timeout rejecting) fails open and returns defaultValue,
+      // rather than propagating and taking down callers such as /auth/login.
+      await this.init();
+
       const value = await this.client.variation(flagKey, context, defaultValue);
       this.log.debug(`Flag "${flagKey}" evaluated to:`, value);
       return value;
--- a/packages/spacecat-shared-launchdarkly-client/test/index.test.js
+++ b/packages/spacecat-shared-launchdarkly-client/test/index.test.js
@@ -434,6 +434,29 @@ describe('LaunchDarklyClient', () => {
       // Reset for other tests
       mockClient.variation.resolves(true);
     });
+
+    it('should return default value when initialization fails (LD outage)', async () => {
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @ravverma,

**Verdict: Approve** - clean, minimal fix that correctly addresses the outage scenario.
**Complexity: TRIVIAL** - 2 files, <20 lines, single-package bug fix.
_Changes: Moves `await this.init()` inside `variation()`'s existing try/catch so LaunchDarkly initialization failures return `defaultValue` instead of propagating (2 files)._

<details>

- suggestion: Consider logging init failures at `warn` instead of `error` - since this is now a designed fail-open path, sustained LD outages will produce a continuous `error` stream that may trigger volume-based alerts unnecessarily - `packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js:155`
- suggestion: Document (in the JSDoc or a conventions note) that callers of `variation()` must choose a restrictive/safe `defaultValue` for security-sensitive flags, since an LD outage will silently fall back to it - `packages/spacecat-shared-launchdarkly-client/src/launchdarkly-client.js:148`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 19s | <b>Cost:</b> $2.20 | <b>Commit:</b> <code>03fa4b45249d5fe26588adc22c43e9c178669c2a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1828, opened 2026-07-20
feat(data-access): canonical v2 changeDetails schema + validator for FixEntity (SITES-47997)

alinarublea:
## What

Slice 1 of implementing ADR [adobe/mysticat-architecture#200](https://github.com/adobe/mysticat-architecture/blob/main/platform/decisions/deploy-action-provenance-and-verify.md) (*deploy-action, provenance & universal pre/post-verify*, **SITES-47997**): a **machine-readable, validated v2 shape** for `FixEntity.changeDetails` — the structured deploy-action record (*who/when*, *what/which-pages proposed*, *result of the call*).

- **New `change-details.schema.js`** — a Joi schema (`changeDetailsV2Schema`), the `validateChangeDetails` validator, and the enum bundle (`SURFACES`, `ACTOR_TYPES`, `CALL_STATUSES`, `APPLIED`, `CHANGE_RESULT_STATUSES`, `VERIFY_VERDICTS`). Encodes the ADR's blocking constraints:
  - `applied` is the enum **`ALL | PARTIAL | NONE`** (never a boolean)
  - `callStatus` includes **`no_op`** — the case SITES-47077 / SITES-47078 silently mis-read as success
  - `deployResponsePayload` **≤ 4 KB** cap + `deployResponseSha256` hex integrity hash
  - **bidirectional `(targetPath, property)` key-matching** between `target.changes` (intent) and `result.changeResults` (outcome): no orphan outcomes, and `applied: ALL` must record every proposed change
- **Wired into the `changeDetails` attribute validator**, **reader-tolerant**: records without `schemaVersion: 2` keep the pre-existing non-empty-object guard. Additive — the column stays `type: 'any'`, **no migration**. Validation runs on the create path (`#validateItem`), consistent with the other attribute validators.
- **Exposes `FixEntity.CHANGE_DETAILS`** (enum bundle on the model, to keep generic names off the package-root barrel); exports `validateChangeDetails` / `changeDetailsV2Schema` / `CHANGE_DETAILS_SCHEMA_VERSION` at the package root for the SITES-47091 transition guard, the writers (autofix-worker), and mystique to consume.

## Why

The deploy action was an unvalidated freeform `changeDetails` bag with camelCase-vs-snake_case drift across two runtimes and no defined "result of the call" — the systemic root of the confirmed false positives in SITES-47077 / SITES-47078. This makes the shape first-class and validated so those become impossible-by-construction once the guard gates on `result` (later slice).

## Scope / next

This is **slice (a)** of the ADR rollout. Deliberately **not** in this PR:
- `publishedBy` column + `executedBy/At` → `deployedBy/At` rename (dual-write) — needs a companion mysticat-data-service migration
- Gating the `DEPLOYED`/`PUBLISHED` transition on `result` presence — extends the merged SITES-47091 guard
- SITES-47869 status-vocab reconciliation

## Testing

`change-details.schema.test.js` — **100% line/branch coverage**; full fix-entity + suggestion suites green (264 passing), lint clean.

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/fix-entity/change-details.schema.js
+++ b/packages/spacecat-shared-data-access/src/models/fix-entity/change-details.schema.js
@@ -1,0 +1,253 @@
+/*
+ * Copyright 2025 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import Joi from 'joi';
+import { isNonEmptyObject } from '@adobe/spacecat-shared-utils';
+
+/**
+ * Canonical v2 shape for `FixEntity.changeDetails` — the structured, validated
+ * "deploy action" record defined by ADR
+ * adobe/mysticat-architecture#200 (platform/decisions/deploy-action-provenance-and-verify.md,
+ * SITES-47997): *who* deployed/published, *when*, *what entity/property* is
+ * proposed to change, *which pages* are affected, and *the result of the call*.
+ *
+ * The DB column stays `type: 'any'` (no destructive migration); this module is
+ * the machine-readable schema the JS write chokepoint validates against, so the
+ * freeform bag cannot silently drift across the two runtimes (autofix-worker JS
+ * + mystique Python). Records without `schemaVersion === 2` are legacy freeform
+ * (v1) and are intentionally NOT schema-validated — the migration is additive
+ * and reader-tolerant (ADR "V1 / V2 backwards compatibility").
+ */
+
+export const CHANGE_DETAILS_SCHEMA_VERSION = 2;
+
+// Deploy CHANNEL / "where" the fix went live.
+export const SURFACES = {
+  ASO: 'ASO',
+  AUTHOR_PUBLISH: 'AUTHOR_PUBLISH',
+  GIT_MERGE: 'GIT_MERGE',
+  SYSTEM: 'SYSTEM',
+};
+
+// WHO acted. Orthogonal to `surface` — DETECTOR is an actor (an audit that
+// observed a live fix), not a deploy channel, so it is an actorType not a surface.
+export const ACTOR_TYPES = {
+  IMS_USER: 'IMS_USER',
+  SERVICE: 'SERVICE',
+  GITHUB: 'GITHUB',
+  DETECTOR: 'DETECTOR',
+  UNKNOWN: 'UNKNOWN',
+};
+
+// Classified outcome of the deploy/apply call — transport-agnostic (JCR writes
+// and git ops classify the same way HTTP does). Set independently by the
+// transport adapter; NOT derived from `changeResults`. `no_op` is the case
+// SITES-47077 / SITES-47078 silently mis-read as success.
+export const CALL_STATUSES = {
+  SUCCESS: 'success',
+  NO_OP: 'no_op',
+  CLIENT_ERROR: 'client_error',
+  SERVER_ERROR: 'server_error',
+  TIMEOUT: 'timeout',
+};
+
+// Whole-action roll-up of the per-change `changeResults[].status`. NOT a boolean
+// — a deploy action may write >1 property, so "2-of-3" is first-class.
+export const APPLIED = {
+  ALL: 'ALL',
+  PARTIAL: 'PARTIAL',
+  NONE: 'NONE',
+};
+
+// Per-change outcome, key-matched to `target.changes` by (targetPath, property).
+export const CHANGE_RESULT_STATUSES = {
+  APPLIED: 'applied',
+  UNCHANGED: 'unchanged',
+  FAILED: 'failed',
+};
+
+// Verify verdict. Distinguishes NEGATIVE (`rejected` — verify ran, the change did
+// not render as expected → revert-eligible) from ERRORED (`verify_failed` —
+// verify could not complete → MUST NOT auto-revert). Mirrors the post-verify POC.
+export const VERIFY_VERDICTS = {
+  VERIFIED: 'verified',
+  REJECTED: 'rejected',
+  INCONCLUSIVE: 'inconclusive',
+  VERIFY_FAILED: 'verify_failed',
+};
+
+// `deployResponsePayload` size cap (bytes). MUST hold on a shared DB column — one
+// runaway payload degrades read latency for every consumer. Writers hash-first,
+// then cap/redact, so `deployResponseSha256` always covers the pre-redact payload.
+export const DEPLOY_RESPONSE_PAYLOAD_MAX_BYTES = 4096;
+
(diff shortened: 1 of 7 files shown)


packages/spacecat-shared-data-access/src/models/fix-entity/change-details.schema.js
```diff
@@ -0,0 +1,253 @@
...
+  // skips or reorders an entry cannot silently misalign the proposal (intent)
+  // against the observed outcome. When either side is absent the required-field
+  // rules already reject the record, so we only cross-check once both are arrays.
+  const changeResults = value.result?.changeResults;
+  const changes = value.target?.changes;
+  if (Array.isArray(changeResults) && Array.isArray(changes)) {
+    const intentKeys = new Set(changes.map(changeKey));
+    // (a) no orphan outcome — every changeResult maps to a proposed change.
+    if (changeResults.some((r) => !intentKeys.has(changeKey(r)))) {
+      return helpers.message('result.changeResults contains an entry with no matching target.changes (targetPath, property)');
```
MysticatBot, line 205: issue (blocking): The cross-validation that enforces "every target.change must have a matching result when applied is ALL" only fires when `Array.isArray(changeResults) && Array.isArray(changes)`. Since `changeResults` is optional on `resultSchema`, a record with `{ applied: "ALL", callStatus: "success" }` and **no** `changeResults` array passes validation.

This defeats the invariant: claiming ALL changes were applied without providing any evidence. Consumers reading `applied: ALL` would reasonably expect `changeResults` to enumerate what was applied.

**Fix:** Either make `changeResults` conditionally required when `applied` is `ALL` (via a Joi `.when()`), or extend the cross-validator to reject `applied: ALL` when `changeResults` is absent:

```js
if (value.result?.applied === APPLIED.ALL && !Array.isArray(changeResults)) {
  return helpers.message('result.applied is ALL but changeResults is missing');
}
```

packages/spacecat-shared-data-access/src/models/fix-entity/change-details.schema.js
```diff
@@ -0,0 +1,253 @@
...
+// Composite match key for (targetPath, property). JSON.stringify gives an
+// unambiguous, collision-safe delimiter regardless of what the strings contain.
+const changeKey = (change) => JSON.stringify([change.targetPath, change.property]);
+
+/**
+ * Canonical Joi schema for a v2 `changeDetails` record. Strict (unknown keys are
+ * rejected) so v2 writers cannot re-introduce freeform drift.
+ */
+export const changeDetailsV2Schema = Joi.object({
+  schemaVersion: Joi.number().valid(CHANGE_DETAILS_SCHEMA_VERSION).required(),
```
MysticatBot: issue (blocking): `Joi.number().valid(2)` with Joi's default `convert: true` silently coerces a string `"2"` to the integer `2` for validation purposes - so the record passes. However, the validator only checks for errors (`const { error } = ...validate(value, ...)`) and does not use the coerced return value. The original input (with `schemaVersion: "2"` as a string) is persisted as-is.

If mystique's Python runtime later reads this record and checks `schema_version == 2` (integer comparison), the string `"2"` will not match, causing the Python reader to misclassify the record as legacy freeform.

**Fix (pick one):**
1. Add `.strict()` to the `Joi.number().valid(CHANGE_DETAILS_SCHEMA_VERSION)` rule to reject string coercion (fail fast, force writers to provide an integer).
2. Or normalize explicitly in `validateChangeDetails` before validation:
```js
if (schemaVersion === '2') value = { ...value, schemaVersion: 2 };
```

Option 1 is simpler and enforces the ADR's integer discriminator contract at the boundary.

MysticatBot (changes requested):
Hey @alinarublea,

**Verdict: Request changes** - two semantic gaps in the validator that should be closed before this ships.
**Complexity: HIGH** - large diff (688 added lines, 7 files).
_Changes: Adds a canonical v2 Joi validation schema and attribute validator for `FixEntity.changeDetails`, implementing slice (a) of ADR mysticat-architecture#200 (7 files)._
Note: Recommend a human read before merge - this change modifies project documentation (`CLAUDE.md`) that anchors AI and developer guidance for the data-access package. The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] `applied: ALL` without `changeResults` bypasses completeness check - `change-details.schema.js:195` (details inline)
2. [Important] String `schemaVersion: '2'` passes validation but persists un-normalized, cross-runtime divergence risk - `change-details.schema.js:178` (details inline)

<details>

- nit: TypeScript declarations in `index.d.ts` are incomplete - missing `CHANGE_DETAILS` bundle and individual enum types (`SURFACES`, `ACTOR_TYPES`, etc.); TS consumers of `FixEntity.CHANGE_DETAILS.*` will have no type support - `index.d.ts:58`
- suggestion: `evidence` and `revert` fields on verify sub-schemas accept `Joi.any()` without a size cap, same storage-degradation risk that motivated the 4KB cap on `deployResponsePayload` - `change-details.schema.js:95`
- nit: `rejects` test helper invokes `validateChangeDetails` twice per test case; if a future custom validator mutates state, the second call may not reproduce the first failure - combine into a single assertion chain - `change-details.schema.test.js:148`
- suggestion: Add an explicit test for `applied: NONE` with a non-empty `changeResults` array to document that the permissive path is intentional (currently only `PARTIAL` is tested) - `change-details.schema.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 2s | <b>Cost:</b> $5.26 | <b>Commit:</b> <code>ac29f158fc8e7f8823090bc8f559a4cbf880da66</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - both prior blocking findings addressed correctly; no new issues.
**Complexity: HIGH** - large diff (821 lines, 7 files).
_Changes: Adds `.strict()` to the schemaVersion discriminator and a guard rejecting `applied: ALL` without `changeResults`, addressing both prior blocking findings (7 files)._

- `applied: ALL` without `changeResults` now correctly rejected with clear error message
- String `schemaVersion: '2'` now rejected via `.strict()` - closes the cross-runtime divergence risk with mystique Python

<details>

- suggestion: Add a test for `applied: ALL` with an empty `changeResults: []` array - the code correctly rejects it via the completeness cross-check (c), but an explicit test would document this edge case and prevent regressions if the guard logic is refactored - `change-details.schema.test.js`
- suggestion: CLAUDE.md states "Records without `schemaVersion: 2` are legacy freeform (v1)" - after the `.strict()` fix, string `'2'` triggers v2 validation and throws rather than falling through to the legacy guard. Consider clarifying the discriminator routing (only `undefined/null/1/'1'` are legacy; anything else triggers v2 validation) - `packages/spacecat-shared-data-access/CLAUDE.md:17`
- suggestion: CLAUDE.md Shape section could note the new `applied: ALL` requires `changeResults` constraint, since it is a blocking validation rule consumers will hit - `packages/spacecat-shared-data-access/CLAUDE.md:26`

</details>

Note: Recommend a human read before merge - this change modifies project documentation (`CLAUDE.md`) that anchors AI and developer guidance for the data-access package. The bot review is a complement to, not a replacement for, a human read here.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 55s | <b>Cost:</b> $3.19 | <b>Commit:</b> <code>5826f9b5f909fec59fb13124ab85af6c78fd167f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1830, opened 2026-07-20
feat(data-access): derive portable JSON Schema artifact for changeDetails v2 (SITES-47997)

alinarublea:
## What

**Stacked on #1828** (base is that branch — will retarget to `main` once #1828 merges). Derives a **portable JSON Schema artifact** from the Joi `changeDetailsV2Schema` so the **Python runtime (mystique) can validate `FixEntity.changeDetails` against the same contract** as the JS chokepoint — ADR [adobe/mysticat-architecture#200](https://github.com/adobe/mysticat-architecture/blob/main/platform/decisions/deploy-action-provenance-and-verify.md) blocking-#4 ("machine-readable, not prose").

The **Joi schema stays the single source of truth**; the JSON Schema is generated from it (chosen approach: *derive JSON Schema from Joi*).

- `scripts/generate-change-details-schema.js` — derives a **draft-2019-09** JSON Schema from `changeDetailsV2Schema` via `joi-to-json` (devDep). Run: `npm run generate:schemas`.
- `schemas/fix-entity-change-details.v2.schema.json` — the committed, **published** artifact mystique vendors (verified in the npm tarball; `scripts/` excluded).
- **Drift-guard test** — regenerates from the live Joi schema and `deep.equal`s the committed artifact; **fails CI if the Joi schema changed without regenerating**.
- `sha256` pattern made flag-free (`[a-fA-F0-9]`, no `/i`) so the derived `pattern` is portable to Python's `re`; `deployResponsePayload` carries a field-level description of the 4 KB cap.

## Verified

- Artifact compiles + validates correctly under a real draft-2019-09 validator (ajv): valid record passes, bad enum fails, string `schemaVersion: '2'` fails the `const 2` — **discriminator parity with the Joi `.strict()` behavior**.
- 100% coverage on the schema source; lint clean; `npm pack` ships the schema and excludes the generator.

## Caveat (documented in artifact `description`, generator JSDoc, and CLAUDE.md)

JSON Schema captures **structure/enums/required/patterns only**. The Joi `.custom()` cross-field invariants — `changeResults`↔`target.changes` key-match, `applied:ALL` completeness, `deployResponsePayload` 4 KB cap — are **not expressible** and must be ported by consumers that need them (mystique will port those in the companion PR).

## Next (Part B — mystique repo)

Vendor this artifact, validate in `apply.py` before persist, emit the v2 `target`/`result` shape (writes freeform v1 today), port the 3 custom invariants, add a CI drift check against the published package.

### DIFF
--- a/packages/spacecat-shared-data-access/scripts/generate-change-details-schema.js
+++ b/packages/spacecat-shared-data-access/scripts/generate-change-details-schema.js
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2025 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { writeFileSync, mkdirSync } from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+import parse from 'joi-to-json';
+
+import {
+  changeDetailsV2Schema,
+  CHANGE_DETAILS_SCHEMA_VERSION,
+} from '../src/models/fix-entity/change-details.schema.js';
+
+// draft-2019-09 is the newest draft joi-to-json emits and is fully supported by
+// Python's `jsonschema` (Draft201909Validator).
+export const JSON_SCHEMA_DRAFT = 'json-draft-2019-09';
+export const SCHEMA_ID = 'https://ns.adobe.com/spacecat/fix-entity/change-details/v2';
+// Package-root-relative path of the generated artifact.
+export const ARTIFACT_PATH = 'schemas/fix-entity-change-details.v2.schema.json';
+
+/**
+ * Build the portable JSON Schema artifact DERIVED from the canonical Joi schema
+ * (`changeDetailsV2Schema`). The Joi schema is the single source of truth; this
+ * artifact is what non-JS runtimes (mystique / Python) vendor and validate against.
+ *
+ * NOTE — JSON Schema captures structure + enums + required + patterns only. The
+ * cross-field invariants enforced by the Joi `.custom()` rules are NOT expressible
+ * here and must be ported by consumers that need them:
+ *   - result.changeResults <-> target.changes (targetPath, property) key-match
+ *   - applied:ALL completeness (every proposed change has a recorded outcome)
+ *   - deployResponsePayload 4 KB size cap
+ *
+ * @returns {object} the JSON Schema document.
+ */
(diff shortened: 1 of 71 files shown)


MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - well-scoped schema derivation with a solid drift-guard; one fragility in the generator's spread order worth fixing.
**Complexity: MEDIUM** - medium diff; new cross-runtime contract artifact with dependency signal.
_Changes: Derives a portable JSON Schema artifact from the Joi changeDetailsV2Schema so the Python runtime (mystique) can validate against the same contract, with a CI drift-guard test. (8 files)._
Note: Recommend a human read before merge - this change amends an architectural invariant documented in `packages/spacecat-shared-data-access/CLAUDE.md`. The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

1. [Important] Spread order in generator overwrites intentional metadata if joi-to-json emits conflicting keys - `scripts/generate-change-details-schema.js:46` (details inline)

<details>

- suggestion: Pin `joi-to-json` to exact version (`"5.0.5"` not `"^5.0.5"`) since the derived artifact is a committed cross-runtime contract and semver-range drift would cause spurious CI failures - `packages/spacecat-shared-data-access/package.json:58`
- suggestion: Add a test asserting `committed.required` does not include `'result'` (optional at create time) to guard against accidental promotion via tool upgrades - `test/unit/models/fix-entity/change-details.schema-artifact.test.js`
- nit: Root CLAUDE.md's "Per-Package File Structure" section does not document the new `scripts/` and `schemas/` directories; consider updating if this pattern will be reused - `CLAUDE.md`
- suggestion: Consider validating the committed artifact against the draft-2019-09 meta-schema in a test to ensure the artifact is valid JSON Schema, not just structurally equal to whatever joi-to-json emits

</details>

packages/spacecat-shared-data-access/scripts/generate-change-details-schema.js
```diff
@@ -0,0 +1,72 @@
...
+    description: 'Canonical v2 deploy-action record for FixEntity.changeDetails '
+      + '(SITES-47997, ADR adobe/mysticat-architecture#200). GENERATED from the Joi '
+      + 'source of truth (changeDetailsV2Schema in spacecat-shared-data-access) — DO NOT '
+      + 'EDIT BY HAND; run `npm run generate:schemas -w packages/spacecat-shared-data-access`. '
+      + 'Structure/enums/required/patterns only: the cross-field invariants '
+      + '(changeResults<->target.changes key-match, applied:ALL completeness, '
+      + 'deployResponsePayload 4KB cap) are Joi-enforced and NOT expressible in JSON Schema.',
+    ...body,
+  };
+}
```
MysticatBot, line 60: issue (blocking): The spread order here means `...body` is applied AFTER the explicit `$schema`, `$id`, `title`, and `description` fields. If `joi-to-json` ever emits any of those keys in its output, `body` will silently overwrite your intentional values.

Fix by spreading `body` first:
```js
return {
  ...body,
  $schema: 'https://json-schema.org/draft/2019-09/schema',
  $id: SCHEMA_ID,
  title: `FixEntity changeDetails v${CHANGE_DETAILS_SCHEMA_VERSION}`,
  description: '...',
};
```

This is a one-line reorder that prevents a future `joi-to-json` minor/patch bump from corrupting the published artifact's identity fields.

PR #1839, opened 2026-07-23
feat(data-access): list configuration versions via S3 object versioning

alinarublea:
## Why

The `spacecat-configuration` admin skill is REST-only, yet admins still needed **direct S3/AWS-console access for one thing: version discovery.** The global config is a singleton stored as S3 object versions, and the data layer only exposed `findLatest()` / `findByVersion(v)` — there was no way to *enumerate* versions. So to feed a historical VersionId to config diff/restore/export, an admin had to open the prod S3 bucket.

This adds the enumeration primitive that closes that gap (the `GET /configurations/versions` endpoint in spacecat-api-service consumes it).

## What

- **`ConfigurationCollection.listVersions({ limit, keyMarker, versionIdMarker, detail })`** — S3 `ListObjectVersions`, newest first, with pagination markers. When `detail` (default), each row is enriched with `updatedBy`/`updatedAt` via a **metadata-only `HeadObject`** (no multi-MB body download).
- **`create()`** now stamps `updatedBy`/`updatedAt` into **S3 user-metadata** at write time so enrichment stays cheap. Versions written before this return `null` (graceful).
- `limit` is coerced + clamped to `[1, 1000]`, which also bounds the enrichment fan-out.
- Systemic head failures (AccessDenied/throttling) log at **error** so an all-`null` page reads as an outage, not "old versions"; benign `NoSuch*` degrade quietly.
- `.d.ts` types + unit tests. `configuration.collection.js` at **100% coverage**; full data-access suite green.

## Notes for reviewers

- Callers must page on `isTruncated` + returned markers, **not** `versions.length` (documented on the method) — `MaxKeys` bounds the raw S3 result before the exact-key filter.
- This is the first of a 3-PR chain; **merge/release this first**, then api-service bumps the dep.

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/configuration/configuration.collection.js
+++ b/packages/spacecat-shared-data-access/src/models/configuration/configuration.collection.js
@@ -12,6 +12,8 @@
 
 import {
   GetObjectCommand,
+  HeadObjectCommand,
+  ListObjectVersionsCommand,
   PutObjectCommand,
 } from '@aws-sdk/client-s3';
 
@@ -108,6 +110,13 @@ class ConfigurationCollection {
         Key: S3_CONFIG_KEY,
         Body: JSON.stringify(configData),
         ContentType: 'application/json',
+        // Stamp the audit fields into S3 user-metadata so `listVersions` can
+        // surface who/when for each version via a cheap metadata-only HeadObject
+        // (no full-body download). Keys are lowercased by S3.
+        Metadata: {
+          updatedby: String(configData.updatedBy),
+          updatedat: String(configData.updatedAt),
+        },
       });
 
       const response = await this.s3Client.send(command);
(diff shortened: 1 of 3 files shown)


MysticatBot (changes requested):
Hey @alinarublea,

**Verdict: Request changes** - two material issues to address before merge.
**Complexity: MEDIUM** - medium diff, new API surface on shared library.
_Changes: Adds `listVersions()` to ConfigurationCollection for S3 object version enumeration with optional HeadObject metadata enrichment (3 files)._

## Must fix before merge

1. [Important] Unbounded parallel HeadObject fan-out at high page sizes - `configuration.collection.js:328` (details inline)
2. [Important] Metadata can permanently stamp literal "undefined" into immutable S3 user-metadata - `configuration.collection.js:117` (details inline)

<details>

- nit: `metadata.updatedby || null` coerces empty string to null - consider `?? null` (nullish coalescing) to distinguish "metadata key present but empty" from "metadata key absent" - `configuration.collection.js:244`
- suggestion: Add a test that calls `create(mockRecord)` without an explicit `updatedBy` and asserts the default `'system'` value flows through to `Metadata.updatedby`
- suggestion: TypeScript could use overloaded signatures or a discriminated union to distinguish the `detail=true` shape (with `updatedBy`/`updatedAt`) from the `detail=false` shape (without them)
</details>

packages/spacecat-shared-data-access/src/models/configuration/configuration.collection.js
```diff
@@ -205,6 +214,135 @@ class ConfigurationCollection {
...
+          versionId: version.VersionId,
+          lastModified: version.LastModified instanceof Date
+            ? version.LastModified.toISOString()
+            : version.LastModified,
+          isLatest: Boolean(version.IsLatest),
+          size: version.Size,
+        }));
+
+      const versions = detail
+        ? await Promise.all(rawVersions.map((version) => this.#enrichVersion(version)))
```
MysticatBot: issue (blocking): **Unbounded parallel HeadObject fan-out.** When `detail=true` (the default) and `limit` is set high (up to 1000), `Promise.all(rawVersions.map(...))` fires up to 1000 concurrent HeadObject requests in a single event-loop tick. In a Lambda environment this can exhaust socket pools, trip S3 throttling (503 SlowDown), or hit the per-invocation network fd limit. The `#enrichVersion` catch block degrades silently per-row, so a throttled burst would return a page full of nulls with only error-level logs as evidence.

The JSDoc acknowledges this ("bounds the enrichment fan-out") but the code does not actually bound it - `maxKeys` limits what S3 returns, not the parallelism of downstream HeadObject calls.

**How to fix:** Either cap the enrichment concurrency independently (e.g. batch in chunks of 10-25), or clamp `maxKeys` to a lower ceiling when `detail=true` (e.g. `Math.min(maxKeys, 50)` for the enrichment path). At the default `limit=25` this is benign, but the 1000 ceiling with `detail=true` as default is a latent production issue.

packages/spacecat-shared-data-access/src/models/configuration/configuration.collection.js
```diff
@@ -108,6 +110,13 @@ class ConfigurationCollection {
         Key: S3_CONFIG_KEY,
         Body: JSON.stringify(configData),
         ContentType: 'application/json',
+        // Stamp the audit fields into S3 user-metadata so `listVersions` can
+        // surface who/when for each version via a cheap metadata-only HeadObject
+        // (no full-body download). Keys are lowercased by S3.
+        Metadata: {
+          updatedby: String(configData.updatedBy),
```
MysticatBot: issue (blocking): **Metadata can permanently stamp literal "undefined" or "null".** `String(configData.updatedBy)` produces the literal string `"undefined"` when `updatedBy` is `undefined`, and `"null"` when it is `null`. S3 user-metadata is immutable per-version, so a bad stamp is permanent for that version and will surface in the admin UI.

The existing `create()` logic likely defaults `updatedBy` to `'system'` upstream, but this stamp has no independent guard - if the upstream default ever changes or a code path bypasses it, the metadata is silently corrupted.

**How to fix:** Guard the stamp:
```js
Metadata: {
  updatedby: configData.updatedBy ? String(configData.updatedBy) : '',
  updatedat: configData.updatedAt ? String(configData.updatedAt) : '',
},
```

MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - both prior blocking findings are properly addressed; no new issues.
**Complexity: MEDIUM** - medium diff, new API surface on shared library.
_Changes: Adds `listVersions()` to ConfigurationCollection with bounded-concurrency HeadObject enrichment and fixes the metadata stamp (3 files)._

<details>

- nit: `metadata.updatedby || null` coerces empty string to null - consider `?? null` for semantic precision (academic for data written by this code; only pre-stamp legacy versions could surface empty string) - `configuration.collection.js:244`
- nit: TypeScript types `size` as `number` but S3 SDK types `Size` as `number | undefined` for delete markers - in practice the key filter excludes those, so cosmetic only - `index.d.ts:67`
</details>

- Concurrency bound added via `ENRICH_CONCURRENCY = 25` with serial batch processing in `#enrichVersions`
- Metadata stamp now uses pre-validated `configData.updatedBy`/`updatedAt` directly (guaranteed non-empty strings), eliminating the `String()` coercion risk

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 30s | <b>Cost:</b> $4.56 | <b>Commit:</b> <code>4afe5c8540b5d440dac9e5f00bc731a5752a259c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1849, opened 2026-07-29
feat(content-client): dual Google Drive service-account with fallback

tr1eder:
## 1. Abstract
Adds optional dual Google Drive service-account support to `@adobe/spacecat-shared-content-client`. When a fallback service account is configured, `ContentClient` retries a Google Drive operation against it if the primary service account cannot access the folder.

## 2. Reasoning
Part of the EDS Google Drive service-account migration (SITES-47990): the old Helix SA (`experience-success-studio@helix-225321`) is being replaced by the new Adobe SA (`aem-sites-optimizer@adbe-gcp0843`). `ContentClient` is the single seam every SpaceCat service uses to read and apply EDS Google Drive content — it builds one Google service account from `GDRIVE_*` env vars and hands it to `@adobe/spacecat-helix-content-sdk`. During the migration a customer's Drive folder may be shared with only one of the two service accounts, so a single-account client fails for any customer whose folder is not yet re-shared with that account. This closes the V1 gap (SITES-48193): the same dual-SA behaviour is already live in `spacecat-auth-service` (onboarding validate) and `mystique` (V2), but the V1 autofix apply path — which goes through `ContentClient` — had no fallback.

- `ContentClient.createFrom` now also builds a fallback Google Drive service-account config from `GDRIVE_*_FALLBACK` env vars. The fallback reuses the Google-universal constant fields (type, auth_uri, token_uri, universe_domain, auth-provider cert URL) from the primary and only overrides the per-account fields (email, private key, private key id, client id, project id, client cert URL). The fallback private key is newline-unescaped the same way the primary is.
- Google Drive operations now run through a fallback-aware wrapper. On a folder-sharing / permission failure (403 / 404 / not-found / access-denied), it rebuilds the SDK client with the fallback account, retries the operation once, and then keeps using the fallback account for the remainder of that `ContentClient` instance's lifetime (so subsequent calls do not re-probe the primary).
- Behaviour is unchanged when no fallback is configured (single account, as today). The fallback is Google-Drive-only — OneDrive is untouched — and the Helix-admin-token methods (`getEditURL` / `getResourcePath` / `getLivePreviewURLs`) do not use the Google Drive account and are unaffected.
- Operationally visible: when the fallback is used, a warning is logged identifying the primary account that failed and the fallback account being tried.

## 4. Required information
- Jira / issue: SITES-48193 (sub-task of SITES-47990) — https://jira.corp.adobe.com/browse/SITES-48193

- `spacecat-autofix-worker` — contract / consumer. Its V1 Google Drive autofix apply (alt-text, broken-backlinks, broken-internal-links, meta-tags) goes through `ContentClient` and is currently broken for folders shared only with the new SA. To adopt this fix it must bump `@adobe/spacecat-shared-content-client` to the release containing this change, set the `GDRIVE_*_FALLBACK` secrets in Vault, and redeploy.
- `spacecat-api-service` — consumed, no behaviour change. It also uses `ContentClient`, but only via the Helix-admin-token methods (edit / preview URL resolution), which do not use the Google Drive service account, so it is unaffected by this migration.

- Vault `dx_mysticat/dev/autofix-worker` was patched (2026-07-29) with the new-SA `GDRIVE_*_FALLBACK` fields ahead of consumer testing; the primary remains the old SA, so runtime behaviour is unchanged until the consumer adopts this release. Stage and prod are not yet patched.
- The same dual-SA pattern is already live on prod in `spacecat-auth-service`: recordedfuture's onboarding validate exercised the fallback (primary/old SA "file not found" → fallback/new SA validated), which is the behaviour this change brings to the `ContentClient` apply path.
- No end-to-end Google Drive apply has been exercised against this code yet — that requires the `spacecat-autofix-worker` dependency bump and a deploy.

## 7. Test plan
- Local: no end-to-end run was possible without a real Google Drive site; the fallback build, the retry-on-sharing-error path, the sticky-fallback-across-calls behaviour, the no-retry-on-non-sharing-error path, both-accounts-fail propagation, and OneDrive-skips-fallback are covered by unit tests in this package (run by CI).
- dev / stage / prod: after this is released, bump `@adobe/spacecat-shared-content-client` in `spacecat-autofix-worker`; set `GDRIVE_*_FALLBACK` in Vault `dx_mysticat/{dev,stage,prod}/autofix-worker` (dev already done) and redeploy; then run alt-text autofix on an internal Friends-and-Family Google Drive site whose folder is shared with the new SA only, and confirm the fallback fires in the autofix-worker CloudWatch logs (warning: "retrying with fallback SA").

- Related — `adobe/spacecat-auth-service#613`: merged and live on prod; the reference implementation of this dual-SA pattern for the onboarding validate path.
- Related — the ASO UI "Connect to AEM Sites" dialog change (SITES-48438) and the public-docs change (SITES-48014, docs PR #72): flip the customer-facing SA email to the new account.
- Depends-on / blocks — this change must merge and release, then `spacecat-autofix-worker` bumps the dependency; it must all land before the migration cleanup (SITES-48016) removes the old SA, or V1 Google Drive autofix breaks.
- Ordered sequence: (1) merge + release this spacecat-shared change; (2) bump `@adobe/spacecat-shared-content-client` in `spacecat-autofix-worker`; (3) set Vault `GDRIVE_*_FALLBACK` for stage + prod autofix-worker; (4) redeploy autofix-worker; (5) later, SITES-48016 removes the old SA.

### DIFF
--- a/packages/spacecat-shared-content-client/src/clients/content-client.js
+++ b/packages/spacecat-shared-content-client/src/clients/content-client.js
@@ -211,24 +211,30 @@ const removeRedirectLoops = (currentRedirects, newRedirects, log) => {
   return noCycleRedirects;
 };
 
-// Google Drive returns 404 "File not found" when the folder is NOT SHARED with the service
-// account (not only when a document is genuinely missing), and 403 / "does not have permission"
-// when shared read-only. So both 403 and 404 signal a possible sharing problem and must trigger
-// the fallback (verified on prod: the old SA got "File not found" for a folder shared only with
-// the new SA). Accepted side effect: a genuinely missing document also triggers one extra
-// fallback attempt before the error propagates. Keep in sync with the spacecat-auth-service
-// `categorizeError` predicate (src/google-drive/handler.js).
-const GDRIVE_SHARING_ERROR = /\b(403|404)\b|forbidden|not found|access denied|insufficient|does not have permission|not shared/i;
+// A "folder not shared with this service account" failure surfaces as an HTTP 403 (shared
+// read-only / "does not have permission" / insufficientFilePermissions) or 404 ("File not found" —
+// Drive returns 404 for a folder the SA cannot see; verified on prod: the old SA got "File not
+// found" for a folder shared only with the new SA). Prefer the SDK error's STRUCTURED status code;
+// only fall back to a few specific reason phrases. Do NOT substring-match the whole message for
+// bare "not found" / "insufficient" / loose "403"/"404" — those false-match unrelated errors (e.g.
+// "module not found", "insufficient memory", a 404 in a trace id). Keep in sync with the
+// spacecat-auth-service `categorizeError` predicate (src/google-drive/handler.js).
+const GDRIVE_SHARING_REASON = /forbidden|insufficientFilePermissions|does not have permission|not shared|file not found/i;
 
 /**
  * True when a Google Drive error looks like a folder-sharing / permission problem (the service
- * account cannot see or edit the folder) — the signal to retry with the fallback SA.
+ * account cannot see or edit the folder) — the signal to retry with the fallback SA. Prefers the
+ * error's structured HTTP status (403/404) over message matching.
  * @param {Error & {code?: any, status?: any}} error
  * @returns {boolean}
  */
-const isGDriveSharingError = (error) => GDRIVE_SHARING_ERROR.test(
-  `${error?.message || ''} ${error?.code || ''} ${error?.status || ''}`,
-);
+const isGDriveSharingError = (error) => {
+  const status = Number(error?.status ?? error?.code);
+  if (status === 403 || status === 404) {
+    return true;
+  }
+  return GDRIVE_SHARING_REASON.test(error?.message || '');
+};
 
 /**
  * Build the fallback Google Drive service-account config, or null when no fallback is configured.
@@ -368,11 +374,9 @@ export default class ContentClient {
    * primary error also pins to the fallback, which is acceptable for the not-shared case this
    * targets). A no-op passthrough for OneDrive or when no fallback is configured.
    *
-   * Retry safety: a sharing/permission failure surfaces on the FIRST Drive access (resolving the
-   * document / redirects file), i.e. before any write in a compound read-then-write op, so
-   * re-running the whole `op` on the fallback does not double-apply a write. A future write op that
-   * can throw a sharing-matching error AFTER a partial mutation would break that assumption — wrap
-   * only its read then.
+   * Callers pass ONLY the initial Drive read here (via `#getDocument` / `#getRedirectsFile`) — the
+   * point at which a folder-not-shared error surfaces. Writes run afterwards, outside this wrapper,
+   * on the winning SA's handle, so a retry never replays a (non-idempotent) write.
    * @template T
    * @param {() => Promise<T>} op
    * @returns {Promise<T>}
@@ -397,6 +401,14 @@ export default class ContentClient {
       await this.#initClient();
       return op();
     }
+  }
+
+  async #getDocument(docPath) {
+    return this.#withGDriveFallback(() => this.rawClient.getDocument(docPath));
+  }
+
+  async #getRedirectsFile() {
+    return this.#withGDriveFallback(() => this.rawClient.getRedirects());
   }
 
   #logDuration(message, startTime) {
@@ -472,10 +484,8 @@ export default class ContentClient {
     this.log.debug(`Getting page metadata for ${this.site.getId()} and path ${path}`);
 
     const docPath = this.#resolveDocPath(path);
-    const metadata = await this.#withGDriveFallback(async () => {
-      const document = await this.rawClient.getDocument(docPath);
-      return document.getMetadata();
-    });
+    const document = await this.#getDocument(docPath);
+    const metadata = await document.getMetadata();
 
     this.#logDuration('getPageMetadata', startTime);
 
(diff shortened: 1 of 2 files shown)


MysticatBot (changes requested):
Hey @tr1eder,

**Verdict: Request changes** - write operations wrapped in the fallback closure can double-apply on partial failure, and the error-matching regex is broad enough to trigger spuriously.
_Changes: Adds optional dual Google Drive service-account fallback to ContentClient, retrying on permission errors with a secondary SA (2 files)._

## Must fix before merge
1. [Important] Non-idempotent write replay on partial failure inside `#withGDriveFallback` - `content-client.js:~515` (details inline)
2. [Important] Overly broad `GDRIVE_SHARING_ERROR` regex matches legitimate non-sharing errors - `content-client.js:213` (details inline)
3. [Important] No fallback test coverage for write-path methods - `content-client.test.js` (details inline)

<details>

- nit: Test sandbox convention - new test block uses `sinon.spy()` instead of `sandbox.spy()` per the CLAUDE.md rule "Each test file uses a sandbox" - `content-client.test.js:~1265`
- suggestion: Use private class fields (`#fallbackConfig`, `#usingFallback`) to match the existing `#initClient`/`#withGDriveFallback` pattern and prevent external mutation - `content-client.js:340-341`
- suggestion: Validate fallback config at construction time (`validateConfiguration(fallbackConfig, contentSourceType)`) to surface misconfig early rather than at runtime during fallback - `content-client.js:~295`
</details>

packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -416,12 +514,13 @@ export default class ContentClient {
 
   async getRedirects() {
```
MysticatBot, line 522: issue (blocking): **Non-idempotent write replay on partial failure.**

`updateRedirects` wraps both the read (`redirectsFile.get()`) and the write (`redirectsFile.append(...)`) in a single `#withGDriveFallback` closure. If `append()` throws an error whose message matches `GDRIVE_SHARING_ERROR` (e.g. a 403 from a rate-limit or quota error, a transient permission-propagation delay, or the broad "not found" substring appearing in a network error), the fallback path re-runs the entire closure from scratch - including reading current redirects again and appending. Since `append` is not idempotent, this can produce duplicate redirects.

The same structural risk applies to `updatePageMetadata` (read-merge-write replayed), though metadata writes are idempotent-by-value so the practical impact is lower.

**Why it matters:** The JSDoc on `#withGDriveFallback` acknowledges: "a sharing/permission failure surfaces on the FIRST Drive access ... before any write in a compound read-then-write op, so re-running the whole op on the fallback does not double-apply a write." But `updateRedirects` violates that invariant today since `append()` could throw a regex-matching error after a successful read.

**How to fix:** Either (a) split the closure so only the initial Drive read is wrapped in the fallback boundary and the write happens outside it, or (b) track whether a mutating call has been attempted and skip the fallback retry if so. Option (a) is simpler and avoids reasoning about which errors can fire after a write.

packages/spacecat-shared-content-client/src/clients/content-client.js
```diff
@@ -211,6 +211,57 @@ const removeRedirectLoops = (currentRedirects, newRedirects, log) => {
   return noCycleRedirects;
 };
```
MysticatBot, line 213: issue (blocking): **Overly broad regex matches legitimate non-sharing errors.**

The regex `/\b(403|404)\b|not found|access denied|insufficient|permission|not shared/i` is tested against the full concatenation of `error.message`, `error.code`, and `error.status`. Several alternations are dangerously generic:

- `"not found"` matches any error containing those words (e.g. "Spreadsheet tab redirects not found", "Configuration property not found", "Module not found").
- `"insufficient"` matches "insufficient memory", "insufficient quota", etc.
- `\b403\b` / `\b404\b` match those numbers anywhere in a message string (trace IDs, timing values).

This compounds with finding 1: a false-positive regex match on a write-phase error triggers the non-idempotent replay.

**How to fix:** Prefer a structured check first (`error.code === 403 || error.code === 404` or `error.status === 403`) and only fall through to free-text matching with more anchored phrases (e.g. "does not have permission", "file not found", "folder not shared") when `.code` is not set. At minimum, add negative test cases for plausible false-positive messages to document the accepted surface.

packages/spacecat-shared-content-client/test/clients/content-client.test.js
```diff
@@ -1235,4 +1235,187 @@ describe('ContentClient', () => {
...
+      );
+      expect(client.fallbackConfig).to.deep.equal({
+        auth_provider_x509_cert_url: 'https://auth-provider.uri', // constant, reused from primary
+        auth_uri: 'https://auth.uri', // constant, reused
+        client_email: 'fallback-email', // overridden
+        client_id: 'fallback-client-id', // overridden
+        client_x509_cert_url: 'fallback-cert-url', // overridden
+        private_key: 'fallback\nprivate\nkey', // overridden AND newline-unescaped
+        private_key_id: 'fallback-key-id', // overridden
+        project_id: 'fallback-project-id', // overridden
```
MysticatBot, line 1283: issue (blocking): **No fallback test coverage for write-path methods.**

All fallback-retry tests exercise `getPageMetadata` (a pure read). There are no tests for the fallback behavior of write-path methods (`updateRedirects`, `updatePageMetadata`, `updateBrokenInternalLink`, `updateImageAltText`). Given the retry-safety concern on `updateRedirects` (the `append` call is non-idempotent), a test that verifies the fallback fires correctly on the write path - and ideally one that documents the invariant that sharing errors only surface on the initial Drive read - would lock in the correct behavior.

**How to fix:** Add at least one test for `updateRedirects` that triggers the fallback (error on initial `getRedirects()` call) and verifies `append` is called on the fallback client redirects file, not the primary. Optionally, add a test documenting that a write-phase error (from `append`) does NOT trigger the fallback.

MysticatBot (approved):
Hey @tr1eder,

**Verdict: Approve** - prior findings addressed cleanly; restructuring the fallback to wrap only reads eliminates the write-replay risk, and the narrowed predicate is well-documented.
_Changes: Restructures the dual Google Drive SA fallback to wrap only reads, narrows the error predicate to prefer structured status codes, and adds write-path and edge-case tests (2 files)._

<details>

- nit: `return op()` without `await` in the catch branch of `#withGDriveFallback` loses the async stack frame on fallback failure; functionally equivalent but slightly harder to debug - `content-client.js:~395`
- nit: Every 404 (even a genuinely missing document) triggers the fallback retry, adding one round-trip for real not-found cases during migration; acceptable given the TODO removal timeline - `content-client.js:230`
- suggestion: Export or directly unit-test `isGDriveSharingError` with boundary inputs (string coercion of `error.status`, `error.code` fallback path, null/undefined) to make regressions localizable without full integration setup - `content-client.js:228`
- suggestion: Add a `@param` JSDoc entry for `fallbackConfig` on the constructor, or validate it when non-null, so direct instantiators know the contract - `content-client.js:345`
</details>

- Non-idempotent write replay: reads now wrapped via `#getDocument`/`#getRedirectsFile`; writes run outside the fallback boundary
- Overly broad sharing-error regex: replaced with `isGDriveSharingError()` using structured HTTP status first, narrower reason phrases second
- Missing write-path fallback tests: added updateImageAltText, updateRedirects through-fallback, and write-phase-failure-no-retry tests


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 47s | <b>Cost:</b> $6.53 | <b>Commit:</b> <code>260447f148dc11a8dd4f437dc127bcf02931558f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1851, opened 2026-07-29
fix: validate cdnlogsFilter key against a column allowlist in site config

akshaymagapu:
## Problem
Site config accepts an `llmo.cdnlogsFilter` array whose entries carry a `key` naming a CDN log column. The schema previously allowed any string for `key`, so a value that isn't a real column could be persisted and later used to build Athena filter clauses downstream (audit worker).

## Change
- Constrain `cdnlogsFilter[].key` in the config Joi schema to the known CDN log columns (`url`, `host`, `x_forwarded_host`, `user_agent`, `referer`, `cdn_provider`).
- Normalize case (Athena folds identifiers to lowercase), so `X_forwarded_host` is accepted and stored as `x_forwarded_host`.
- Reject invalid keys at write time in `updateLlmoCdnlogsFilter()` so a bad key surfaces as a validation error instead of being persisted.

This keeps persisted filters aligned with the exact set of columns the audit worker will accept (companion change: adobe/spacecat-audit-worker#2826).

## Tests
- Added coverage for all allowed keys, case normalization, and invalid-key rejection.
- data-access config tests pass; changed files lint clean.

Ref: LLMO-6621

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/config.js
+++ b/packages/spacecat-shared-data-access/src/models/site/config.js
@@ -70,6 +70,30 @@ const RESERVED_SCRAPER_HEADER_NAMES = new Set([
 
 const LLMO_TAG_PATTERN = /^(market|product|topic):\s?.+/;
 const AWS_REGION_PATTERN = /^[a-z]{2}(?:-[a-z]+)+-\d+$/i;
+
+// Columns on the CDN log tables that a cdnlogsFilter entry is allowed to target.
+// `key` is interpolated (unquoted) into Athena SQL by the audit worker's
+// buildSiteFilters(), so it MUST be constrained to this allowlist to prevent SQL
+// injection (VULN-37491). Keep in sync with ALLOWED_FILTER_KEYS in
+// spacecat-audit-worker/src/utils/cdn-utils.js.
+export const CDN_LOGS_FILTER_KEYS = [
+  'url',
+  'host',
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - clean security fix with proper defense-in-depth.
**Complexity: LOW** - small, focused change in a single package.
_Changes: Constrains cdnlogsFilter key to a column allowlist with case normalization and write-time validation to prevent SQL injection (2 files)._

<details>

- nit: The thrown `Error` in `updateLlmoCdnlogsFilter` is generic - callers cannot distinguish a validation failure from an unexpected runtime error without parsing the message string. A typed error or error code would help API layers return 400 vs 500. - `config.js:870`
- suggestion: Add boundary-condition tests for `updateLlmoCdnlogsFilter(null)`, `updateLlmoCdnlogsFilter([])`, and a filter with an empty `value` array to document the intended behavior at edges. - `config.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 28s | <b>Cost:</b> $2.80 | <b>Commit:</b> <code>eaaa241b10008c60b093fe7f7fa6070a23207544</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1850, opened 2026-07-29
feat(site-config): add showWww flag to llmo config

byteclimber:
## Summary
- Adds `showWww` boolean to the `llmo` site config schema (`Joi.boolean().optional()`), plus an `updateLlmoShowWww` setter following the existing single-field llmo setter pattern.
- Lets LLM Optimizer's Brands Management UI (companion PR in `project-elmo-ui`) display a site's domain with its `www.` prefix when the site opts in, instead of the current always-stripped display.

## Context
Two customers (kent.edu, Lumen) were onboarded with `base_url` stripped of `www.` (SpaceCat strips it unconditionally at onboarding). Backend citation-matching already treats www/non-www as equivalent, so this is purely a display-layer opt-in — no other backend behavior changes.

`PATCH /sites/{id}` already deep-merges `config.llmo` (spacecat-api-service), so once this ships, `{ "config": { "llmo": { "showWww": true } } }` persists with no controller changes needed there.

## Test plan
- [x] Unit tests added: schema accepts/rejects `showWww`, `updateLlmoShowWww` setter behavior (mirrors `updateLlmoBrand`/`updateLlmoDataFolder` coverage)
- [x] `npm test -w packages/spacecat-shared-data-access` — 2754 passing
- [x] `npm run lint -w packages/spacecat-shared-data-access` — clean

PR #1852, opened 2026-07-29
fix(site): guard config writes against the config schema

byteclimber:
## Summary
`Site.setConfig()` previously accepted any value with no schema validation at all — only entity `create()` runs attribute validation (`base.collection.js` `#validateItem`/`#prepareItem`); the generic auto-generated update setter just patches the raw value. The only thing touching Joi on the update path was the lenient `Config()` getter, which logs a warning and silently falls back to the raw data on a validation failure.

## Context
Found while manually verifying the `showWww` display fix (#1850): `PATCH /sites/{id}` with an invalid `showWww` value (e.g. a non-boolean) returned `200` and silently stored the bad value instead of being rejected, even though the Joi schema itself correctly flags the error (confirmed via the exact log line: `"llmo.showWww" must be a boolean"`).

This isn't specific to `showWww` or even to `config` — it's a general create-vs-update validation asymmetry in this data-access layer. Scoping this fix to `Site.config` only (not `BaseModel.save()` platform-wide) per discussion, since a platform-wide fix has much higher blast radius across every entity/attribute.

## Approach
- Overrides `setConfig` on `Site` to validate via `validateConfiguration` before patching, following the exact same override + rollout pattern already established for `Suggestion`/`FixEntity.setStatus` (`STATUS_TRANSITION_ENFORCEMENT`).
- New `CONFIG_VALIDATION_ENFORCEMENT` env var (`off` | `warn` (default) | `enforce`), read via a new `config-validation-guard.js` util mirroring `status-transition-guard.js`.
- Ships in `warn` mode by default — **no behavior change today**. At least one seeded local-dev site already carries config that fails the current schema (an unrelated pre-existing `imports` entry type mismatch), so flipping straight to `enforce` would reject unrelated PATCH requests on that site. Flip to `enforce` once legacy data is confirmed clean/rare.
- Reads are intentionally untouched — the lenient `Config()` getter still tolerates legacy invalid config on existing sites so a bad historical record doesn't break every read of that site.

## Test plan
- [x] New unit tests: `config-validation-guard.test.js` (mirrors `status-transition-guard.test.js` coverage), `Site.setConfig` tests (valid config, warn-mode fallback, enforce-mode throw, off-mode no-op)
- [x] `npm test -w packages/spacecat-shared-data-access` — 2772 passing, no regressions
- [x] Manually verified against a local dev server: invalid `showWww` value is now rejected with `enforce` mode set, and the exact `"llmo.showWww" must be a boolean"` error surfaces

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/site.model.js
+++ b/packages/spacecat-shared-data-access/src/models/site/site.model.js
@@ -18,6 +18,8 @@ import {
   AUTHORING_TYPES,
 } from '@adobe/spacecat-shared-utils';
 import BaseModel from '../base/base.model.js';
+import { validateConfiguration } from './config.js';
+import { guardConfigValidation } from '../../util/config-validation-guard.js';
 
 const HLX_HOST = /\.(?:aem|hlx)\.(?:page|live)$/i;
 export const AEM_CS_HOST = /^author-p(\d+)-e(\d+)/i;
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @byteclimber,

**Verdict: Approve** - Clean implementation following an established pattern; ships safely in warn mode.
_Changes: Adds schema validation guard to Site.setConfig() with env-controlled enforcement modes (off/warn/enforce), mirroring the existing status-transition-guard pattern (4 files)._

<details>

- nit: The guard discards `validateConfiguration`'s sanitized return value and writes the raw input via `patchValue`. This is likely intentional (reads normalize via the `Config()` getter), but a one-line comment in `setConfig` noting the asymmetry would save future readers from investigating whether it's a bug. - `src/util/config-validation-guard.js:66`
- suggestion: The thrown `ValidationError` in enforce mode includes the full Joi error message, diverging from `status-transition-guard.js` which sanitizes its thrown message. If exposing schema detail to API callers is intentional (so they can self-correct), consider documenting the divergence; otherwise sanitize to a generic message and keep detail only in the `log.warn` call. - `src/util/config-validation-guard.js:72`
- nit: The Environment Variables table in `packages/spacecat-shared-data-access/CLAUDE.md` does not list `CONFIG_VALIDATION_ENFORCEMENT`. Adding a row keeps the table complete for developers referencing it. - `packages/spacecat-shared-data-access/CLAUDE.md`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 10m 55s | <b>Cost:</b> $7.66 | <b>Commit:</b> <code>84bf8413a17b6a69d195f396cfe065e969456d20</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1864, opened 2026-08-05
fix(akamai-client): configurable rule-tree timeout + patchRuleTree + accurate timeout message

ABHA61:
## What
`AkamaiClient` gains the pieces the LLMO Akamai onboarding needs for large properties, plus a latent message-bug fix in `spacecat-shared-utils`:

- **Configurable `ruleTreeTimeoutMs`** (default `60000`), applied only to the three calls whose latency scales with rule-tree size — `getRuleTree`, `updateRuleTree`, `patchRuleTree`. Cheap metadata calls (`getLatestVersion`, `createVersion`, `activate`, …) keep tracingFetch's 10s default.
- **New `patchRuleTree`** — applies a JSON Patch (RFC 6902) to a version's stored tree with `validateRules=true`. `If-Match`, when supplied, is sent as an RFC 7232 quoted etag (PAPI rejects a bare token with a generic 400).
- **`getRuleTree` `options.validateRules`** — opt-in full validation pass that returns `errors`/`warnings` (used to tell whether a version is activatable).
- **`index.d.ts`** updated for all of the above.
- **tracing-fetch message fix** — the timeout error always reported `"10000ms"` regardless of the configured timeout (it read a non-existent `_ms` off `@adobe/fetch`'s signal). The abort always fired at the right time; only the message was wrong. The real value is now threaded through.

## Why
On a large Akamai property, PAPI's `validateRules` pass over the whole tree genuinely takes longer than tracingFetch's 10s default, so the rule-tree write was aborted client-side before Akamai could respond — the version got created (a cheap call) but its rule content never landed.

## Testing
- `akamai-client.test.js`: the three rule-tree calls use the configured `ruleTreeTimeoutMs` (not the 10s default); cheap calls unaffected; `patchRuleTree` content-type / quoted If-Match / validateRules; `getRuleTree` validateRules returns errors/warnings.
- `tracing-fetch.test.js`: asserts the timeout message reports the configured value.
- `npm test` + `npm run lint` clean in `spacecat-shared-akamai-client` and `spacecat-shared-utils`.

## Merge order
**Merge this first.** adobe/spacecat-api-service#3013 consumes the published release of this package (its PATCH deploy uses `patchRuleTree`; its deploy-status uses `getRuleTree({ validateRules })`). Once this releases, adobe/spacecat-api-service#3013 swaps its temporary gist-tarball dependency for the published version.

### DIFF
--- a/packages/spacecat-shared-akamai-client/src/akamai-client.js
+++ b/packages/spacecat-shared-akamai-client/src/akamai-client.js
@@ -557,7 +557,11 @@ export default class AkamaiClient {
     }
     const headers = {
       'Content-Type': 'application/json-patch+json',
-      ...(etag ? { 'If-Match': etag } : {}),
+      // RFC 7232 If-Match values must be a quoted string; PAPI enforces this strictly and rejects
+      // an unquoted etag with a generic, unhelpful 400 ("The system cannot understand your
+      // request") regardless of what the patch ops say — reproduced empirically: every op
+      // combination failed identically with a bare `If-Match: <etag>`, and succeeded once quoted.
+      ...(etag ? { 'If-Match': `"${etag}"` } : {}),
     };
     this.log.info(
       `Patching rule tree for property ${propertyId} v${version} `
--- a/packages/spacecat-shared-akamai-client/src/akamai-client.js
+++ b/packages/spacecat-shared-akamai-client/src/akamai-client.js
@@ -28,13 +28,8 @@ const RULE_FORMAT_RE = /^[a-z0-9-]+$/i;
 // printable-ASCII, whitespace-free token (PAPI returns a hex etag) — this stops a malformed value
 // from injecting extra headers via CR/LF, mirroring RULE_FORMAT_RE's guard for ruleFormat.
 const ETAG_RE = /^[!-~]+$/;
-// tracingFetch defaults to a 10s timeout, fine for cheap metadata calls (getLatestVersion,
-// createVersion, activate, ...) but too short for a rule-tree PUT/PATCH/GET on a property with many
-// rules: PAPI's `validateRules=true` check is a semantic pass over the whole tree and its latency
-// scales with rule count. A property that legitimately needs >10s gets its request aborted
-// client-side before Akamai can respond, so the version is created but the rule content never
-// gets written — reproduced against a real customer property (all attempts died at exactly
-// "Request timeout after 10000ms"). Give the size-scaling calls a much longer default instead.
+// Rule-tree GET/PUT/PATCH run PAPI's validateRules pass, whose latency scales with tree size and
+// can exceed tracingFetch's 10s default (aborting the write). Give them a longer default.
 const DEFAULT_RULE_TREE_TIMEOUT_MS = 60000;
 
 function requireText(name, value) {
@@ -427,11 +422,8 @@ export default class AkamaiClient {
     }
     const id = encodePathSegment(propertyId);
     const params = { contractId, groupId };
-    // Opt-in: ask PAPI to validate the stored tree and return its errors/warnings arrays. Off by
-    // default (a plain read is cheap); on, PAPI runs its full validation pass so the response
-    // carries `errors`/`warnings` — used to tell whether a version can be activated. Its latency
-    // scales with tree size (the same pass the PUT runs), which is why the size-scaling timeout
-    // applies.
+    // Opt-in: PAPI runs its full validation pass and returns errors/warnings (used to tell whether
+    // a version is activatable). Off by default; slow and size-scaling, so the timeout applies.
     if (options.validateRules) {
       params.validateRules = 'true';
     }
@@ -574,10 +566,7 @@ export default class AkamaiClient {
     }
     const headers = {
       'Content-Type': 'application/json-patch+json',
-      // RFC 7232 If-Match values must be a quoted string; PAPI enforces this strictly and rejects
-      // an unquoted etag with a generic, unhelpful 400 ("The system cannot understand your
-      // request") regardless of what the patch ops say — reproduced empirically: every op
-      // combination failed identically with a bare `If-Match: <etag>`, and succeeded once quoted.
+      // PAPI requires an RFC 7232 quoted etag; a bare token yields a generic, unhelpful 400.
       ...(etag ? { 'If-Match': `"${etag}"` } : {}),
     };
     this.log.info(
--- a/packages/spacecat-shared-akamai-client/src/akamai-client.js
+++ b/packages/spacecat-shared-akamai-client/src/akamai-client.js
@@ -28,6 +28,14 @@ const RULE_FORMAT_RE = /^[a-z0-9-]+$/i;
 // printable-ASCII, whitespace-free token (PAPI returns a hex etag) — this stops a malformed value
 // from injecting extra headers via CR/LF, mirroring RULE_FORMAT_RE's guard for ruleFormat.
 const ETAG_RE = /^[!-~]+$/;
+// tracingFetch defaults to a 10s timeout, fine for cheap metadata calls (getLatestVersion,
+// createVersion, activate, ...) but too short for a rule-tree PUT/PATCH/GET on a property with many
+// rules: PAPI's `validateRules=true` check is a semantic pass over the whole tree and its latency
+// scales with rule count. A property that legitimately needs >10s gets its request aborted
+// client-side before Akamai can respond, so the version is created but the rule content never
+// gets written — reproduced against a real customer property (all attempts died at exactly
+// "Request timeout after 10000ms"). Give the size-scaling calls a much longer default instead.
+const DEFAULT_RULE_TREE_TIMEOUT_MS = 60000;
 
 function requireText(name, value) {
   if (!hasText(value)) {
@@ -165,6 +173,9 @@ export default class AkamaiClient {
    * @param {string} [config.accountSwitchKey]
    * @param {string[]} [config.notifyEmails] - Required only to call activate();
    *   emails PAPI notifies about activation progress.
+   * @param {number} [config.ruleTreeTimeoutMs] - Timeout for calls whose latency scales with
+   *   rule-tree size (getRuleTree, updateRuleTree, patchRuleTree). Defaults to 60s, well above
+   *   tracingFetch's generic 10s default used by every other (cheap, metadata-only) call.
    * @param {object} [log]
    */
   constructor(config = {}, log = console) {
@@ -179,6 +190,7 @@ export default class AkamaiClient {
     this.#accessToken = config.accessToken;
     this.accountSwitchKey = config.accountSwitchKey;
     this.notifyEmails = config.notifyEmails;
+    this.ruleTreeTimeoutMs = config.ruleTreeTimeoutMs || DEFAULT_RULE_TREE_TIMEOUT_MS;
     this.log = log;
   }
 
@@ -191,7 +203,9 @@ export default class AkamaiClient {
     return `https://${this.host}${path}${qs ? `?${qs}` : ''}`;
   }
 
-  async #request(method, path, { params, body, headers } = {}) {
+  async #request(method, path, {
+    params, body, headers, timeout,
+  } = {}) {
     const bodyStr = body ? JSON.stringify(body) : undefined;
 
     // The EG1-HMAC-SHA256 signature is bound to the exact request URL, so a followed redirect
@@ -222,6 +236,7 @@ export default class AkamaiClient {
             ...headers,
           },
           body: bodyStr,
+          ...(timeout ? { timeout } : {}),
         });
       } catch (e) {
         throw new Error(`PAPI ${method} ${path} request failed: ${e.message}`);
@@ -415,7 +430,7 @@ export default class AkamaiClient {
     const data = await this.#request(
       'GET',
       `/papi/v1/properties/${id}/versions/${version}/rules`,
-      { params: { contractId, groupId } },
+      { params: { contractId, groupId }, timeout: this.ruleTreeTimeoutMs },
     );
     // `etag` is the version's optimistic-concurrency token; patchRuleTree passes it back as
     // If-Match so a concurrent edit fails the PATCH instead of silently clobbering.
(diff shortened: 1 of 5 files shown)


packages/spacecat-shared-utils/src/tracing-fetch.js
```diff
@@ -124,17 +124,18 @@ const createTimeoutError = (timeout) => {
...
-const fetchWithTimeout = async (resource, options, signal) => {
+const fetchWithTimeout = async (resource, options, signal, timeout) => {
   try {
     const fetchOptions = { ...options, signal };
     return await adobeFetch(resource, fetchOptions);
   } catch (error) {
     if (error instanceof AbortError) {
-      // Extract timeout from signal (implementation detail, but necessary)
-      // eslint-disable-next-line no-underscore-dangle
-      const timeout = signal._ms || 10000;
```
nit23uec, line 137: This change is backward incompatible.  Why cant we reuse the abort mechanism itself for timeou.

packages/spacecat-shared-akamai-client/src/akamai-client.js
```diff
@@ -28,6 +28,14 @@ const RULE_FORMAT_RE = /^[a-z0-9-]+$/i;
...
 // from injecting extra headers via CR/LF, mirroring RULE_FORMAT_RE's guard for ruleFormat.
 const ETAG_RE = /^[!-~]+$/;
+// tracingFetch defaults to a 10s timeout, fine for cheap metadata calls (getLatestVersion,
+// createVersion, activate, ...) but too short for a rule-tree PUT/PATCH/GET on a property with many
+// rules: PAPI's `validateRules=true` check is a semantic pass over the whole tree and its latency
+// scales with rule count. A property that legitimately needs >10s gets its request aborted
+// client-side before Akamai can respond, so the version is created but the rule content never
+// gets written — reproduced against a real customer property (all attempts died at exactly
+// "Request timeout after 10000ms"). Give the size-scaling calls a much longer default instead.
+const DEFAULT_RULE_TREE_TIMEOUT_MS = 60000;
```
nit23uec, line 33: This wont work for sync calls as API GW has 15 sec limit. We have to convert this to async job will polling.

MysticatBot (changes requested):
Hey @ABHA61,

**Verdict: Request changes** - two correctness issues should be addressed before merge.
**Complexity: MEDIUM** - medium diff across two packages (akamai-client + utils).
_Changes: Adds configurable rule-tree timeout, getRuleTree validateRules option, and fixes the tracing-fetch timeout message to report the actual configured value (5 files)._

## Must fix before merge
1. [Important] `ruleTreeTimeoutMs` uses `||` instead of `??`, silently ignoring explicit `0` - `packages/spacecat-shared-akamai-client/src/akamai-client.js:188` (details inline)
2. [Important] Unconditional etag double-quoting produces malformed header if caller passes already-quoted value - `packages/spacecat-shared-akamai-client/src/akamai-client.js:570` (details inline)

<details>

- nit: JSDoc `@returns` for `getRuleTree` still declares the old shape (missing `errors`/`warnings`) - `packages/spacecat-shared-akamai-client/src/akamai-client.js:416`
- nit: New `options` parameter on `getRuleTree` has no `@param` tag in JSDoc - `packages/spacecat-shared-akamai-client/src/akamai-client.js:418`
- nit: `ruleTreeTimeoutMs` is a public mutable property while the class uses private fields for other config (`#host`, `#clientSecret`, `#accessToken`) - `packages/spacecat-shared-akamai-client/src/akamai-client.js:188`
- suggestion: `createFrom` factory does not read `AKAMAI_RULE_TREE_TIMEOUT_MS` from env, so callers using the standard Lambda factory path cannot configure the timeout without constructing the client manually - `packages/spacecat-shared-akamai-client/src/akamai-client.js`
- nit: Timeout tests use 30ms vs 80ms delay (50ms margin) which is thin for CI under load; consider widening the ratio or using fake timers - `packages/spacecat-shared-akamai-client/test/akamai-client.test.js`
</details>

Note: The backward-incompatibility concern raised on `fetchWithTimeout` is a non-issue - that function is module-private (not exported), so the signature change has no external consumer impact. The API GW 15s limit concern is a valid operational observation about the calling service but outside this PR's scope; the configurable timeout correctly prevents the premature 10s client-side abort.

packages/spacecat-shared-akamai-client/src/akamai-client.js
```diff
@@ -179,6 +185,7 @@ export default class AkamaiClient {
     this.#accessToken = config.accessToken;
     this.accountSwitchKey = config.accountSwitchKey;
     this.notifyEmails = config.notifyEmails;
+    this.ruleTreeTimeoutMs = config.ruleTreeTimeoutMs || DEFAULT_RULE_TREE_TIMEOUT_MS;
```
MysticatBot: issue (blocking): `config.ruleTreeTimeoutMs || DEFAULT_RULE_TREE_TIMEOUT_MS` uses `||`, which treats `0` as falsy and silently substitutes the 60s default. A config store returning `""` or `null` also silently becomes 60000 with no feedback.

Use nullish coalescing (`??`) to default only on `undefined`/`null`, and consider adding a positive-number guard consistent with the `requireText` pattern elsewhere:

```js
this.ruleTreeTimeoutMs = config.ruleTreeTimeoutMs ?? DEFAULT_RULE_TREE_TIMEOUT_MS;
```

Or with validation:
```js
if (config.ruleTreeTimeoutMs != null) {
  if (!Number.isFinite(config.ruleTreeTimeoutMs) || config.ruleTreeTimeoutMs <= 0) {
    throw new Error('ruleTreeTimeoutMs must be a positive number');
  }
}
this.ruleTreeTimeoutMs = config.ruleTreeTimeoutMs ?? DEFAULT_RULE_TREE_TIMEOUT_MS;
```

packages/spacecat-shared-akamai-client/src/akamai-client.js
```diff
@@ -540,7 +566,8 @@ export default class AkamaiClient {
     }
     const headers = {
       'Content-Type': 'application/json-patch+json',
-      ...(etag ? { 'If-Match': etag } : {}),
+      // PAPI requires an RFC 7232 quoted etag; a bare token yields a generic, unhelpful 400.
+      ...(etag ? { 'If-Match': `"${etag}"` } : {}),
```
MysticatBot: issue (blocking): Unconditional `"${etag}"` wrapping produces `""abc123""` on the wire if the caller passes an already-quoted etag value. The existing `ETAG_RE` regex (`/^[!-~]+$/`) accepts `"` (ASCII 34 is in the `!-~` range), so it will not reject an already-quoted value.

The current round-trip path (`getRuleTree().etag` returns `data.etag` from PAPI JSON, which is typically bare) is safe, but the public contract does not enforce this.

Fix: normalize before quoting:
```js
const quoted = etag.startsWith('"') && etag.endsWith('"') ? etag : `"${etag}"`;
...(etag ? { 'If-Match': quoted } : {}),
```

Or document explicitly in the JSDoc/typedef that `etag` must be the bare token (not the HTTP header value) and validate accordingly.

packages/spacecat-shared-utils/src/tracing-fetch.js
ABHA61: Good catch on the signature, but this stays internal-only: `fetchWithTimeout` is a module-private helper (not exported) — the only public export, `tracingFetch(url, options)`, keeps its exact signature, so no caller sees a different API.

And we *are* reusing the abort mechanism — the `signal` (`@adobe/fetch`'s TimeoutSignal) is still the one and only thing that enforces the timeout and fires the abort. The extra `timeout` arg is passed purely so `createTimeoutError` can put the *real* ms value into the message: `@adobe/fetch`'s signal object doesn't expose its own ms back for us to read. Before this, the message hard-coded `10000ms` regardless of the configured timeout — the abort always fired at the right time, only the text was wrong. Happy to drop it if you'd rather the message stay generic, but there's no second timeout path being introduced here.

packages/spacecat-shared-akamai-client/src/akamai-client.js
ABHA61: Agreed the ~15s API-GW/CDN first-byte limit means a synchronous deploy call can't sit and wait for a slow rule-tree write — and that's exactly what the api-service + FE side handle. adobe/spacecat-api-service#3013 adds a read-only `GET .../akamai/deploy-status`, and adobe/project-elmo-ui#2729 polls it: the browser's deploy call does get cut at ~15s, but the Lambda keeps running to completion (up to its 900s cap) and the FE reconciles the true outcome from the poll instead of showing a false failure.

We deliberately chose poll-reconcile over an SQS async job + worker. Unlike `activate` — which rides Akamai's own durable async activation with a re-pollable id — plan/deploy's PAPI calls are one-shot with no server-side job id to hand back, and the EdgeGrid creds are never persisted anywhere, so a separate worker execution would require them to live outside the request Lambda's memory (a security stance we didn't want to change for this).

So `ruleTreeTimeoutMs` isn't trying to beat the gateway limit synchronously — it only stops our *own* client from aborting the write before Akamai responds, so the write lands server-side for the poll to observe. And in #3013 the write is now a small JSON-Patch delta (~4 KB) rather than a full-tree PUT, so it completes in ~30–65s server-side even on large properties.

packages/spacecat-shared-akamai-client/src/akamai-client.js
ABHA61: Fixed in c16e3ea3 — the fallback is now `typeof ms === 'number' && Number.isFinite(ms) && ms > 0 ? ms : DEFAULT_RULE_TREE_TIMEOUT_MS`, so an explicit `0`, `null`, `NaN`, or otherwise-invalid value no longer silently becomes the 60s default.

packages/spacecat-shared-akamai-client/src/akamai-client.js
ABHA61: Fixed in c16e3ea3 — the etag is now stripped of any surrounding quotes before it's wrapped: `` `"${String(etag).replace(/^"+|"+$/g, '')}"` ``. So an already-quoted value from PAPI isn't double-quoted, and a bare token still gets wrapped once. `ETAG_RE` (the CR/LF-injection guard) is unchanged.

MysticatBot (approved):
Hey @ABHA61,

**Verdict: Approve** - both prior blocking findings are addressed; no new issues at blocking severity.
**Complexity: MEDIUM** - medium diff across two packages (akamai-client + utils).
_Changes: Adds configurable rule-tree timeout with validation, getRuleTree validateRules option, etag quote-normalization, and fixes the tracing-fetch timeout message to report the actual configured value (5 files)._

- `ruleTreeTimeoutMs` fallback now uses a proper type-and-range guard instead of `||`
- Etag quoting now strips existing quotes before wrapping, preventing double-quoting

<details>

- nit: `tracingFetch` still uses `options.timeout || 10000` (the same `||` pattern this PR documents as a bug one layer up) - not reachable from this PR's caller since the akamai-client validates, but inconsistent - `packages/spacecat-shared-utils/src/tracing-fetch.js:157`
- suggestion: Add constructor tests for invalid `ruleTreeTimeoutMs` values (0, -1, Infinity, NaN, string) to document the three-conjunct guard contract - `packages/spacecat-shared-akamai-client/test/akamai-client.test.js`
- suggestion: Add a test with an already-quoted etag input to exercise the `replace(/^"+|"+$/g, '')` normalization path - `packages/spacecat-shared-akamai-client/test/akamai-client.test.js:689`
- nit: `ruleTreeTimeoutMs` is a public mutable property while sibling config uses private fields (`#clientToken`, `#clientSecret`) - defense-in-depth observation, not exploitable - `packages/spacecat-shared-akamai-client/src/akamai-client.js:190`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 14s | <b>Cost:</b> $5.99 | <b>Commit:</b> <code>c7887f21fa386aae413710cf4a8cbeee95d6546a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nit23uec (approved):

packages/spacecat-shared-akamai-client/src/akamai-client.js
nit23uec: this call wouldn't return in case it exceeeds GW timeout - and the response will be given by GW directly - we have to differentiate somehow between the genuine errors and timeout errors that will lead to retrying from client.

packages/spacecat-shared-utils/src/tracing-fetch.js
nit23uec: all the usages for fetchWithTimeout are relying on signal._ms for Abort Error handling. This PR removes that.

packages/spacecat-shared-akamai-client/src/akamai-client.js
ABHA61: Good point — yeah, the client does handle that distinction. A genuine Akamai rejection comes back as a `422` with the created `newVersion`, so the FE surfaces it and doesn't retry. A gateway timeout (the 503/504 the GW returns itself, or a dropped connection with no status) is treated as ambiguous — since the Lambda keeps running past the ~15s cut-off — so instead of retrying blind, the FE polls the read-only `deploy-status` endpoint to read the real Akamai state and get the definitive result.

The classifier is `isAmbiguousDeployFailure()`: `5xx / 408 / no-status → poll`, `4xx or 422-with-newVersion → surface`. Retries also resume into the already-created version via `retryVersion`, so a re-attempt never mints a duplicate. This all lives on the api-service + FE side (#3013 / adobe/project-elmo-ui#2729).

packages/spacecat-shared-utils/src/tracing-fetch.js
ABHA61: Thanks — I looked into this. `fetchWithTimeout` is a private helper here (not exported, only called from `tracingFetch`), so the extra argument isn't a breaking change for any caller.

On `signal._ms`: that field is actually always `undefined` in this version of `@adobe/fetch` — the timeout is kept in a private internal symbol, not `_ms`. So the old `signal._ms || 10000` was always falling back to `10000`, which is why the timeout message always said "10000ms" regardless of the configured value. The abort itself always fired correctly; only the message was wrong. I just sourced the value from the `timeout` that's already in scope to fix the message.

Happy to switch to keeping it on the signal instead if you'd prefer — just let me know.

PR #1865, opened 2026-08-05
fix: optimize edge deploy/rollback api

dipratap:
**What**
The edge-deploy and edge-rollback APIs were creeping up toward the 15s gateway timeout. Digging in, the culprit is the per-URL loop in deploySuggestions and rollbackSuggestions - we fetch, patch, and upload each URL's config to S3 one at a time. That's N × (GET + PUT) in series, so latency scales linearly with the number of URLs in an opportunity.

Since every URL writes to its own distinct S3 object, the work is fully independent. This PR runs those per-URL operations in parallel with a bounded concurrency cap (5) instead of sequentially.

**Changes**
Added a small mapWithConcurrency helper (worker-pool, order-preserving, bounded).
Parallelized the per-URL loop in deploySuggestions and rollbackSuggestions.
Left the shared-metaconfig read-modify-write and CDN invalidation untouched — those are already single/batched and can't be parallelized safely.

### DIFF
--- a/packages/spacecat-shared-tokowaka-client/src/index.js
+++ b/packages/spacecat-shared-tokowaka-client/src/index.js
@@ -40,6 +40,7 @@ import {
 import { getEffectiveBaseURL } from './utils/site-utils.js';
 import { removePatternFromMetaconfig, addPatternsToMetaconfig } from './utils/metaconfig-utils.js';
 import { fetchHtmlWithWarmup, calculateForwardedHost } from './utils/custom-html-utils.js';
+import { mapWithConcurrency } from './utils/concurrency-utils.js';
 import {
   EDGE_OPTIMIZE_PROXY_BASE_URL_DEFAULT,
   PRIVATE_HOST_RE,
@@ -73,6 +74,13 @@ export {
 const HTTP_BAD_REQUEST = 400;
 const HTTP_INTERNAL_SERVER_ERROR = 500;
 const HTTP_NOT_IMPLEMENTED = 501;
+
+/**
+ * Max number of per-URL S3 config operations (fetch + upload) to run in parallel
+ * during deploy/rollback. Each URL maps to a distinct S3 object, so the work is
+ * independent; the cap avoids exhausting the S3 socket pool on large opportunities.
+ */
+const URL_CONFIG_CONCURRENCY = 10;
 
 /**
  * Tokowaka Client - Manages edge optimization configurations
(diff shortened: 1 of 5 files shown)


packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -970,24 +982,31 @@ class TokowakaClient {
...
+          opportunity,
+          mapper,
+          opportunityType,
+          s3Paths,
+          rolledBackUrls,
+        );
+      },
+      URL_CONFIG_CONCURRENCY,
+    );
+    totalRemovedCount += removedCounts.reduce((sum, n) => sum + n, 0);
```
MysticatBot: issue (blocking): `s3Paths` and `rolledBackUrls` are declared outside `mapWithConcurrency` and mutated via `.push()` from within concurrent workers in `#rollbackPerUrlConfig`. While safe in Node.js single-threaded concurrency (no memory corruption), this produces non-deterministic array ordering and is inconsistent with the deploy path, which correctly returns values from the mapper and collects them afterward.

The deploy path was refactored cleanly:
```js
return { s3Path, fullUrl };
// then post-loop: extract from processed results
```

The rollback path should follow the same pattern. Have `#rollbackPerUrlConfig` return `{ s3Path, fullUrl, removed }` instead of mutating passed-in arrays, then assemble `s3Paths`/`rolledBackUrls`/`totalRemovedCount` from results after `mapWithConcurrency` completes.

packages/spacecat-shared-tokowaka-client/src/utils/concurrency-utils.js
```diff
@@ -0,0 +1,54 @@
...
+ * in other lanes: as soon as one task finishes, the next queued item starts.
+ *
+ * The first rejection propagates (like Promise.all); use this only when a single
+ * failure should abort the whole batch.
+ *
+ * @template T, R
+ * @param {Array<T>} items - Items to process
+ * @param {(item: T, index: number) => Promise<R>} mapper - Async mapper per item
+ * @param {number} [concurrency=10] - Max number of tasks running at once
+ * @returns {Promise<Array<R>>} - Results in input order
```
MysticatBot, line 27: issue (blocking): JSDoc says `@param {number} [concurrency=10]` but the function signature defaults to `concurrency = 5` (line 30). The test titled "defaults concurrency to 10 when not provided" uses `at.most(10)` which passes with 5, so the mismatch is not caught.

Fix: update JSDoc to `@param {number} [concurrency=5]` and rename the test to match.

MysticatBot (changes requested):
Hey @dipratap,

**Verdict: Request changes** - two issues to address before merge.
**Complexity: MEDIUM** - medium diff, single-service performance optimization.
_Changes: Parallelizes per-URL S3 config operations in deploy and rollback paths using a bounded-concurrency worker pool, and normalizes URL paths to prevent duplicate S3 writes (5 files)._

## Must fix before merge
1. [Important] Shared mutable arrays pushed from concurrent workers in rollback path - `src/index.js:~988` (details inline)
2. [Important] JSDoc documents wrong default value for concurrency parameter - `src/utils/concurrency-utils.js:28` (details inline)

<details>

- nit: Test "defaults concurrency to 10 when not provided" asserts `at.most(10)` which passes with actual default of 5; title and assertion should match reality - `test/utils/concurrency-utils.test.js:69`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 9s | <b>Cost:</b> $5.74 | <b>Commit:</b> <code>25013988c581e2f4191e231f35a61615ae621021</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @dipratap,

**Verdict: Approve** - all prior findings addressed; no new blocking issues.
**Complexity: MEDIUM** - medium diff, single-service performance optimization.
_Changes: Parallelizes per-URL S3 config operations in deploy and rollback paths using a bounded-concurrency worker pool, normalizes URL paths to prevent duplicate S3 writes, and refactors rollback helper to return results instead of mutating accumulators (5 files)._

<details>

- suggestion: `mapWithConcurrency` does not guard against `NaN` concurrency - `Math.max(1, NaN)` evaluates to `NaN`, producing zero workers and silently returning an array of `undefined`. Add a `Number.isFinite` check or throw on invalid input for safer reuse - `packages/spacecat-shared-tokowaka-client/src/utils/concurrency-utils.js:34`
- nit: The `normalizePath` change in `suggestion-utils.js` is a separate behavioral fix (merging trailing-slash URL variants) bundled with the concurrency work - a separate commit would give cleaner bisect/revert - `packages/spacecat-shared-tokowaka-client/src/utils/suggestion-utils.js:56`
- suggestion: Consider exporting `mapWithConcurrency` from the package barrel if it may be useful to other consumers of the tokowaka-client package - `packages/spacecat-shared-tokowaka-client/src/utils/concurrency-utils.js:30`
</details>

- Shared mutable arrays pushed from concurrent workers in rollback path - `#rollbackPerUrlConfig` now returns a result object
- JSDoc documented wrong default value for concurrency parameter - now correctly says `[concurrency=5]`
- Test title and assertion mismatch on default concurrency - title and `expect` now both reference 5


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 5s | <b>Cost:</b> $5.51 | <b>Commit:</b> <code>e34ee4d6d7d344bea3cf90b6c3f51011f5dc36af</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nit23uec (approved):

PR #1871, opened 2026-08-11
feat: schema changes for trial users entity to allow searching by external user id

vivesing:
## PR Summary: Add `externalUserId` lookup for TrialUser

**Goal:** Expose the `external_user_id` DB index (added in mysticat-data-service) through the JS data-access layer for `TrialUser`, so callers can look up trial users by their external IDP user ID.

### Review feedback addressed

- **Blocking:** `index.d.ts` was missing type declarations for the new accessors (TS consumers would get a compile error) — fixed.
- **Suggestion:** documented that `externalUserId` isn't guaranteed unique, so `findBy*` returns an arbitrary match — callers needing all matches should use `allBy*`.
- **Nit:** added the not-found IT assertion to pin the `null` contract.
- **Nit:** flagged via comment thaf the 5-index cap enforced by`SchemaBuilder`, so the next lookup index needs budget consideration.

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts
+++ b/packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts
@@ -43,9 +43,19 @@ export interface TrialUserCollection extends BaseCollection<TrialUser> {
   allByProviderAndExternalUserId(provider: ProviderType, externalId: string): Promise<TrialUser[]>;
   allByOrganizationId(organizationId: string): Promise<TrialUser[]>;
   allByEmailId(emailId: string): Promise<TrialUser[]>;
+  allByExternalUserId(externalUserId: string): Promise<TrialUser[]>;
+  allByExternalUserIdAndUpdatedAt(
+    externalUserId: string,
+    updatedAt: string
+  ): Promise<TrialUser[]>;
   findByProvider(provider: ProviderType): Promise<TrialUser | null>;
   findByProviderAndExternalUserId(provider: ProviderType, externalId: string):
     Promise<TrialUser | null>;
   findByOrganizationId(organizationId: string): Promise<TrialUser | null>;
   findByEmailId(emailId: string): Promise<TrialUser | null>;
+  findByExternalUserId(externalUserId: string): Promise<TrialUser | null>;
+  findByExternalUserIdAndUpdatedAt(
+    externalUserId: string,
+    updatedAt: string
+  ): Promise<TrialUser | null>;
 }
--- a/packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts
+++ b/packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts
@@ -53,7 +53,17 @@ export interface TrialUserCollection extends BaseCollection<TrialUser> {
     Promise<TrialUser | null>;
   findByOrganizationId(organizationId: string): Promise<TrialUser | null>;
   findByEmailId(emailId: string): Promise<TrialUser | null>;
+  /**
+   * externalUserId is optional and not enforced unique (no DB unique constraint), so this
+   * returns one arbitrary match when multiple trial users share the same externalUserId.
+   * Use allByExternalUserId if you need every matching record.
+   */
   findByExternalUserId(externalUserId: string): Promise<TrialUser | null>;
+  /**
+   * externalUserId is optional and not enforced unique (no DB unique constraint), so this
+   * returns one arbitrary match when multiple trial users share the same externalUserId.
+   * Use allByExternalUserIdAndUpdatedAt if you need every matching record.
+   */
   findByExternalUserIdAndUpdatedAt(
     externalUserId: string,
     updatedAt: string
--- a/packages/spacecat-shared-data-access/test/it/trial-user/trial-user.test.js
+++ b/packages/spacecat-shared-data-access/test/it/trial-user/trial-user.test.js
@@ -88,6 +88,12 @@ describe('TrialUser IT', async () => {
     );
   });
 
+  it('returns null when no trial user matches the external user id', async () => {
+    const trialUser = await TrialUser.findByExternalUserId('nonexistent-id');
+
+    expect(trialUser).to.be.null;
+  });
+
   it('gets all trial users by organization id', async () => {
     const sampleTrialUser = sampleData.trialUsers[0];
     const organizationId = sampleTrialUser.getOrganizationId();
--- a/packages/spacecat-shared-data-access/test/it/trial-user/trial-user.test.js
+++ b/packages/spacecat-shared-data-access/test/it/trial-user/trial-user.test.js
@@ -94,6 +94,54 @@ describe('TrialUser IT', async () => {
     expect(trialUser).to.be.null;
   });
 
+  it('gets all/one trial users sharing an externalUserId, pinning the arbitrary-match contract', async () => {
+    const organizationId = sampleData.organizations[0].getId();
+    const sharedExternalUserId = 'ext-user-shared';
+
+    const first = await TrialUser.create({
+      organizationId,
+      emailId: 'shared1@example.com',
+      externalUserId: sharedExternalUserId,
+      status: 'INVITED',
+      updatedBy: 'system',
+    });
+    const second = await TrialUser.create({
+      organizationId,
+      emailId: 'shared2@example.com',
+      externalUserId: sharedExternalUserId,
+      status: 'INVITED',
+      updatedBy: 'system',
+    });
+
+    const allMatches = await TrialUser.allByExternalUserId(sharedExternalUserId);
+    expect(allMatches).to.be.an('array');
+    expect(allMatches.map((trialUser) => trialUser.getId())).to.have.members([
+      first.getId(),
+      second.getId(),
+    ]);
+
+    const singleMatch = await TrialUser.findByExternalUserId(sharedExternalUserId);
+    expect(singleMatch).to.be.an('object');
+    expect([first.getId(), second.getId()]).to.include(singleMatch.getId());
+
+    const secondUpdatedAt = second.getUpdatedAt();
+
+    const allByUpdatedAt = await TrialUser.allByExternalUserIdAndUpdatedAt(
+      sharedExternalUserId,
+      secondUpdatedAt,
+    );
+    expect(allByUpdatedAt).to.be.an('array');
+    expect(allByUpdatedAt.length).to.equal(1);
+    expect(allByUpdatedAt[0].getId()).to.equal(second.getId());
+
+    const findByUpdatedAt = await TrialUser.findByExternalUserIdAndUpdatedAt(
+      sharedExternalUserId,
+      secondUpdatedAt,
+    );
+    expect(findByUpdatedAt).to.be.an('object');
+    expect(findByUpdatedAt.getId()).to.equal(second.getId());
+  });
+
   it('gets all trial users by organization id', async () => {
     const sampleTrialUser = sampleData.trialUsers[0];
     const organizationId = sampleTrialUser.getOrganizationId();
--- a/packages/spacecat-shared-data-access/src/models/trial-user/trial-user.schema.js
+++ b/packages/spacecat-shared-data-access/src/models/trial-user/trial-user.schema.js
@@ -53,6 +53,11 @@ const schema = new SchemaBuilder(TrialUser, TrialUserCollection)
     { composite: ['organizationId'] },
     { composite: ['updatedAt'] },
   )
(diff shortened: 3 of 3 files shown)


packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts
```diff
@@ -43,9 +43,19 @@ export interface TrialUserCollection extends BaseCollection<TrialUser> {
...
   findByProvider(provider: ProviderType): Promise<TrialUser | null>;
   findByProviderAndExternalUserId(provider: ProviderType, externalId: string):
     Promise<TrialUser | null>;
   findByOrganizationId(organizationId: string): Promise<TrialUser | null>;
   findByEmailId(emailId: string): Promise<TrialUser | null>;
+  findByExternalUserId(externalUserId: string): Promise<TrialUser | null>;
+  findByExternalUserIdAndUpdatedAt(
+    externalUserId: string,
+    updatedAt: string
+  ): Promise<TrialUser | null>;
```
MysticatBot, line 70: issue (blocking): Return type should be `Promise<TrialUser | null>`, not `Promise<TrialUser[]>`.

All other `findBy*` methods in this interface return a single entity or null. The `allBy*` variant (correctly typed above) returns the array. As written, TypeScript consumers will get a type that contradicts the runtime behavior of the auto-generated `findBy*` accessor from SchemaBuilder.

**Fix:** Change the return type to match the convention:
```typescript
findByExternalUserIdAndUpdatedAt(
  externalUserId: string,
  updatedAt: string
): Promise<TrialUser | null>;
```

MysticatBot (changes requested):
Hey @vivesing,

**Verdict: Request changes** - type declaration has a return-type bug that will break TS consumers.
**Complexity: LOW** - small diff, single entity, single package.
_Changes: Adds an `externalUserId` lookup index to the TrialUser entity with corresponding type declarations and a not-found integration test (3 files)._

## Must fix before merge
1. [Important] `findByExternalUserIdAndUpdatedAt` return type is `Promise<TrialUser[]>` but should be `Promise<TrialUser | null>` per `findBy*` convention - `index.d.ts:60` (details inline)

<details>

- suggestion: Add a positive-match integration test for `findByExternalUserId` (and ideally the compound `*AndUpdatedAt` accessor) to prove the index resolves records, not just the not-found path - `test/it/trial-user/trial-user.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 59s | <b>Cost:</b> $2.95 | <b>Commit:</b> <code>f1233f550323bf889c3487b14ab1dc1b79de4c47</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts
vivesing: already correct retrun type

packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts
```diff
@@ -43,9 +43,19 @@ export interface TrialUserCollection extends BaseCollection<TrialUser> {
...
+  allByExternalUserIdAndUpdatedAt(
+    externalUserId: string,
+    updatedAt: string
+  ): Promise<TrialUser[]>;
   findByProvider(provider: ProviderType): Promise<TrialUser | null>;
   findByProviderAndExternalUserId(provider: ProviderType, externalId: string):
     Promise<TrialUser | null>;
   findByOrganizationId(organizationId: string): Promise<TrialUser | null>;
   findByEmailId(emailId: string): Promise<TrialUser | null>;
+  findByExternalUserId(externalUserId: string): Promise<TrialUser | null>;
```
calvarezg, line 61: issue (blocking): The non-uniqueness caveat is invisible to TypeScript consumers.

`externalUserId` is optional and not enforced unique, so `findByExternalUserId` / `findByExternalUserIdAndUpdatedAt` return one arbitrary match. That warning currently lives only as a comment in `trial-user.schema.js`. Consumers import the typed interface from `index.d.ts` and never open the schema file, so the caveat never reaches editor hovers or generated docs.

Fix: add a JSDoc block above `findByExternalUserId` and `findByExternalUserIdAndUpdatedAt` here carrying the same caveat and pointing callers to `allByExternalUserId` when they need every match. Note: no `.d.ts` in this package currently uses JSDoc, so this is a small net-new convention rather than matching an existing one.

packages/spacecat-shared-data-access/test/it/trial-user/trial-user.test.js
```diff
@@ -88,6 +88,12 @@ describe('TrialUser IT', async () => {
     );
   });
 
+  it('returns null when no trial user matches the external user id', async () => {
```
calvarezg, line 91: issue (blocking): Three of the four newly-declared accessors have no test coverage.

This PR declares `allByExternalUserId`, `allByExternalUserIdAndUpdatedAt`, `findByExternalUserId`, and `findByExternalUserIdAndUpdatedAt` in `index.d.ts`. Only `findByExternalUserId` is exercised (a pre-existing positive test plus this PR's new not-found case). A repo-wide search finds no test touching the other three. Because `index.d.ts` is hand-maintained and JS callers ignore it, a wrong param order or a dropped sort-key composite would be caught by nothing - not a test, not the type checker.

Fix: add at least one IT per untested accessor, following the sibling `allByOrganizationId` test as a template. A single case that seeds two trial users sharing an `externalUserId`, then asserts `allByExternalUserId` returns both while `findByExternalUserId` returns exactly one, also pins the documented "arbitrary match" contract.

calvarezg (changes requested):
Hey @vivesing,

**Verdict: Request changes** - the change is low-risk (types, comments, one test), but two items should land before merge.
**Complexity: MEDIUM** - small diff, but it touches the data-access model/schema surface.
_Changes: Adds hand-written `index.d.ts` type declarations for the `externalUserId` lookup accessors, explanatory schema comments, and a not-found integration test; the underlying index itself already shipped in the merged sibling PR 1869 (3 files)._

## Must fix before merge
1. [Important] `externalUserId` non-uniqueness caveat is missing from the consumer-facing type surface - `packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts:56` (details inline)
2. [Important] Three of the four newly-declared accessors have no test coverage - `packages/spacecat-shared-data-access/test/it/trial-user/trial-user.test.js:91` (details inline)

<details>

- nit: PR title/description says "schema changes ... to allow searching by external user id", but the index already existed on `main` (shipped in the merged sibling PR); this diff is type declarations, comments, and one test. Correct the description so reviewers and ops do not assume a data-layer migration.
- suggestion: `externalUserId` has no `validate` (max length / character allowlist), unlike sibling attributes in the same schema; it originates from an external IDP claim, so a cheap validator would harden the write path - `packages/spacecat-shared-data-access/src/models/trial-user/trial-user.schema.js`
- nit: `findByExternalUserIdAndUpdatedAt` / `allByExternalUserIdAndUpdatedAt` match exactly on `updatedAt`, which is server-set on every write, so a caller cannot supply it without already holding the record; a one-line note on intended use (e.g. pagination continuation) would help - `packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts`
- nit: the index-budget comment does not note that `addReference('belongs_to', ...)` also consumes the same 5-index budget; one clause would prevent a future author silently exceeding it - `packages/spacecat-shared-data-access/src/models/trial-user/trial-user.schema.js`
</details>

### Out of scope, worth tracking
- The `externalUserId` lookup has no `provider` dimension, so it cannot disambiguate which IDP issued an id; two providers can assign the same subject id to different people. Provider scoping was removed org-wide previously, and callers live in other repos, so this is a follow-up design conversation, not a change for this PR.
- `index.d.ts` still declares `getProvider` / `setProvider` / `allByProvider*` / `findByProvider*`, which reference a `provider` attribute the schema no longer defines. This is pre-existing drift (not touched by this diff) but is a live compile-green / runtime-crash trap for TS consumers. No CI check catches `.d.ts`-vs-schema drift in either direction.
- The explicit `organizationId` index duplicates the auto-generated `belongs_to` index composite, burning one index slot for no extra query capability (pre-existing).
- The "5-index cap" comment reflects a DynamoDB-era model; the backend is now Postgres via PostgREST, so the cap is a self-imposed constant rather than a live GSI-per-table limit. Worth re-justifying or re-framing.
- The DB-level uniqueness stance for `external_user_id` lives in `mysticat-data-service` (different repo); confirm it does not contradict the "not enforced unique" caveat.

- Prior return-type finding on the compound accessor - current code returns `Promise<TrialUser | null>`, consistent with the `findBy*` convention.

packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts
vivesing: fixed

packages/spacecat-shared-data-access/test/it/trial-user/trial-user.test.js
vivesing: fixed

calvarezg (approved):
Hey @vivesing,

**Verdict: Approve** - both prior blocking findings are addressed; only two optional test nits remain.
**Complexity: MEDIUM** - small diff, but it touches the data-access model/schema surface.
_Changes: Adds hand-written `index.d.ts` type declarations for the `externalUserId` lookup accessors, JSDoc documenting the non-unique arbitrary-match contract, schema comments, and integration tests covering all four accessors; the index itself shipped in the merged sibling PR 1869 (3 files)._

<details>

- suggestion: the `allByExternalUserIdAndUpdatedAt` assertion `length === 1` relies on the two `create()` calls getting distinct server-set `updatedAt` values. A same-millisecond collision is unlikely given the PostgREST round-trip between them, but not impossible; assert `expect(second.getUpdatedAt()).to.not.equal(first.getUpdatedAt())` first, or compare by ID membership, to remove the latent flake - `packages/spacecat-shared-data-access/test/it/trial-user/trial-user.test.js`
- suggestion: the two records the new test creates are not removed and persist for the rest of the file's run. No current assertion breaks (the shared-organizationId test only checks length > 0 and org match), but a future exact-count test on that org would break non-obviously; add `await first.remove(); await second.remove();` at the end, matching the create/remove pattern used elsewhere in the suite - `packages/spacecat-shared-data-access/test/it/trial-user/trial-user.test.js`
</details>

- Non-uniqueness caveat is now on the consumer-facing type surface - JSDoc added above `findByExternalUserId` and `findByExternalUserIdAndUpdatedAt` in `index.d.ts`, matching the schema comment.
- The three previously-untested accessors (`allByExternalUserId`, `allByExternalUserIdAndUpdatedAt`, `findByExternalUserIdAndUpdatedAt`) now have real behavioral assertions in a new integration test that seeds two trial users sharing an `externalUserId`.

MysticatBot (approved):
Hey @vivesing,

**Verdict: Approve** - both prior blocking findings are addressed; only optional polish remains.
**Complexity: LOW** - small diff, single entity, single package.
_Changes: Adds externalUserId lookup index to TrialUser with type declarations, JSDoc non-uniqueness caveats, and integration tests covering all four generated accessors (3 files)._

<details>

- nit: JSDoc on `findByExternalUserIdAndUpdatedAt` is copy-pasted from `findByExternalUserId` without adjusting for the narrower compound query - when both `externalUserId` and `updatedAt` are provided, a collision requires two users sharing the same external ID AND the same millisecond-precision timestamp, making the "arbitrary match" caveat overstated for this method - `packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts:65`
- suggestion: Consider documenting that `allByExternalUserId` with `options.between` on `updatedAt` is the idiomatic pattern for time-range queries against this index, since the generated `*AndUpdatedAt` accessors use exact-match equality only - `packages/spacecat-shared-data-access/src/models/trial-user/index.d.ts`
</details>

- Return-type bug on `findByExternalUserIdAndUpdatedAt` now correctly returns `Promise<TrialUser | null>`.
- Non-uniqueness caveat is now on the consumer-facing type surface via JSDoc.
- All four new accessors now have behavioral integration tests pinning the arbitrary-match contract.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 5s | <b>Cost:</b> $4.08 | <b>Commit:</b> <code>c263a6218f5702ec99c7819ffc4d26c8d44d7a17</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1875, opened 2026-08-11
fix(seo-client): scope broken-backlinks to sub-path via target_url filter

rpapani:
## Problem (SITES-49721)

`getBrokenBacklinksV2` produced whole-domain results for **sub-path sites** (e.g. `example.gov/us`), so a sub-path site saw broken backlinks for the entire domain instead of only its own section.

Root cause (confirmed against the live Semrush API):

- The `/apis/v4/backlinks-external/v0/broken-links` endpoint with `scope=ROOT_DOMAIN` **ignores the path** in the `url` param and returns whole-domain broken backlinks.
- `scope=SUBFOLDER` is **not supported** — the endpoint rejects it with HTTP 400.
- To scope to a sub-path you must pass the **hostname** as `url` **and** add a server-side `target_url LIKE '%<hostname><path>%'` filter.

## Changes to `getBrokenBacklinksV2`

1. **Derive `{ hostname, pathname }` once** from the input `url` (via `prependSchema` + `new URL(...)`, `stripWWW` on the host) and send the **hostname (domain)** as the `url` param — never the path-carrying value.
2. **Guard the URL parse.** A malformed non-empty `url` previously threw. It now logs a warn, queries the whole domain, and adds no sub-path clause — mirroring the existing `getTopPages` precedent in the same file. Falls back to using the raw `url` as the request param.
3. **Append `AND target_url LIKE '%<hostname><path>%'`** only when there is a real sub-path. A single trailing slash on the path is normalized first, so `example.com/foo` and `example.com/foo/` scope identically; a root path (`/`) adds no clause, so root-domain behavior is unchanged.
4. **Remove `response_code=200`** from the filter. That clause matched the **source** page's response code (always 200 for a live source), a proven no-op that does **not** exclude broken 4xx targets — removing it changes result counts by 0. The endpoint is inherently "broken targets only".
5. **404 visibility**: the 404 (no data / not indexed) mapping now logs a distinct **WARNING** (`[SEO] broken-links 404 (no data / not indexed) for <url> — returning empty; may be transient`) instead of info, so a transient "no data" is not silently indistinguishable from a genuine empty result. It still returns the empty result (no throw).

## Tests

Extended the `getBrokenBacklinksV2` suite (neutral example URLs): sub-path url adds the `target_url LIKE` clause and sends the hostname as `url`; root url adds no clause; a trailing slash normalizes to the same clause; a malformed url falls back to the raw value with no clause and logs a warn; `response_code=200` is absent from the emitted filter; the 404 path logs the new warning (and not info). Package coverage stays at 100% lines/branches/statements (191 passing).

## Release / companion PR

- The coarse server-side substring match here is refined to a **precise directory boundary** downstream in the companion **adobe/spacecat-audit-worker** PR (e.g. `%example.com/foo%` would otherwise also match `/foo-bar`). **This should not be released independently** of that companion change.
- **Known limitation**: an underscore in a path is a SQL `LIKE` single-char wildcard, so a path containing `_` is a minor over-match. LIKE metacharacters are intentionally not escaped here (Semrush filter escaping semantics are unverified); the downstream directory-boundary filter refines this.

### DIFF
--- a/packages/mysticat-shared-seo-client/src/client.js
+++ b/packages/mysticat-shared-seo-client/src/client.js
@@ -10,7 +10,9 @@
  * governing permissions and limitations under the License.
  */
 
-import { hasText, isValidUrl, isArray } from '@adobe/spacecat-shared-utils';
+import {
+  hasText, isValidUrl, isArray, prependSchema, stripWWW,
+} from '@adobe/spacecat-shared-utils';
 import { context as h2, h1 } from '@adobe/fetch';
 
 import { ENDPOINTS } from './endpoints.js';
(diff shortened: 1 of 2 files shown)


FentPams (approved):
✅ Approve — sending the hostname as `url` + the `target_url LIKE` scope is a sound server-side pre-filter, and dropping the no-op `response_code=200` filter plus the 404→`warn` upgrade are good calls. Coverage is thorough.

One recommended follow-up (non-blocking — spacecat-audit-worker#2882 enforces the real directory boundary downstream, and `url` is config/operator-sourced, not user input): `pathname` is interpolated raw into `'%${hostname}${pathname}%'` with no LIKE-escaping. `_` and `%` are LIKE wildcards, so an ordinary path like `/foo_bar` over-matches `/fooXbar`, and a `'` in a path would break the filter string. The over-match is harmless (it's boundary-filtered in #2882), but escaping `_`/`%`/`'` would make the pre-filter tighter and more robust. Please land this together with #2882 so the coarse match is always refined.

PR #1876, opened 2026-08-13
fix(html-analyzer): detect Cookiebot (Cybot) consent dialogs

Danialkafeel:
## Summary
- Cookiebot (Cybot/Usercentrics) consent banners use `id="CybotCookiebotDialog"` with `role="region"` + `aria-labelledby`, which didn't match any existing class/ID/ARIA selector in the cookie-banner filter — so the banner text leaked into extracted page content.
- Added `[id^="CybotCookiebot"]` to the ID selector list to catch it.

## Test plan
- [ ] `npm test -w packages/spacecat-shared-html-analyzer`
- [x] Added unit test reproducing the `CybotCookiebotDialog` structure, asserting its text is stripped


packages/spacecat-shared-html-analyzer/test/index.test.js
```diff
@@ -217,6 +217,24 @@ describe('HTML Visibility Analyzer', () => {
       expect(text).to.not.include('Manage consent preferences');
     });
 
+    it('should remove Cybot Cookiebot dialog', async () => {
```
MysticatBot, line 220: issue (blocking): The test uses `id="CybotCookiebotDialog"` which is already matched by the pre-existing `#CybotCookiebotDialog` entry in `COOKIE_BANNER_ID_SELECTORS`. This test would pass even if the new `[id^="CybotCookiebot"]` prefix selector were removed.

To actually validate the prefix selector, use a Cookiebot ID that is NOT already in the list - e.g. `CybotCookiebotWidget` or `CybotCookiebotOverlay`.

Additionally, @khushboo5723's feedback about the selector approach (using explicit IDs + class selectors consistent with how OneTrust is handled) is worth incorporating before merge.

MysticatBot (changes requested):
Hey @Danialkafeel,

**Verdict: Request changes** - the test does not validate the new selector, and the existing reviewer feedback about approach should be addressed.
**Complexity: TRIVIAL** - 1-line selector addition + test, 2 files.
_Changes: Adds an attribute prefix selector for Cookiebot consent banners to the cookie-banner filter list (2 files)._

## Must fix before merge
1. [Important] Test does not validate the new prefix selector - `packages/spacecat-shared-html-analyzer/test/index.test.js:220` (details inline)

<details>

- nit: Redundant selector pair - `[id^="CybotCookiebot"]` is a superset of `#CybotCookiebotDialog` on the next line; one should be removed - `packages/spacecat-shared-html-analyzer/src/html-filter.js:72`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 34s | <b>Cost:</b> $2.39 | <b>Commit:</b> <code>81e70992984fe5b557140d73dc6c3785b521e573</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1877, opened 2026-08-13
feat(data-access): paginated sites query filtered by entitlement tier/productCode

bottadobe:
## Summary

  Adds `SiteCollection.allByEnrollmentFiltered(...)` — a single-query, paginated
  way to fetch sites filtered by entitlement **tier** and/or **product code**,
  composable with a caller-supplied `where` and `orderBy`. Intended for the
  api-service sites listing that needs tier/product filtering with stable
  pagination.

  ## What's new
  
  `allByEnrollmentFiltered({ tier, productCode }, { where, orderBy, limit, cursor, returnCursor })` on `SiteCollection`:

  - Filters via a two-level inner-join embed (`sites -> site_enrollments -> entitlements`) in a **single PostgREST query** — no N+1 / batch-fetch round-trips.
  - Requires at least one of `tier` / `productCode` (throws `DataAccessError` otherwise).
  - Returns hydrated `Site` instances; the embedded enrollments array is stripped before hydration so it can't leak onto the record.
  - Offset-cursor pagination: honors the **exact** `limit` (no silent cap, no +1) and returns `{ data, cursor }` when `returnCursor` is set, else a bare array — kept symmetric with `Site.all`.
  - Default sort `updatedAt desc` plus an `id` tiebreaker for stable page boundaries; validates `orderBy` (unknown attribute / invalid direction throw a clear `DataAccessError`), matching `Site.all` / base `#queryPage`.

  ## Why embedding (not a flat join)

  PostgREST resource embedding nests matching children under each parent row
  instead of flattening into a cartesian product, so `.range()` offset pagination
  stays correct over DISTINCT parent sites even when a site has multiple matching
  enrollments. `!inner` at both embed levels turns the embedded filters into INNER
  JOINs, excluding sites with no matching enrollment/entitlement.

  ## Testing
  
  - **22 unit tests** (stubbed PostgREST chain): tier/product filters, `where` composition, ordering + id tiebreaker, cursor/limit math, hydration + embed-stripping, error handling, and `orderBy` validation.
  - **8 integration tests** against seeded fixtures: tier / product / combined filtering, `where` composition, distinct-parent pagination, dedup across multiple matching enrollments, and the empty + validation cases.
  - `npm run lint` and `npm test` (data-access) green; coverage thresholds met.

### DIFF
--- a/packages/spacecat-shared-data-access/src/models/site/site.collection.js
+++ b/packages/spacecat-shared-data-access/src/models/site/site.collection.js
@@ -244,7 +244,8 @@ class SiteCollection extends BaseCollection {
    *   applied to the sites table (e.g. baseURL substring / deliveryType / isLive).
    * @param {object} [options.orderBy] - `{ attribute, direction }`; defaults to
    *   `updatedAt` desc. Always followed by an `id` tiebreaker for stable paging.
-   * @param {number} [options.limit] - Max parent rows to return (exact).
+   * @param {number} [options.limit] - Max parent rows to return (exact). A
+   *   non-positive or non-integer value falls back to DEFAULT_PAGE_SIZE.
    * @param {string} [options.cursor] - Base64 offset cursor (see decodeCursor).
    * @param {boolean} [options.returnCursor] - Return `{ data, cursor }` shape.
    * @returns {Promise<Site[] | { data: Site[], cursor: string|null }>}
@@ -307,9 +308,11 @@ class SiteCollection extends BaseCollection {
       query = query.order(idField, { ascending });
     }
 
-    // Honor the exact limit (no cap, no +1) so the api-service N+1 hasMore
-    // detection stays symmetric with Site.all's postgrest path.
-    const effectiveLimit = Number.isInteger(limit) ? limit : DEFAULT_PAGE_SIZE;
+    // Honor the exact (positive) limit (no cap, no +1) so the api-service N+1
+    // hasMore detection stays symmetric with Site.all's postgrest path. A
+    // non-positive or non-integer limit falls back to DEFAULT_PAGE_SIZE so a
+    // caller-supplied 0/negative can't produce an inverted PostgREST range.
+    const effectiveLimit = Number.isInteger(limit) && limit > 0 ? limit : DEFAULT_PAGE_SIZE;
     const offset = decodeCursor(cursor);
     query = query.range(offset, offset + effectiveLimit - 1);
 
--- a/packages/spacecat-shared-data-access/src/models/site/site.collection.js
+++ b/packages/spacecat-shared-data-access/src/models/site/site.collection.js
@@ -14,6 +14,13 @@ import { hasText, isValidHelixPreviewUrl, isValidUrl } from '@adobe/spacecat-sha
 
 import DataAccessError from '../../errors/data-access.error.js';
 import BaseCollection from '../base/base.collection.js';
+import {
+  applyWhere,
+  decodeCursor,
+  DEFAULT_PAGE_SIZE,
+  encodeCursor,
+  toDbField,
+} from '../../util/postgrest.utils.js';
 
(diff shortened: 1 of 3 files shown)


packages/spacecat-shared-data-access/src/models/site/site.collection.js
```diff
@@ -206,6 +213,130 @@ class SiteCollection extends BaseCollection {
...
+    }
+    query = query.order(orderField, { ascending });
+    const idField = this.fieldMaps.toDbMap[this.idName];
+    if (idField !== orderField) {
+      query = query.order(idField, { ascending });
+    }
+
+    // Honor the exact limit (no cap, no +1) so the api-service N+1 hasMore
+    // detection stays symmetric with Site.all's postgrest path.
+    const effectiveLimit = Number.isInteger(limit) ? limit : DEFAULT_PAGE_SIZE;
```
MysticatBot: issue (blocking): Zero or negative `limit` produces invalid PostgREST range bounds.

`Number.isInteger(0)` and `Number.isInteger(-5)` are both `true`. With `limit: 0`, this computes `.range(offset, offset - 1)` (inverted range - undefined PostgREST behavior). Negative values are worse.

The method is called directly from the api-service with a caller-supplied limit, so defensive validation is warranted even though BaseCollection `#queryPage` has the same trust-the-caller pattern (that method receives pre-sanitized input from internal callers).

**Fix:**
```js
const effectiveLimit = (Number.isInteger(limit) && limit > 0) ? limit : DEFAULT_PAGE_SIZE;
```

MysticatBot (changes requested):
Hey @bottadobe,

**Verdict: Request changes** - one input-validation gap that produces undefined PostgREST behavior.
**Complexity: HIGH** - large diff (500+ lines, predominantly tests).
_Changes: Adds `SiteCollection.allByEnrollmentFiltered()` for paginated, single-query site filtering by entitlement tier/product code via PostgREST inner-join embedding (3 files)._

## Must fix before merge
1. [Important] Zero/negative `limit` passes `Number.isInteger()` and produces invalid PostgREST range bounds - `site.collection.js:312` (details inline)

<details>

- nit: `delete siteRow.site_enrollments` triggers V8 hidden-class de-optimization on the shallow clone - destructuring rest (`const { site_enrollments: _, ...siteRow } = row`) is more idiomatic - `site.collection.js:326`
- suggestion: Add a unit test for `returnCursor: true` with empty result set to explicitly verify `{ data: [], cursor: null }` shape
- suggestion: Consider an upper-bound cap on `limit` (e.g. 10000) as defense-in-depth against accidental full-table scans from unsanitized caller input

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 36s | <b>Cost:</b> $5.68 | <b>Commit:</b> <code>377e218be161a241f7662d1f90abdf654daa76e4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @bottadobe,

**Verdict: Approve** - the prior blocking finding is resolved; no new issues.
**Complexity: HIGH** - large diff; database/model signal.
_Changes: Adds `&& limit > 0` guard to `allByEnrollmentFiltered()` so zero/negative limit falls back to DEFAULT_PAGE_SIZE instead of producing invalid PostgREST range bounds (2 files)._

- Zero/negative `limit` now guarded with `&& limit > 0`, falling back to DEFAULT_PAGE_SIZE. Test added for both `0` and `-5`.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 1s | <b>Cost:</b> $2.19 | <b>Commit:</b> <code>9ef4ef71d025b2bdf1e679065ce53d5e74fad9fb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1878, opened 2026-08-13
feat(tokowaka-client): mirror applyStale onto suggestion.data for patched deploys

dhavkuma:
## Summary
- `deploySuggestions()` now returns `suggestionIdsHavingPatches`, the set of suggestion IDs for which a patch was actually written to the S3 config (derived from `newConfig.patches[].suggestionId` per URL).
- `#deployPerUrlSuggestions()` mirrors `metadata.applyStale` onto `suggestion.data.applyStale` (display-only, same pattern as `edgeDeployed`) — but only for suggestions in `suggestionIdsHavingPatches`. Suggestions that deploy successfully without producing a patch (e.g. prerender-only, which never generates DOM patches) are left without the field, since there's nothing for `applyStale` to keep alive.
- `stripSuggestion()` (used on rollback) now also clears `applyStale` alongside `edgeDeployed`/`tokowakaDeployed`, so a rollback doesn't leave a stale flag behind.
- The functional patch-level `applyStale` (stamped onto `newConfig.patches` in S3) was already implemented prior to this change; this PR closes the suggestion.data mirror half of it.

## Test plan
- [x] `npm test -w packages/spacecat-shared-tokowaka-client` — 988 passing, 100% line/branch coverage
- [x] `npm run lint -w packages/spacecat-shared-tokowaka-client` — clean
- [x] New unit tests cover: `applyStale` mirrored when `metadata.applyStale` is true/false, excluded for prerender (no patch generated), and defaults safely when a caller's `deploySuggestions` stub omits `suggestionIdsHavingPatches`

Companion PR (consumes this once released): adobe/spacecat-api-service (to follow)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1879, opened 2026-08-13
feat(tokowaka-client): add setApplyStaleForSuggestions to toggle applyStale without redeploy

dhavkuma:
## Summary
- New `TokowakaClient.setApplyStaleForSuggestions(site, opportunity, suggestions, { applyStale, updatedBy })` sets or clears `applyStale` on already-deployed, per-URL suggestions **without redeploying**: updates the matching S3 patch and mirrors the same value onto `suggestion.data`, invalidating CDN only for URLs actually changed.
- Eligibility: only per-URL suggestions (pattern/domain-wide are excluded) that have a matching patch already in S3 are eligible; everything else is returned in `failedSuggestions` with a reason, mirroring the `rollbackSuggestions`/`deploySuggestions` result shape.
- This is the shared-client half of a new customer-facing "keep this optimization alive" toggle that doesn't require a full redeploy. Companion PR: adobe/spacecat-api-service (new `PATCH .../suggestions/edge-apply-stale` endpoint, to follow).

Builds on top of the not-yet-merged `applyStale` deploy-time plumbing from adobe/spacecat-shared#1878 (this PR's diff includes those commits since #1878 hasn't merged yet).

## Test plan
- [x] `npm test -w packages/spacecat-shared-tokowaka-client` — 100% line/branch/statement coverage
- [x] `npm run lint -w packages/spacecat-shared-tokowaka-client` — clean
- [x] New tests cover: set/clear on matching patch, mirror onto suggestion.data, no-op-but-still-succeeds when patch already matches, ineligible suggestions (no patch, pattern/domain-wide, no existing config), multi-URL CDN invalidation scoping, and S3/CDN error handling

### DIFF
--- a/packages/spacecat-shared-tokowaka-client/src/index.js
+++ b/packages/spacecat-shared-tokowaka-client/src/index.js
@@ -774,6 +774,7 @@ class TokowakaClient {
       return {
         succeededSuggestions: [],
         failedSuggestions: ineligibleSuggestions,
+        suggestionIdsHavingPatches: new Set(),
       };
     }
 
@@ -829,8 +830,9 @@ class TokowakaClient {
 
         // Upload to S3
         const s3Path = await this.uploadConfig(fullUrl, config);
-        // eslint-disable-next-line max-len
-        const suggestionIdsFromNewPatches = newConfig.patches.map((patch) => patch.suggestionId).filter(Boolean);
+        const suggestionIdsFromNewPatches = newConfig.patches
+          .map((patch) => patch.suggestionId)
+          .filter(Boolean);
         return { s3Path, fullUrl, suggestionIdsFromNewPatches };
       },
       URL_CONFIG_CONCURRENCY,
@@ -1518,9 +1520,14 @@ class TokowakaClient {
         const updated = { ...currentData, edgeDeployed: deploymentTimestamp };
         // Only mirror applyStale for suggestions that actually produced a patch — e.g.
         // prerender-only suggestions deploy successfully but never generate one, so there
-        // is nothing for applyStale to keep alive.
+        // is nothing for applyStale to keep alive. Matches the edgeDeployed contract
+        // (present-when-true, absent otherwise) instead of ever writing an explicit false.
         if (suggestionIdsHavingPatches.has(s.getId())) {
-          updated.applyStale = applyStale;
+          if (applyStale) {
+            updated.applyStale = true;
+          } else {
+            delete updated.applyStale;
+          }
         }
         const statusesToExcludeFromOptimization = ['STALE', 'LAST_MOD_MISSING'];
         if (statusesToExcludeFromOptimization.includes(updated.edgeOptimizeStatus)) {
--- a/packages/spacecat-shared-tokowaka-client/src/index.js
+++ b/packages/spacecat-shared-tokowaka-client/src/index.js
@@ -741,7 +741,9 @@ class TokowakaClient {
    * @param {Object} site - Site entity
    * @param {Object} opportunity - Opportunity entity
    * @param {Array} suggestions - Array of suggestion entities to deploy
-   * @returns {Promise<Object>} - Deployment result with succeeded/failed suggestions
+   * @returns {Promise<Object>} - Deployment result with succeeded/failed suggestions, plus
+   *   suggestionIdsHavingPatches: a Set of suggestion IDs for which a patch was actually generated
+   *   (excludes e.g. prerender-only suggestions, which deploy successfully with no patch).
    */
   async deploySuggestions(site, opportunity, suggestions, metadata = {}) {
     const { applyStale = false } = metadata;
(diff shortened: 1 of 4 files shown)


packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1872,6 +1902,138 @@ class TokowakaClient {
...
+        const updatedPatches = patches.map((patch) => {
+          const matchedSuggestion = urlSuggestions.find((s) => s.getId() === patch.suggestionId);
+          if (!matchedSuggestion) {
+            return patch;
+          }
+          succeededIds.add(matchedSuggestion.getId());
+
+          if (applyStale) {
+            if (patch.applyStale === true) {
+              return patch;
```
MysticatBot, line 1968: issue (blocking): **succeededIds populated before uploadConfig causes S3/DB state divergence on partial failure.**

Inside `setApplyStaleForSuggestions`, the `patches.map()` callback adds suggestion IDs to `succeededIds` as soon as a patch match is found:

```javascript
succeededIds.add(matchedSuggestion.getId());
```

But `uploadConfig` is called after this loop. If `uploadConfig` throws:
1. The matched IDs are already in `succeededIds`
2. The catch block skips them (`!succeededIds.has(...)` is false)
3. They end up in `succeededSuggestions` at the end
4. Their `suggestion.data` is mutated and persisted to the database
5. But the S3 config was never actually written

Result: the database says `applyStale: true` but S3 still has the old patch without it. The edge serves behavior that contradicts what the DB/UI shows.

**Fix:** Collect matched IDs in a per-URL local set, and only merge into `succeededIds` after `uploadConfig` succeeds (or the no-op path confirms no write was needed). Move the "No patch found" forEach to after the upload as well, so it cannot double-push on uploadConfig failure.

packages/spacecat-shared-tokowaka-client/src/index.js
```diff
@@ -1872,6 +1902,138 @@ class TokowakaClient {
...
+    opportunity,
+    suggestions,
+    { applyStale, updatedBy } = {},
+  ) {
+    const baseURL = getEffectiveBaseURL(site);
+
+    const perUrlSuggestions = suggestions.filter((s) => {
+      const data = s.getData();
+      return data?.url && !Array.isArray(data?.allowedRegexPatterns);
+    });
```
MysticatBot, line 1933: issue (blocking): **applyStale defaults to undefined with no validation guard.**

The destructuring `{ applyStale, updatedBy } = {}` means `applyStale` is `undefined` when the caller omits it or passes `{}`. Since the method branches on truthiness (`if (applyStale) { ... }`), an undefined value silently takes the "clear" path and removes `applyStale` from every matched patch and suggestion.

Compare `deploySuggestions` which uses `const { applyStale = false } = metadata;` (explicit default). For this standalone method named `setApplyStaleForSuggestions`, a caller forgetting to pass the parameter should not silently mutate production S3 configs.

**Fix:** Add an early guard:
```javascript
if (typeof applyStale !== 'boolean') {
  throw new Error('applyStale option must be a boolean');
}
```
Or at minimum default to `false` in the destructuring so the contract is explicit.

MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - data consistency bug when uploadConfig fails after succeededIds is populated.
**Complexity: MEDIUM** - medium diff, single package.
_Changes: Adds setApplyStaleForSuggestions to toggle the applyStale flag on already-deployed S3 patches without redeploying, plus wires applyStale mirroring into the existing deploy path (4 files)._

## Must fix before merge
1. [Critical] succeededIds populated before uploadConfig - suggestions reported as succeeded when S3 write never completed - `packages/spacecat-shared-tokowaka-client/src/index.js` (details inline)
2. [Important] No guard on applyStale parameter - undefined silently triggers the "clear" path for all matched suggestions - `packages/spacecat-shared-tokowaka-client/src/index.js` (details inline)

<details>

- suggestion: Use concurrent URL processing (processInBatches or similar) to match the established pattern in deploySuggestions, which uses URL_CONFIG_CONCURRENCY for the same read-modify-write cycle - `packages/spacecat-shared-tokowaka-client/src/index.js`
- suggestion: Consider extracting a shared private helper for the S3 read-modify-write-invalidate cycle, since clearApplyStaleFromPatches duplicates most of the same logic and they could diverge on bug fixes over time
- nit: failedSuggestions shape ({ suggestion, reason, statusCode }) diverges from deploySuggestions which returns raw entities - consider aligning or documenting the difference in JSDoc
- nit: Add a test for uploadConfig failure to cover the data consistency scenario in the critical finding above
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 47s | <b>Cost:</b> $7.15 | <b>Commit:</b> <code>eb277f44e70add61428d9e77d2e1007d3a45a4c6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #1881, opened 2026-08-17
Add a path-preserving site-identity normalizer for Semrush primary_url and internal Site lookup


rainer-friederich:
## Summary

A SpaceCat Site can carry a subpath in its `base_url`, and a Semrush project can track that subpath. Deriving both from one site needs **two different values**, and today only one exists.

- The Semrush project's `domain` must be a bare FQDN. Sending a path returns HTTP 400.
- The Semrush project's `settings.ai.primary_url` is the URL actually tracked, and it accepts a subdomain, a subpath, or both.
- Internal Site lookup needs the full identity, so that `nba.com/kings` resolves to its own Site rather than to `nba.com`.

`spacecat-shared` already owns the URL primitives for this (`packages/spacecat-shared-utils/src/url-helpers.js` holds `isWithinSiteScope`, `canonicalizeUrl`, `composeBaseURL`, `stripWWW`, and `isWithinSiteScope` is already subpath-aware). The missing piece is a path-preserving **identity** primitive to sit beside the existing host-only one, so every consumer computes it the same way instead of each repo rolling its own.

## The upstream contract

Verified live against `adobe-hackathon.semrush.com` on 2026-08-17. Full evidence and the provisioning consequences are in https://github.com/adobe/serenity-docs/issues/348 .

`domain` on create:

| sent | result |
|---|---|
| `shop.example.com` | 200, stored `example.com` — normalized to eTLD+1 |
| `example.com/shop` | **400** `Field validation for 'Domain' failed on the 'fqdn' tag` |

`primary_url` via `PATCH` (create ignores it; `PATCH` requires `type`):

| sent | stored |
|---|---|
| `shop.example.com/uk/en` | `shop.example.com/uk/en` |
| `https://shop.example.com/uk` | `shop.example.com/uk` — scheme stripped |
| `example.com/shop/` | `example.com/shop` — trailing slash normalized away upstream |

So the host-only derivation is correct and load-bearing for `domain` — it must not be replaced. What is missing is the second value.

## Current state

`spacecat-api-service/src/support/url-utils.js:31-41`:

```js
export function hostnameFromUrlString(value) {
  if (!value || !hasText(value)) return null;
  try {
    const url = new URL(value.includes('://') ? value : `https://${value}`);
    return url.hostname || null;
  } catch { return null; }
}
```

It is the single normalizer behind every brand → Semrush derivation, serving two roles that need to diverge:

| Call site | Role | Correct value |
|---|---|---|
[...]

PR #1882, opened 2026-08-17
feat(utils): add siteIdentityFromUrlString for path-preserving site identity

rainer-friederich:
## 1. Abstract

Adds `siteIdentityFromUrlString` to `@adobe/spacecat-shared-utils`: a path-preserving site-identity primitive that sits beside the existing host-only `hostnameFromUrlString`, together with a conformance corpus shared byte-for-byte with a Python counterpart in `mysticat-data-service`.

## 2. Reasoning

A SpaceCat Site can carry a subpath in its `base_url`, and a Semrush AIO project can track that subpath. Deriving both from one site needs **two different values**, and only the host-only one existed.

- A Semrush project's `domain` must be a bare FQDN. Sending a path returns HTTP 400, and the upstream normalizes whatever it receives to the registrable domain.
- A Semrush project's `settings.ai.primary_url` is the URL actually tracked, and it accepts a subdomain, a subpath, or both.

With one normalizer serving both roles, a brand whose site is `nba.com/kings` is recorded against the whole of `nba.com`. The same host-only assumption was independently re-implemented in Python, which is the drift this primitive exists to stop.

`siteIdentityFromUrlString(value)` returns the lowercased host plus the normalized path, with scheme, credentials, port, query and fragment removed — `https://www.NBA.com/kings/?x=1` becomes `www.nba.com/kings`. It returns `null` when the input is unparseable, matching the null handling call sites already have.

Normalization rules are fixed by the corpus rather than left to callers:

- The host is lowercased; the path keeps its case, since a case-sensitive origin may serve `/OMES` and `/omes` differently.
- A single trailing slash is stripped, so `experian.co.uk/business/` and `experian.co.uk/business` agree. Both spellings exist in prod.
- A trailing `.html` is **not** stripped. `example.com/abc.html` scopes a site to one page and `example.com/abc` to a whole section, and `oklahoma.gov/omes` and `oklahoma.gov/omes.html` are both live.
- `www.` is **not** collapsed against apex — they are distinct sites.
- The result is deliberately scheme-less, because that is the form the upstream stores. Returning the scheme-ful spelling would make every drift comparison report a difference that is not there and rewrite the same value forever.

`hostnameFromUrlString` is untouched: it remains correct and load-bearing for `domain`, where a path is rejected outright. For a path-free site URL the new primitive returns exactly what it returns, which is what makes this a no-op for the ~12,486 path-free sites in prod.

Behaviour delta: none on its own. Nothing in this repo calls the new function yet; it is the primitive the consumers in section 5 build on.

## 4. Required information

- Jira / issue: https://github.com/adobe/serenity-docs/issues/348 (umbrella) and https://github.com/adobe/spacecat-shared/issues/1881 (this workstream)
- Other: the upstream contract was established live against `adobe-hackathon.semrush.com` and is recorded in the umbrella issue; the vendored Project Engine spec lives at `serenity-docs/docs/prompt-management/attachments/projectengine-swagger-public.yaml`

- **spacecat-api-service** — consumes. Splits its two Semrush derivations onto this primitive and `hostnameFromUrlString`. It pins `@adobe/spacecat-shared-utils` to an exact version, so it cannot adopt this until the release cut from this PR is published.
- **mysticat-data-service** — contract. Holds the Python counterpart `site_identity` and a byte-identical copy of the corpus fixture. The two implementations write and compare the same Semrush field, so they are pinned to one corpus by a sha256 assertion in each suite: editing either copy alone fails that copy's own suite, which is the signal to sync the other.

The two implementations were differential-fuzzed against each other rather than trusted to a hand-written corpus, because a hand-written corpus only pins the cases someone thought of.

- An initial hand-picked set of 56 inputs produced **10 disagreements**. Two of them occur in real customer data: internationalized hosts (`bücher.de`) and non-ASCII paths (`example.com/ünï`), where Node's WHATWG `URL` punycodes and percent-encodes and Python's `urlsplit` does not. Left alone, one side would have written the encoded form while the other computed the raw form, so those brands would have reported drift on every reconcile and had a correct value rewritten forever. The remaining disagreements were leading slashes, protocol-relative input, backslashes and IPv6 brackets.
- After converging both implementations, they agree on **4,000 randomized inputs** (scheme × host × port × path × query, including IDN and pre-encoded paths) and on **30 real site `base_url` values read from the dev database**, with zero divergences.
- The corpus now includes the internationalized and non-ASCII cases rather than excluding them, so the agreement is pinned rather than assumed.

## 7. Test plan

**Locally, beyond unit tests:** both implementations were run head-to-head over the fuzz corpora and the real dev `base_url` sample described in section 6; the sha256 pinning the JS and Python corpus copies together was verified to match the file on both sides.

**Verification once released:** this package has no runtime surface of its own. The behaviour it enables is verified in the consumers — provision a brand whose site `base_url` carries a subpath and confirm the resulting Semrush project reads back `domain` as the apex and `settings.ai.primary_url` as the full subpath. Against the prod fleet, the invariant to hold is that for every site whose `base_url` has no path, the new primitive returns exactly what `hostnameFromUrlString` returns today.

This is the first of four coordinated PRs on branch `feat/348`.

- https://github.com/adobe/spacecat-api-service/pull/3073 — this PR **blocks** it. That repo pins `@adobe/spacecat-shared-utils` to an exact version and imports this primitive, so it cannot go green until this PR merges and semantic-release publishes a new version, which it then bumps to. It is open as a draft for that reason.
- https://github.com/adobe/mysticat-data-service/pull/919 — related. Independent at runtime (different language, no npm dependency), but holds the Python counterpart and the shared corpus copy.
- https://github.com/adobe/project-elmo-ui/pull/2875 — related. Fully independent; rendering only.

Order: merge this PR first and let it release. `mysticat-data-service` and `project-elmo-ui` may merge at any time, before or after. `spacecat-api-service` merges last, after bumping to the released version.


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean, well-tested utility addition with no blocking issues.
**Complexity: MEDIUM** - 173 added lines across 6 files; single-service utility.
_Changes: Adds a path-preserving URL identity normalizer (`siteIdentityFromUrlString`) to spacecat-shared-utils, with a sha256-pinned conformance corpus shared with the Python counterpart (6 files)._

<details>

- suggestion: Scheme-relative input (`//host/path`) passes the slash guard then hits `https://${raw}` producing `https:////host/path`, which the URL parser collapses correctly. An explicit `new URL('https:' + raw)` branch for `//`-prefixed inputs would make the parse path obvious to future readers. The corpus pins the behavior either way. - `packages/spacecat-shared-utils/src/url-helpers.js:612`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 4s | <b>Cost:</b> $3.62 | <b>Commit:</b> <code>15d40d9ad51aab0aa97b6209aece63934f5a3d4d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Thanks — review addressed. 1 non-blocking item, 1 decision.

**Non-blocking (1)**

- **Fixed (as a comment, not a branch)** — scheme-relative parse path, `url-helpers.js`. The observation is right: `//host/path` has no `://`, so it becomes `https:////host/path` and depends on the WHATWG parser skipping the slash run after the scheme. I documented the mechanism rather than adding an explicit `new URL('https:' + raw)` branch: the branch would duplicate a parse the parser already specifies, and it adds a path to cover against this package's 97% branch floor for no behavioural gain. The corpus row `//example.com/x` → `example.com/x` remains the thing that actually pins it.

On the degraded-review note: there is no separate spec document for this change. The design of record is the umbrella issue https://github.com/adobe/serenity-docs/issues/348 , which carries the upstream `domain` vs `settings.ai.primary_url` contract established live against the Semrush hackathon environment; it is linked from section 4.

MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - clean utility addition, prior suggestion addressed by expanded comment.
**Complexity: MEDIUM** - 178 added lines across 6 files; single-service utility.
_Changes: Adds a path-preserving URL identity normalizer (`siteIdentityFromUrlString`) to spacecat-shared-utils, with a sha256-pinned conformance corpus shared with the Python counterpart (6 files)._

- Scheme-relative parse path now documented inline (expanded comment explaining `//host/path` -> `https:////host/path` WHATWG parser behavior, pinned by corpus)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 56s | <b>Cost:</b> $3.31 | <b>Commit:</b> <code>197c76185e609ca2128a14ce7404b14505245782</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #348, opened 2024-08-28
fix: sharepoint resolve path twice

alinarublea:
Please ensure your pull request adheres to the following guidelines:
- [ ] make sure to link the related issues in this description
- [ ] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes

## Related Issues

Thanks for contributing!

PR #1884, opened 2026-08-17
feat(utils): add siteIdentityFromUrlString for Semrush primary_url (#348)

byteclimber:
## What
Adds `siteIdentityFromUrlString` to `spacecat-shared-utils`: derives a stable site identity (lowercased host + normalized path; strips scheme/userinfo/port/query/fragment; strips a single trailing slash; preserves a trailing `.html`; does not collapse `www.`; returns `null` on unparseable input).

## Why
serenity-docs#348 — Semrush `settings.ai.primary_url` is the URL a brand actually tracks (may carry a subdomain/subpath), distinct from the host-only `domain`. This is the shared primitive so the api-service (JS) and the data-service migration (Python) derive the same value under the same rules.

## Tests
8 new unit tests (subpath, trailing slash, `.html` preserved, case, scheme/userinfo/port/query/fragment stripping, www not collapsed, null cases). 100/97 coverage maintained.

Base is the session branch (this work layers on the PromiseTokenSession commit); will retarget to `main` once the base merges.

### DIFF
--- a/packages/spacecat-shared-utils/src/index.d.ts
+++ b/packages/spacecat-shared-utils/src/index.d.ts
@@ -167,6 +167,15 @@ export declare function canonicalizeUrl(
 export declare function composeBaseURL(domain: string): string;
 
 /**
+ * Derives a stable "site identity" (lowercased host + normalized path, no
+ * scheme/credentials/port/query/fragment) from a URL-ish string. Preserves the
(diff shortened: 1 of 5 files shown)


byteclimber:
@MysticatBot please review

byteclimber:
@MysticatBot please review

PR #1885, opened 2026-08-18
feat(utils): add siteIdentityFromUrlString for Semrush primary_url (#348)

byteclimber:
## What
Adds `siteIdentityFromUrlString` to `spacecat-shared-utils`: derives a stable site identity (lowercased host + normalized path; strips scheme/userinfo/port/query/fragment; strips a single trailing slash; preserves a trailing `.html`; does not collapse `www.`; returns `null` on unparseable input).

## Why
serenity-docs#348 — Semrush `settings.ai.primary_url` is the URL a brand actually tracks (may carry a subdomain/subpath), distinct from the host-only `domain`. This is the shared primitive so api-service (JS) and the data-service migration (Python) derive the same value.

## Tests
8 new unit tests + the new export registered in the index-exports assertion. 100/97 coverage maintained.

## Note
This is the `main`-targeting version of the shared #348 change (supersedes #1884, which was layered on the session branch and thus skipped by MysticatBot as a non-default-branch PR). Self-contained — no dependency on other session-branch work.

### DIFF
--- a/packages/spacecat-shared-utils/src/url-helpers.js
+++ b/packages/spacecat-shared-utils/src/url-helpers.js
@@ -88,6 +88,53 @@ function composeBaseURL(domain) {
 }
 
 /**
+ * Derives a stable "site identity" from a URL-ish string: the lowercased host
+ * plus its normalized path, with no scheme, credentials, port, query or
+ * fragment. Unlike {@link composeBaseURL} (which also strips the path and
+ * `www.`), this preserves the path so two sites on the same host but different
+ * subpaths — e.g. `nba.com/kings` vs `nba.com/knicks` — yield distinct
+ * identities. It mirrors how Semrush normalizes `settings.ai.primary_url`.
+ *
+ * Normalization rules (fixed by tests, not left to callers):
+ * - The host is lowercased; the path is left case-sensitive.
+ * - Scheme, userinfo, port, query and fragment are stripped.
+ * - A single trailing slash is removed (`example.com/a/` -> `example.com/a`),
+ *   so a bare host and a bare-root URL agree (`example.com`).
+ * - A trailing `.html` is preserved — `example.com/x.html` (one page) and
+ *   `example.com/x` (a section) are deliberately distinct (SITES-49656).
+ * - `www.` is NOT collapsed against the apex, matching the `domain` derivation.
+ * - Returns `null` on unparseable input, so call sites keep their null handling.
+ *
+ * @param {string} value - A URL or host(/path) string, with or without scheme.
+ * @returns {string|null} The site identity, or `null` if unparseable.
+ */
+function siteIdentityFromUrlString(value) {
+  if (!hasText(value)) {
+    return null;
+  }
+  const trimmed = value.trim();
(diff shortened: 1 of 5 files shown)


byteclimber:
Superseded: `siteIdentityFromUrlString` is already merged to `main` (a sibling PR landed the identical primitive + tests). This PR is a duplicate and only fails lint on duplicate declarations. Closing as redundant — WS1 of serenity-docs#348 is already done on main.

packages/spacecat-shared-utils/src/url-helpers.js
```diff
@@ -87,6 +87,53 @@ function composeBaseURL(domain) {
   return baseURL;
 }
 
+/**
```
MysticatBot, line 90: issue (blocking): **Duplicate function definition - PR is stale against main.**

`siteIdentityFromUrlString` already exists on `main` (line 605 of this file) as an inline `export function` with a more robust implementation. This PR adds a second definition at line 87 plus a named-export entry. In an ES module, this produces a `SyntaxError` at parse time (duplicate binding in module scope).

The same duplication exists in `index.d.ts` (line 157 on main already declares this) and `index.js` (already re-exports it).

The main-branch version also includes a leading-slash guard (`if (raw.startsWith('/') && !raw.startsWith('//')) return null`) that this PR version lacks, preventing inputs like `/just/a/path` from being misinterpreted.

**Fix:** Rebase on main. If the work has already landed (it has), close this PR. If there are intentional differences in normalization rules compared to what shipped, submit those as an incremental modification to the existing function.

MysticatBot (changes requested):
Hey @byteclimber,

**Verdict: Request changes** - this PR introduces a duplicate definition of a function that already exists on main; merging would produce a SyntaxError.
**Complexity: LOW** - single utility function addition.
_Changes: Adds `siteIdentityFromUrlString` URL-normalization utility to spacecat-shared-utils (5 files)._

## Must fix before merge

1. [Critical] Duplicate `siteIdentityFromUrlString` definition - will cause SyntaxError at parse time - `packages/spacecat-shared-utils/src/url-helpers.js:87` (details inline)

<details>

- suggestion: Non-web scheme URLs (`javascript://`, `ftp://`) with a parseable hostname produce a truthy result instead of null. The main-branch version does not guard against this either, but restricting to http(s) protocols would make the contract more honest. - `packages/spacecat-shared-utils/src/url-helpers.js:120`
- suggestion: Only the final trailing slash is stripped; consecutive slashes like `/a//` yield an identity still ending with `/`. Unlikely in Semrush data but worth a test to pin the decision. - `packages/spacecat-shared-utils/src/url-helpers.js:130`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 3s | <b>Cost:</b> $6.63 | <b>Commit:</b> <code>6228113c2e3964e51f45303ab3732d6a8be7f130</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

